Issue #3, opened 2018-03-15
Bug #500753 - Concurrency failure when persisting across multiple threads


dazey3:
Linking to [BZ #500753](https://bugs.eclipse.org/bugs/show_bug.cgi?id=500753)

dazey3:
Closed in BZ, closing this issue

PR #4, opened 2018-03-15
[master] Bug 500753: Synchronize initialization of InsertQuery

dazey3:
for #3 

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
@@ -19,6 +19,8 @@
  *       - 365931: @JoinColumn(name="FK_DEPT",insertable = false, updatable = true) causes INSERT statement to include this data value that it is associated with
  *     02/14/2018-2.7.2 Lukas Jungmann
  *       - 530680: embedded element collection within an entity of protected isolation does not merged changes into clones correctly
+ *     03/14/2018-2.7 Will Dazey
+ *       - 500753: Synchronize initialization of InsertQuery
  ******************************************************************************/
 package org.eclipse.persistence.mappings;
 
@@ -35,6 +37,7 @@ import java.util.Vector;
 import org.eclipse.persistence.descriptors.ClassDescriptor;
 import org.eclipse.persistence.descriptors.DescriptorEvent;
 import org.eclipse.persistence.descriptors.DescriptorEventManager;
+import org.eclipse.persistence.descriptors.DescriptorQueryManager;
 import org.eclipse.persistence.descriptors.changetracking.AttributeChangeTrackingPolicy;
 import org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy;
 import org.eclipse.persistence.descriptors.changetracking.ObjectChangeTrackingPolicy;
@@ -2526,11 +2529,15 @@ public class AggregateCollectionMapping extends CollectionMapping implements Rel
      * Returns clone of InsertObjectQuery from the reference descriptor, if it is not set - create it.
      */
     protected InsertObjectQuery getInsertObjectQuery(AbstractSession session, ClassDescriptor desc) {
-        InsertObjectQuery insertQuery = desc.getQueryManager().getInsertQuery();
-        if (insertQuery == null) {
-            insertQuery = new InsertObjectQuery();
-            desc.getQueryManager().setInsertQuery(insertQuery);
+        DescriptorQueryManager queryManager = desc.getQueryManager();
+        if (!queryManager.hasInsertQuery()) {
+            synchronized (queryManager) {
+                if (!queryManager.hasInsertQuery()) {
+                    queryManager.setInsertQuery(new InsertObjectQuery());
+                }
+            }
         }
+        InsertObjectQuery insertQuery = queryManager.getInsertQuery();
         if (insertQuery.getModifyRow() == null) {
             AbstractRecord modifyRow = new DatabaseRecord();
             for (int i = 0; i < getTargetForeignKeyFields().size(); i++) {
--- a/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/concurrency/TestConcurrencyPersistence.java
+++ b/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/concurrency/TestConcurrencyPersistence.java
@@ -1,0 +1,128 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation. All rights reserved.
+ * This program and the accompanying materials are made available under the 
+ * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0 
+ * which accompanies this distribution. 
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at 
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ *     03/14/2018-2.7 Will Dazey
+ *       - 500753: Synchronize initialization of InsertQuery
+ ******************************************************************************/
+package org.eclipse.persistence.jpa.test.concurrency;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.EntityTransaction;
+import javax.persistence.RollbackException;
+
+import org.eclipse.persistence.jpa.test.concurrency.model.User;
+import org.eclipse.persistence.jpa.test.concurrency.model.UserTag;
+import org.eclipse.persistence.jpa.test.framework.DDLGen;
+import org.eclipse.persistence.jpa.test.framework.Emf;
+import org.eclipse.persistence.jpa.test.framework.EmfRunner;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(EmfRunner.class)
+public class TestConcurrencyPersistence {
+
+    @Emf(createTables = DDLGen.DROP_CREATE, classes = { User.class, UserTag.class })
+    private EntityManagerFactory emf;
+
+    /**
+     * Persisting multiple entities simultaneously on separate threads has the possibility of
+     * causing a null descriptor to be set on the InsertQuery. 
+     * This test spins up threads and runs them multiple times to replicate the race condition. 
+     * This test should pass everytime it's run, but there exists the possibility that it 
+     * passes when none of the threads actually test the collision.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void testInsertConcurrency() throws Exception {
+
+        Queue<Exception> errors = new ConcurrentLinkedQueue<>();
+        int runs = 10;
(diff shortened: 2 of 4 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=500753

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
```diff
@@ -2526,11 +2529,15 @@ public void preInsert(WriteObjectQuery query) throws DatabaseException, Optimist
      * Returns clone of InsertObjectQuery from the reference descriptor, if it is not set - create it.
      */
     protected InsertObjectQuery getInsertObjectQuery(AbstractSession session, ClassDescriptor desc) {
-        InsertObjectQuery insertQuery = desc.getQueryManager().getInsertQuery();
-        if (insertQuery == null) {
-            insertQuery = new InsertObjectQuery();
-            desc.getQueryManager().setInsertQuery(insertQuery);
+        DescriptorQueryManager queryManager = desc.getQueryManager();
+        if (!queryManager.hasInsertQuery()) {
+            synchronized (queryManager) {
```
lukasj: Can we avoid using synchronized? I recall there was https://bugs.eclipse.org/bugs/show_bug.cgi?id=436871 - maybe cause and fix can be the same?

dazey3: Are we worried about performance degradation from blocking threads? I'm not sure I see here how to prevent multiple threads from stepping on each other without using a synchronize block. 
I refrained from changing DescriptorQueryManager.setInsertQuery() code since I don't believe the issue is there. If multiple threads are calling setInsertQuery() simultaneously, I think it should be the responsibility of the callers to make sure they do not interfere with each other.
In this case, we have two (or more) threads that see (insertQuery = null) at the same time. They all then attempt to call setInsertQuery() almost simultaneously; passing unique object references. The last one to actually initialize is set on the QueryManager, but the other threads proceed assuming theyre object was set. This is the race condition. Making sure 1, and only 1, call is actually made to initialize the object seemed the most prudent fix to me.
I am open to suggestions if you feel there is another approach you have in mind.

dazey3: Any additional thoughts Lukas? I would like to get this issue delivered as soon as possible, but I would also like to address any issues or alternative implementations you may have in mind for this fix.

lukasj: I don't have strong opinion here, it's just that perf seems to have higher value than correctness (based on some parts of codebase). One idea - what about adding `insertQuery.setDescriptor(desc);` right after `new Insert...` in the original code?

dazey3: I do agree that ```insertQuery.setDescriptor(desc);``` may solve this issue. The InsertObjectQuery would then definitely have a descriptor set on it and the exception should be solved. However, I still think this is not a correct solution. What about ```setIsUserDefined(true)```, which is also part of the implementation of ```DescriptorQueryManager.setInsertQuery(InsertObjectQuery insertQuery)```? Should we duplicate that line as well? What about the fact that we are returning ```insertQuery``` at the end? The bad thread (who was not actually set on the QueryManager), will be returning a bad reference. Now we have a reference leak that could cause further issues.

Fixing issues in this way feels 'hacky' to me and ultimately duplicates code. Certainly an explicit  ```synchronize``` should be avoided due to bad performance. Blocking threads for extended periods of time is a real issue and we should not introduce that. However, I think the performance impact of this code is negligible. With my proposed change, threads should only be blocked ```if (!queryManager.hasInsertQuery())```. If a thread has already set the InsertObjectQuery, then all other threads will dodge the synchronize. I see no difference in performance between my code change and having two threads both entering ```if (insertQuery == null)``` simultaneously + creating a new instance + calling ```queryManager.setInsertQuery()```. The only difference is on the off change two threads enter the first if check at the same time, one will wait until the other finishes setting.

lukasj: It actually solves the issue, I've just tried that :-) I'm definitely not saying it is the right and/or complete solution, it just shows some alternative path towards expected behaviour. There are definitely some areas I do not completely understand yet but to share what I've noticed:

For non-aggregates - `classdescriptor.initialize(session)` which in turn always calls `DescriptorQueryManager.setInsertQuery(..)` is called (..so if I got this right then for non-aggregates insertQuery should always be somehow pre-prepared, say like a template for cloning, and never be null). For aggregates, particular mapping is responsible for calling `classdescriptor.initialize(session)` on the cloned instance of classdescriptor. In `AggregateCollectionMapping` this (cloning) falls to `initializeReferenceDescriptor` where `clonedDescriptor.initialize(session);` is called but apparently no queries are being initialized for aggregates due to conditions in `DescriptorQueryManager.initialize(session)` (..so if I got this part right then for aggregates insertQuery is never being pre-prepared and always null on the first call).

Now it is getting more interesting:
the method which is being altered here contains following comment: `Returns clone of InsertObjectQuery from the reference descriptor, if it is not set - create it.` - that is not what the method itself does - there is no cloning happening there (should the clone be returned only if query exists or always and also set on the queryManager or it is the comment which is wrong?), other than that - what about `setModifyRow` method which is called here and also later by the caller of this method with slightly different argument?

Based on this and also on comparing alternatives in different `Aggregate*Mapping` and various `get*Query` and/or `build*query` I tend to think that at least setting the query on the querymanager is wrong - no other similar code is doing that.

Also if we look at `getAndPrepareModifyQueryForInsert` method then we can easily spot that `checkPrepare` where NPE happens is called _before_ classdescriptor is set on the almost the same query being executed - based on this I tend to disagree with current fix

dazey3: I went ahead and delivered based on your comment that you didn't have a strong opinion. I am willing to continue looking at this and create further pull requests for this issue, but I worry this is a rat hole. I stand by my solution though since this synchronize will create the optimal, single initialization path. Whether initialization should be done here when ```== null``` or not... I'm not sure.  I agree that initialization should be done sooner and null values should not run around, but I am not familiar with why this code initializes here in the first place so I am not sure where this initialization should have been done. The java doc on AggregateCollectionMapping.getInsertObjectQuery may be incorrect so I don't think we can base anything on that.

Looking at ```getAndPrepareModifyQueryForInsert```, I am confused a bit by this code:

    InsertObjectQuery insertQueryFromDescriptor = getInsertObjectQuery(session, objReferenceDescriptor);
    //This call leads to the NPE
    insertQueryFromDescriptor.checkPrepare(session, insertQueryFromDescriptor.getModifyRow());

    //Why are we cloning now after checkPrepare? Was this clone meant to be done in getInsertObjectQuery()?
    InsertObjectQuery insertQuery = (InsertObjectQuery)insertQueryFromDescriptor.clone();
    insertQuery.setObject(object);
    //This 'objReferenceDescriptor' reference matches the reference on DescriptorQueryManager.descriptor, right? Which was already set on 'insertQueryFromDescriptor'?
    insertQuery.setDescriptor(objReferenceDescriptor);

EclipseLink's cloning logic is one of strategies I have never understood completely. When to clone() and when not to is beyond me.

lukasj: let's make it working properly first and work on improving the fix later - I have nothing against this approach as long as it does not break existing behaviour ;-)

The problem here is that initialization is not complete due to the descriptor not being set on the query and not because of sync IMHO. I'm usually trying to look around to see how it is being done on other places - see ie handling of DeleteQuery in [UnitOfWorkImpl](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java#L1449), or start from `objectAddedDuringUpdate` in [AggregateCollectionMapping](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java#L2313) and [CollectionMapping](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/CollectionMapping.java#L1698) - that does not necessarily mean that there is no issue on other places, it's just that it is less probable that the same problem would be spread across more usages - non of the usages is setting the query on the `querymanager` and each is working with "new" instance of the query, not saying that I've found no other call to  `querymanager.set*Query` anywhere in the codebase apart from descriptors and seession init code. Unless this is some really special case - which I doubt it is -  I believe that `queryManager.setInsertQuery(new InsertObjectQuery())` call does not belong to AggregateCollectionMapping (- as you have already said - this call may have consequences).

As for the query lifecycle/cloning logic - as I understand it:
* new instance is created
* checkPrepare (which forces prepare if needed) ensures that the query basic initial state (="unmodifiable" properties) is there and sets up the query for eventual execution - at this point we can take it as a "template" to which we're passing arguments for execution; if any of "unmodifiable" properties is updated later then checkPrepare(->prepare) must be called again
* execute runs the query on the given context - it clones itself for concurrency, then it sets context on itself and runs
* the `clone()` during execute can (and likely should) be avoided if we know we won't reuse the query later

Based on above I think that the implementation of `getInsertObjectQuery` method should follow the javadoc there, ie this way:
```java
    protected InsertObjectQuery getInsertObjectQuery(AbstractSession session, ClassDescriptor desc) {
        InsertObjectQuery insertQuery = desc.getQueryManager().getInsertQuery();
        if (insertQuery == null) {
            insertQuery = new InsertObjectQuery();
            insertQuery.setDescriptor(desc);
        } else {
            // Ensure original query has been prepared.
            insertQuery.checkPrepare(session, insertQuery.getTranslationRow());
            insertQuery = (InsertObjectQuery)insertQuery.clone();
        }
        return insertQuery;
    }
```
as for updates to `getAndPrepareModifyQueryForInsert` method which is used from multiple places, I think that calls to `checkPrepare()` and `clone()` should be removed as these should have been already handled by `getInsertObjectQuery()` above and it should be checked whether `insertQuery.setIsExecutionClone(true);` can be added there.

Do you think what I've written makes sense?

dazey3: Ok, this solution changes ```AggregateCollectionMapping.getInsertObjectQuery``` a bit more than I had felt comfortable doing, but I think it looks like ti would work to me. I am still not sure I understand when to clone() exactly, but I guess if it's to refrain from using original instances, this should work. Whether we should then be calling ```queryManager.setInsertQuery``` here... Im not sure. Also ```ModifyQuery.setModifyRow``` isn't being called anymore, is this a problem? I am not sure of the ramifications of these changes, but if the tests pass... right?

lukasj: The call to `insertQuery.setModifyRow` is there twice, the latter (in `getAndPrepareModifyQueryForInsert`) overwrites the former (from `getInsertObjectQuery`). The thing I'm not sure about is that the former (the one from `getInsertObjectQuery`) looks more correct to me than the latter - it takes into account possible ordering and collection member being Embeddable

dazey3: I went ahead and created PR #25. We can continue the discussion there so that we can comment on code directly

PR #8, opened 2018-03-19
Bug 531873 - XmlSchema Prefix is not honoured if root element is  nil

rfelcman:
MOXy marshaller ignores custom namespace prefix mapping specified in package-info class. It generates default one "ns0" instead custom "PRE". After patch it can use mapping from package-info class. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=531349 .


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=531873

rfelcman:
Sorry I produced there some mistake. This pull request is related with different bug (531349). In the title is wrong bug number. I'll create new one with correct title.

PR #20, opened 2018-03-23
Bug 27026457

sureshbn:
PROBLEM DESCRIPTION 
--------------------------

Behavior change in glass fish vs eclipselink  
12 (Eclipselink):
Test Bug 26988460 EJBIntegration ...
=== Details as string:  <?xml version='1.0' encoding='UTF-8'?> <addName
xmlns='http://ejb.wellsfargo.com/' xmlns:tns='http://ejb.wellsfargo.com/'>
<arg0 xmlns=''> <firstName>AAA</firstName> <lastName>B</lastName> </arg0>
</addName>
=== Details as XML:
<?xml version = '1.0' encoding = 'UTF-8'?>
<addName xmlns="http://ejb.wellsfargo.com/"
xmlns:tns="http://ejb.wellsfargo.com/"> <arg0 xmlns="">
<firstName>AAA</firstName> <lastName>B</lastName> </arg0> </addName>
=== method name = public abstract java.lang.String
com.wellsfargo.ejb.LocalBeanRemote.addName(com.wellsfargo.pojo.Details)
=== payload = {parameters=oracle.xml.parser.v2.XMLElement@1bb5a082}
<addName xmlns="http://ejb.wellsfargo.com/"
xmlns:tns="http://ejb.wellsfargo.com/"> <arg0 xmlns="">
<firstName>AAA</firstName> <lastName>B</lastName> </arg0> </addName>

****** Results ******
Details First Name= AAA
Details Last Name= B
Details Account Number = null
Details Acccount Name = null

11g (Glassfish):
Test Bug 26988460 EJBIntegration ...
=== Details as string:  <?xml version='1.0' encoding='UTF-8'?> <addName
xmlns='http://ejb.wellsfargo.com/' xmlns:tns='http://ejb.wellsfargo.com/'>
<arg0 xmlns=''> <firstName>AAA</firstName> <lastName>B</lastName> </arg0>
</addName>
=== Details as XML:
<?xml version = '1.0' encoding = 'UTF-8'?>
<addName xmlns="http://ejb.wellsfargo.com/"
xmlns:tns="http://ejb.wellsfargo.com/"> <arg0 xmlns="">
<firstName>AAA</firstName> <lastName>B</lastName> </arg0> </addName>
=== method name = public abstract java.lang.String
com.wellsfargo.ejb.LocalBeanRemote.addName(com.wellsfargo.pojo.Details)
=== payload = {parameters=oracle.xml.parser.v2.XMLElement@77847718}
<addName xmlns="http://ejb.wellsfargo.com/"
xmlns:tns="http://ejb.wellsfargo.com/"> <arg0 xmlns="">
<firstName>AAA</firstName> <lastName>B</lastName> </arg0> </addName>

****** Results ******
Details First Name= AAA
Details Last Name= B
Details Account Number = []
Details Acccount Name = []

Proposed fix
--------------
Issue can be replicated with user given test case . This fix will return an empty array for List [same as glassfish] in this specific seanario

### DIFF
--- a/moxy/eclipselink.moxy.test/src/org/eclipse/persistence/testing/jaxb/xmlmarshaller/UnmarshallerCollectionTestCases.java
+++ b/moxy/eclipselink.moxy.test/src/org/eclipse/persistence/testing/jaxb/xmlmarshaller/UnmarshallerCollectionTestCases.java
@@ -1,0 +1,109 @@
+package org.eclipse.persistence.testing.jaxb.xmlmarshaller;
+
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.xml.bind.JAXBElement;
+import javax.xml.bind.annotation.XmlList;
+import javax.xml.namespace.QName;
+import javax.xml.stream.XMLStreamReader;
+
+import org.eclipse.persistence.jaxb.JAXBContextFactory;
+import org.eclipse.persistence.jaxb.JAXBUnmarshaller;
+import org.eclipse.persistence.jaxb.TypeMappingInfo;
+import org.eclipse.persistence.jaxb.TypeMappingInfo.ElementScope;
+import org.eclipse.persistence.testing.jaxb.typemappinginfo.TypeMappingInfoTestCases;
+
+public class UnmarshallerCollectionTestCases  extends TypeMappingInfoTestCases {
+	
+    protected final static String XML_RESOURCE = "org/eclipse/persistence/testing/oxm/jaxb/UnmarshallCollection.xml";
+
+    public UnmarshallerObject testField;
+
+    @XmlList
+    public Object xmlListAnnotationField;
+
+    public UnmarshallerCollectionTestCases(String name) throws Exception {
+        super(name);
+        setupParser();
+        init();
+    }
+
+    public void init() throws Exception {
+        setControlDocument(XML_RESOURCE);
+        setTypeMappingInfos(getTypeMappingInfos());
+    }
+
+    protected TypeMappingInfo[] getTypeMappingInfos()throws Exception {
+    	  if(typeMappingInfos == null) {
+              typeMappingInfos = new TypeMappingInfo[1];
+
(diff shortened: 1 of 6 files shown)


lukasj:
Cannot accept this with  failing ip-validation check - there's at least missing sign-off. See https://wiki.eclipse.org/Development_Resources/Contributing_via_Git for details

moxy/eclipselink.moxy.test/src/org/eclipse/persistence/testing/jaxb/xmlmarshaller/UnmarshallerCollectionTestCases.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    protected String getNoXsiTypeControlResourceName() {
+        return XML_RESOURCE;
+    }
+
+   public void testUnmarshallerCollectionNullTestCase(){
+    	try{
+	    	if(null != XML_INPUT_FACTORY) {
+	    		InputStream instream = ClassLoader.getSystemResourceAsStream(resourceName);
```
lukasj: No TABS are allowed.

lukasj:
the logic of (un)marshalling collections is described ie at http://blog.bdoughan.com/2012/12/jaxb-representing-null-and-empty.html. If the domain object being unmarshalled is hand-written then the solution is to explicitly initialize the field to empty list, either in the constructor or as part of the field declaration.

The proposed solution looks more like a hack which complete ignores what happens in [MappingsGenerator](https://github.com/eclipse-ee4j/eclipselink/blob/master/moxy/org.eclipse.persistence.moxy/src/org/eclipse/persistence/jaxb/compiler/MappingsGenerator.java#L3468) where it explicitly forbids setting property in the target object to empty collection if no elements are found in the source XML. Changing the flag here from `false` to `true` makes the test passing as well and is definitely better solution. But making this change needs to be carefully tested and it is very likely that it would introduce incompatibilities with other existing applications. The correctness of this change wrt spec is also questionable.

lukasj:
closing based on my previous comment

PR #23, opened 2018-03-23
[master] Bug 487838: Metamodel gen fails if class hierarchy contains enclosed interfaces

lukasj:
Note that the test uses Stream and lambda call implying dependency on JDK8+ which should be fine for master (JDK9+) and 2.7 (JDK8+) streams


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=487838

PR #24, opened 2018-03-23
Add configuration for travis

lukasj:
While going through issues in old mirror I came across https://github.com/eclipse/eclipselink.runtime/pull/31 and extended that PR to also run LRG tests...

Do we want this in the main repo? I can imagine it can be extended to cover more DBs and tight to every PR... It is not meant to replace nightly jobs on eclipse.org infra. Opinions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: java
 sudo: required
 
+branches:
+  only:
+    - travis_experiment
+
 addons:
   apt:
     sources:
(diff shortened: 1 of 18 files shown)


lukasj:
So now it runs :-) The run currently takes around 37minutes, adding core-lrg will add additional ~11minutes to overall time but we will have to split the main test suite to smaller parts if we want to run it. An alternative would be to use in-memory hypersonic DB in this environment - it runs less tests then mysql but it is much much faster...

Travis "limitations" as I see them now:
* https://docs.travis-ci.com/user/common-build-problems/#Log-Length-exceeded
* https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

PR #25, opened 2018-03-23
Bug 500753: clean up implementation

dazey3:
for #3 


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=500753

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
```diff
@@ -2526,31 +2526,18 @@ public void preInsert(WriteObjectQuery query) throws DatabaseException, Optimist
...
-            desc.getObjectBuilder().buildTemplateInsertRow(session, modifyRow);
-            getContainerPolicy().addFieldsForMapKey(modifyRow);
-            if(this.listOrderField != null) {
-                modifyRow.put(this.listOrderField, null);
-            }
-            insertQuery.setModifyRow(modifyRow);
+        InsertObjectQuery insertQuery = desc.getQueryManager().getInsertQuery();
+        if (insertQuery == null) {
+            insertQuery = new InsertObjectQuery();
+            insertQuery.setDescriptor(desc);
```
dazey3: So, we have removed ```queryManager.setInsertQuery``` which makes the race condition moot. I'm not sure of the ramifications of this... but perhaps it's right.

dazey3: Another thing to point out is that as long as```desc.getQueryManager().getInsertQuery() == null```, we are going to keep making unique InsertObjectQuerys instance. These each will need to be prepared. The older implementation benefited from a single InsertObjectQuery instance that would only be prepared once.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
```diff
@@ -2561,21 +2548,23 @@ protected InsertObjectQuery getInsertObjectQuery(AbstractSession session, ClassD
...
         ClassDescriptor objReferenceDescriptor = getReferenceDescriptor(object.getClass(), session);
-        InsertObjectQuery insertQueryFromDescriptor = getInsertObjectQuery(session, objReferenceDescriptor);
-        insertQueryFromDescriptor.checkPrepare(session, insertQueryFromDescriptor.getModifyRow());
+        InsertObjectQuery insertQuery = getInsertObjectQuery(session, objReferenceDescriptor);
 
-        InsertObjectQuery insertQuery = (InsertObjectQuery)insertQueryFromDescriptor.clone();
         insertQuery.setObject(object);
         insertQuery.setDescriptor(objReferenceDescriptor);
 
-        AbstractRecord targetForeignKeyRow = new DatabaseRecord();
```
lukasj: this part should have stayed as it was - objectAddedDuringUpdate and pre/postInsert callers are doing what is necessary, sorry I didn't check callers of this method earlier

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
lukasj: Agree with you here, on the other hand we gain correctness. I believe that we can gain some penalty introduced by this back by setting `insertQuery.setIsExecutionClone(true);` later to avoid unnecessary double cloning - one explicit, the other during execute

dazey3: I'm not sure I understand why we are gaining correctness. Previously, we were setting the InsertObjectQuery back on the QueryManager. That was where the issue originally was due to the timing of setting InsertObjectQuery multiple times (hence the synch block I added). Now, we are dropping that and instead mass producing cloned instances. 
This isn't to say I disagree with making the change, but that I cannot say I understand why this code is "correct" over the previous.

lukasj: By gaining correctness I mean:
* we're setting classdescriptor on the query instance before calling checkPrepare
* mapping is not altering QueryManager during execution; only initialization code is supposed to do that

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
dazey3: Ok, I can add this part back. Interesting that we need to make sure callers of this method are setting the right things. I assume you are referencing:

```java
    ContainerPolicy.copyMapDataToRow(extraData, insertQuery.getModifyRow());
    if(this.listOrderField != null && extraData != null) {
        insertQuery.getModifyRow().put(this.listOrderField, extraData.get(this.listOrderField));
    }
```

lukasj: yes, and slightly different code bellow from other 2 callers from within the same class:
```java
ContainerPolicy.copyMapDataToRow(cp.getKeyMappingDataForWriteQuery(wrappedObject, query.getSession()), insertQuery.getModifyRow());
if(this.listOrderField != null) {
    insertQuery.getModifyRow().add(this.listOrderField, index++);
}
```

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
dazey3: We are setting the classdescriptor on the query instance with the current implementation as well, yes? Is there a document somewhere that says we should not be altering QueryManager during execution? Perhaps if ```DescriptorQueryManager.setInsertQuery``` was documented to say that it should only be called during initialization. I have seen documentation like this on other methods.

lukasj: > We are setting the classdescriptor on the query instance with the current implementation as well, yes?

With the one in this PR - yes

>  Is there a document somewhere that says we should not be altering QueryManager during execution? 

I haven't found any. What I've found is general official documentation on the purpose of DescriptorQueryManager itself, ie `The DescriptorQueryManager of each Descriptor lets you customize the query implementation that EclipseLink uses for the data source operations.` - that says nothing about mappings. It even does not mention that those queries may be customized throught the mapping. But IMO it implies that mapping code _should not_ silently modify its state which is shared across whole application. Should the query need to be cached, mapping can do that in its own private/protected field and not necessarily through the manager

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
```diff
@@ -2526,31 +2526,19 @@ public void preInsert(WriteObjectQuery query) throws DatabaseException, Optimist
...
-            insertQuery.setModifyRow(modifyRow);
+        InsertObjectQuery insertQuery = desc.getQueryManager().getInsertQuery();
+        if (insertQuery == null) {
+            insertQuery = new InsertObjectQuery();
+            insertQuery.setDescriptor(desc);
         }
+
+        // Ensure the query has been prepared.
+        insertQuery.checkPrepare(session, insertQuery.getTranslationRow());
+        insertQuery = (InsertObjectQuery)insertQuery.clone();
```
lukasj: this call to clone should be moved to 'else' branch - no need to clone newly created instance, only the one obtained from query manager

dazey3: I wrote it this way to simplify the logic since we need to call ```insertQuery.checkPrepare``` on both the new and the one set on the QueryManager before clone, yes? This is the pattern the previous code used and I was attempting to mirror it.

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/concurrency/model/User.java
```diff
@@ -22,8 +22,10 @@
 import javax.persistence.FetchType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
+import javax.persistence.Table;
 
 @Entity
+@Table(name="CONCURR_USER")
```
dazey3: Some DBs have USER as a restricted table name

dazey3:
@lukasj How do I initiate a new PR build?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateCollectionMapping.java
lukasj: I'd follow pattern from other places, like in [CommitManager](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/CommitManager.java#L213)

dazey3: So you are saying that ```insertQuery.checkPrepare``` doesn't need to be called on the new instance? Again, this is fine, but it is yet another deviation from how the code originally ran. The original failure was due to the descriptor not being set on the new instance followed by ```insertQuery.checkPrepare``` (which expected the descriptor set).
I have no problem making these changes, but my first inclination is not to rewrite everything, it's to preserve as much of the original logic as possible. If you think nixing the ```insertQuery.checkPrepare``` is fine, Ill rerun the tests and cross my fingers.

Issue #26, opened 2018-03-24
Make ValueHolderInterface api generic


vitgorbunov:

lukasj:
fixed by #27 and #393

PR #27, opened 2018-03-24
Issue #26: Make ValueHolderInterface api generic

vitgorbunov:
for #26

### DIFF
--- a/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/framework/LoadBuildSummary.java
+++ b/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/framework/LoadBuildSummary.java
@@ -42,8 +42,8 @@ public class LoadBuildSummary {
     public int fatalErrors;
     public DatabaseLogin databaseLogin;
     public Timestamp timestamp;
-    protected ValueHolderInterface summaries = new ValueHolder();
-    protected ValueHolderInterface results = new ValueHolder();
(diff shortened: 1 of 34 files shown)


vitgorbunov:
Do you mean in production code or including test as well?
Because there is at least 100 test classes that would need migration. Wouldn't it be better to migrate them one-by-one when needed (i.e. when they are being modified by some change)?

lukasj:
> Do you mean in production code or including test as well?

Each of them would be awesome! But also a lot of work. I think we can stick with direct inheritance hierarchy (including method signatures) for the beginning, see ie IndirectCollection where parents and children will use generics but this class won't. 

As for 'one-by-one' approach - I prefer to avoid doing it this way as it sometimes makes code reviews as well as cherry-picking to different branches harder to do. But that's on case-by-case basis

vitgorbunov:
Any comment on this please? Sorry for pushing, but I have some time to work on it this weekend and I'd like to get some feedback about current version before.

vitgorbunov:
Hi @lukasj , 
It's been a while since latest feedback here. Could you take a look?

dazey3:
@vitgorbunov I am fairly certain @lukasj will ask you to update the headers for each class. Take a look at other classes.

Example:
https://github.com/eclipse-ee4j/eclipselink/blob/ea5c77487819d44b821bee545651d85a8cd93535/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java#L32-L37

vitgorbunov:
Thanks!

Issue #50, opened 2018-04-03
Bug #533148: Add new persistence property to turn off SQL deferral


dazey3:
Linking to [BZ #533148](https://bugs.eclipse.org/bugs/show_bug.cgi?id=533148)

PR #51, opened 2018-04-03
[master] Bug 533148: Add new 'sql-call-deferral' property

dazey3:
for #50 

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2017 Oracle and/or its affiliates, IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2018 Oracle and/or its affiliates, IBM Corporation. All rights reserved.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
@@ -39,10 +39,12 @@
  *       - 456067 : Added support for defining query timeout units
  *     09/28/2015 - Will Dazey
  *       - 478331 : Added support for defining local or server as the default locale for obtaining timestamps
- *     12/03/2015-2.6 Dalia Abo Sheasha
- *       - 483582: Add the javax.persistence.sharedCache.mode property
- *     09/14/2017-2.6 Will Dazey
- *       - 522312: Add the eclipselink.sequencing.start-sequence-at-nextval property
+ *     12/03/2015 - 2.6 Dalia Abo Sheasha
+ *       - 483582 : Add the javax.persistence.sharedCache.mode property
+ *     09/14/2017 - 2.6 Will Dazey
+ *       - 522312 : Add the eclipselink.sequencing.start-sequence-at-nextval property
+ *     04/11/2018 - Will Dazey
+ *       - 533148 : Add the eclipselink.jpa.sql-call-deferral property
  ******************************************************************************/
 package org.eclipse.persistence.config;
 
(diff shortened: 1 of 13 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=533148

dazey3:
I added a couple whitespace changes and java changes to cleanup warnings in the jse test bucket as well.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
```diff
@@ -3798,6 +3798,19 @@
      */
     public static final String USE_LOCAL_TIMESTAMP = "eclipselink.locking.timestamp.local." + PersistenceUnitProperties.DEFAULT;
 
+    /**
+     * The "<code>eclipselink.jpa.sqlcall.deferral.default</code>" property defines if SQL calls should be deferred to end of
+     * transaction by default or not. When setting this property to <code>false</code>, the application assumes the responsibility 
+     * of ordering the SQL statements and must therefore be aware of any interdependencies between entities. 
+     * <p>
+     * <b>Allowed Values</b> (case sensitive String)<b>:</b>
```
lukasj: the code in entitymanagersetupimpl does case in-sensitive comparison, so it does not match what is said here - you wanted to say `case insensitive String` here perhaps?

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/property/TestSQLCallDeferall.java
```diff
@@ -0,0 +1,58 @@
...
+        Parent parent = new Parent();
+        em.persist(parent);
+
+        Child child = new Child();
+        child.setParent(parent);
+        em.persist(child);
+
+        // expects: INSERT order: AbstractParent, Parent, Child
+        // ECL default: INSERT order: AbstractParent, Child, Parent(deferred)
+        em.getTransaction().commit();
```
lukasj: I do not see any verification here. Are you looking for something like `org.eclipse.persistence.testing.framework.QuerySQLTracker` (there are more variants of this in tests)?

dazey3: The verification is that an SQL exception will be thrown on commit unless the property is set. If you look at the ```@JoinColumn``` on the Child entity, I am setting a custom foreignKey. Since EclipseLink doesn't process this FK, it would normally try to reorder INSERTs. However, with this new property, the reordering will not occur.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
dazey3: Yep, thx for catching this.

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/property/TestSQLCallDeferall.java
dazey3: I looked into using ```org.eclipse.persistence.sessions.SessionEventAdapter```, which is how ```org.eclipse.persistence.testing.framework.QuerySQLTracker``` seems to works, but there may be a bug in the listener code. If you look at ```AbstractSession.executeQuery(...)```, you can see the eventManager is logging the query execution. However, if you look at ```AbstractSession.executeCall()```, there is no eventManager calls. 

What this means is that only 2 queries are logged in the QuerySQLTracker: AbstractParent and Child. But the one call for Parent is not even logged. This may be a bug that needs to be fixed; I assumed ```SessionEventAdapter``` should log ALL sql, not just queries but calls as well. But in the mean time, I believe this test accurately fails when the property isnt set and passes if the property is deferring calls.

Issue #67, opened 2018-04-13
Bug #532160: eclipselink OraclePlatform QueryHint FIRST_ROWS does not use max parameter


dazey3:
Linking to [BZ #532160](https://bugs.eclipse.org/bugs/show_bug.cgi?id=532160)

PR #68, opened 2018-04-13
[master] Bug 532160: OraclePlatform QueryHint FIRST_ROWS should use max parameter

dazey3:
for #67 


dazey3:
I need to create a test for this still.

PR #69, opened 2018-04-16
Bug 420746 - Unmarshal of JAXBElement field annotated with @XmlElementRef nil not unmarshalling xsi:nil="true" correctly.

rfelcman:
Bug 420746 - Unmarshal of JAXBElement field annotated with @XmlElementRef nil not unmarshalling xsi:nil="true" correctly.
	
MOXy unmarshaller incorrectly unmarshal element declared with @XmlElementRef if attribute (namespace declaration) xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" is positioned outside element with xsi:nil="true" attribute.
There are no other null policies applied, than xsi:nil="true" in the source document. After patch it will set as a field value instance of JAXBElement (from ObjectFactory) with null value instead of null. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=420746 .

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -9,14 +9,11 @@
     <option name="pathToConfigFileJava" value="$PROJECT_DIR$/project-admin/EclipseLink-Eclipse-Format.xml" />
     <option name="selectedJavaProfile" value="EclipseLink" />
   </component>
-  <component name="EntryPointsManager">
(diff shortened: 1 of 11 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=420746

PR #80, opened 2018-04-25
Issue #285: Avoid ClassCastException and IllegalArgumentException in cloneFor implementations

jhanders34:
VectorContainerPolicy is used for all List types.  Likely the Object
passed to cloneFor will not be a Vector, but rather an ArrayList,
LinkedList or some other List derivative.  Getting a ClassCastException
is expected to be the normal code path so doing an instanceof check to
avoid the exception.  instanceof will perform better than the
ClassCastException since you need to do a stack walk in order to
generate the stack trace of the exception.  In general Exceptions should
not be a normal decision mechanism if possible.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/InterfaceContainerPolicy.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/InterfaceContainerPolicy.java
@@ -15,6 +15,9 @@ package org.eclipse.persistence.internal.queries;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
 import java.lang.reflect.*;
 import org.eclipse.persistence.descriptors.changetracking.CollectionChangeEvent;
 import org.eclipse.persistence.exceptions.*;
@@ -99,13 +102,16 @@ public abstract class InterfaceContainerPolicy extends ContainerPolicy {
             return null;
         }
 
-        try {
-            return invokeCloneMethodOn(getCloneMethod(), container);
-        } catch (IllegalArgumentException ex) {
+        Method cloneMethod;
+        Class javaClass = container.getClass();
+        if (javaClass == getContainerClass()) {
+            cloneMethod = getCloneMethod();
+        } else {
             // container may be a superclass of the concrete container class
             // so we have to use the right clone method...
-            return invokeCloneMethodOn(getCloneMethod(container.getClass()), container);
+            cloneMethod = getCloneMethod(javaClass);
         }
+        return invokeCloneMethodOn(cloneMethod, container);
     }
 
     /**
@@ -169,26 +175,77 @@ public abstract class InterfaceContainerPolicy extends ContainerPolicy {
         return cloneMethod;
     }
 
+    private static final class ClassWeakReference extends WeakReference<Class> {
+        private final int hash;
+
+        ClassWeakReference(Class referent) {
+            super(referent);
+            hash = referent.hashCode();
+        }
+
+        ClassWeakReference(Class referent, ReferenceQueue<Class> referenceQueue) {
+            super(referent, referenceQueue);
+            hash = referent.hashCode();
+        }
+
+        @Override
+        public int hashCode() {
+            return hash;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+
+            if (obj instanceof ClassWeakReference) {
+                return get() == ((ClassWeakReference) obj).get();
+            }
+
+            return false;
+        }
+
+        @Override
+        public String toString() {
+            Class referent = get();
+            return new StringBuilder("ClassWeakReference: ").append(referent == null ? null : referent).toString();
+        }
+    }
+
+    private static final ReferenceQueue<Class> refQueue = new ReferenceQueue<>();
+    private static final ConcurrentHashMap<ClassWeakReference, Method> cloneMethodCache = new ConcurrentHashMap<>();
+    
     /**
      * INTERNAL:
      * Return the 'clone()' Method for the specified class.
      * Return null if the method does not exist anywhere in the hierarchy
      */
     protected Method getCloneMethod(Class javaClass) {
+        for (Object key; (key = refQueue.poll()) != null;) {
+            cloneMethodCache.remove(key);
+        }
+        Method cloneMethod = cloneMethodCache.get(new ClassWeakReference(javaClass));
+        if (cloneMethod != null) {
+            return cloneMethod;
+        }
         try {
             // This must not be set "accessible" - clone() must be public, and some JVM's do not allow access to JDK classes.
             if (PrivilegedAccessHelper.shouldUsePrivilegedAccess()){
                 try {
-                    return AccessController.doPrivileged(new PrivilegedGetMethod(javaClass, "clone", (Class[])null, false));
+                    cloneMethod = AccessController.doPrivileged(new PrivilegedGetMethod(javaClass, "clone", (Class[])null, false));
                 } catch (PrivilegedActionException exception) {
                     throw QueryException.methodDoesNotExistInContainerClass("clone", javaClass);
                 }
             } else {
-                return PrivilegedAccessHelper.getMethod(javaClass, "clone", (Class[])null, false);
+                cloneMethod = PrivilegedAccessHelper.getMethod(javaClass, "clone", (Class[])null, false);
             }
         } catch (NoSuchMethodException ex) {
             throw QueryException.methodDoesNotExistInContainerClass("clone", javaClass);
         }
+
+        cloneMethodCache.put(new ClassWeakReference(javaClass, refQueue), cloneMethod);
+        return cloneMethod;
     }
 
     /**
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ArrayListContainerPolicy.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ArrayListContainerPolicy.java
@@ -55,12 +55,12 @@ public class ArrayListContainerPolicy extends ListContainerPolicy {
         if (container == null) {
             return null;
         }
-        try {
+        if (container.getClass() == ArrayList.class) {
             return ((ArrayList)container).clone();
-        } catch (Exception notArrayList) {
-            // Could potentially be another Collection type as well.
-            return new ArrayList((Collection)container);
         }
+
+        // Could potentially be another Collection type as well.
+        return new ArrayList((Collection) container);
     }
 
     /**
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/IndirectListContainerPolicy.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/IndirectListContainerPolicy.java
@@ -53,12 +53,12 @@ public class IndirectListContainerPolicy extends ListContainerPolicy {
             return null;
         }
         // Use Vector as new objects can have a Vector.
-        try {
+        if (container.getClass() == Vector.class) {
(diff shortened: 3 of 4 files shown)


jhanders34:
@dazey3 Here is the pull request for the correct repository again.

jhanders34:
Lukas, 

Thanks for the feedback.  I have not opened a problem/bug for this PR yet.  I am new to contributing to this project and was going to open a bug/problem if the community thought one was necessary.  It is a small change so I wasn't sure if one would be required for it.

See [setIndirectionPolicy in MappingAccessor](https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/mappings/MappingAccessor.java#L277).  For a OnetoMany relationship where the field is defined as a List, CollectionMapping.useCollectionClass is called with Vector.class.  This results in a VectorContainerPolicy being used for all List type fields for OneToMany relationships.

The comment in MappingAccessor states:

>                 // This will cause a ListContainerPolicy type to be used or
                // OrderedListContainerPolicy if ordering is specified.

This is not the case though.  If you change it to a ListContainerPolicy, then you will get other errors when trying to do clone operations.  If you change to use the ListContainerPolicy, it will either look up the clone method every time with reflection (which is bad for performance) or it will need to be changed to keep a mapping of List implementations to clone methods so it knows which clone method to call for which List implementation sent to the cloneFor method.

lukasj:
ListContainerPolicy already remembers the right 'clone' method to use for given class, so the clone lookup is called at most once per policy instance. So to me it looks like that the bug is that wrong  policy is chosen in the MappingAccessor

jhanders34:
Yes it remembers the right 'clone' method IF a concrete class is what is passed to the ListContainerPolicy.  If the List class itself is passed, it will not be able to handle it.  If the toMany relationship defines the field as a List (which is common), then there would need to be a change to handle the List class and need to handle all List implementations that could be passed for that field.  Just because it is an ArrayList one time doesn't mean that it will not be a LinkedList or another implementation of List later in a badly coded class.  BTW, the Vector class has been being passed in MappingAccessor for the last 10 years.  It appears to not be a new change.  If you have a suggestion about how to handle this issue better, I gladly will pursue it.  FYI, I did update the comment in the commit to reflect what I stated above.

lukasj:
What am I missing in (InterfaceContainerPolicy.cloneFor)[https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/InterfaceContainerPolicy.java#L97]?

jhanders34:
getCloneMethod will get an exception if javaClass is List.class since there is no clone method on List.  NoSuchMethodException will happen and then a QueryException will result.  If javaClass is passed the actual class of the Object passed into cloneFor, then it will set the clone method to that class's clone method, but if another implementation of List comes in later then you can run into additional issues because the wrong clone method will be used for the second class.

dazey3:
Hmm, I'm coming into this conversion a bit behind so forgive me if I am off base. 

Lets assume a mapping like this:

```java:
@OneToMany
private List list = new ArrayList();
```

The story starts here:
https://github.com/eclipse-ee4j/eclipselink/blob/5b57bf9c9e74d4d4521f996961b47c45dfc25226/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/mappings/MappingAccessor.java#L2141-L2158

Here in setIndirectionPolicy, we will see that the rawClass type is a List, so we call ```mapping.useCollectionClass(java.util.Vector.class);```

Moving forward to the CollectionMapping impl:
https://github.com/eclipse-ee4j/eclipselink/blob/5b57bf9c9e74d4d4521f996961b47c45dfc25226/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/CollectionMapping.java#L2279-L2283
we see that the ContainerPolicy builds for us the correct impl based on the concreteClass (in this case Vector.class)

This of course leads us to ContainerPolicy where the decision is made:
https://github.com/eclipse-ee4j/eclipselink/blob/5b57bf9c9e74d4d4521f996961b47c45dfc25226/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ContainerPolicy.java#L337-L349
Since the concreteclass type is Vector.class, we return a ``VectorContainerPolicy``` and set that as the policy for the mapping.

Fast forward to clone() time. In order to clone the mapping, we need to use the ContainerPolicy which we set as a VectorContainerPolicy:

https://github.com/eclipse-ee4j/eclipselink/blob/5b57bf9c9e74d4d4521f996961b47c45dfc25226/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/VectorContainerPolicy.java#L53-L63

Inside the cloneFor, the type will be whatever the field is instantiated to (in this case ```ArrayList```). This leads to a ClassCastException since ArrayList cannot be cast to a Vector. Sure, we then catch the exception and new up a Vector, but I believe the argument is that catching exceptions like this should not be our normal code path. The situation where this code path does NOT hit the CCE catch, is when the collection type is legitimately instantiated as a Vector.

lukasj:
soo, it shouldn't be hard to create a test for this, right? What about dynamic/static weaving (where correct type could be [eventually recognized](https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/weaving/TransformerFactory.java#L425))? How does that affect this scenario?

jhanders34:
Yeah I have an easy test case for this where I do a print stack trace when this happens.

lukasj:
@jhanders34 can you share the test, please?

jhanders34:
I apologize for the delay, but other responsibilities have kept from from working on this:  Here is some pseudo code that shows this particular problem:

```            Department dept = new Department();
            dept.setDeptno(1);
            dept.setName("dept");
            dept.setEmployees(new ArrayListExt());
            dept.getEmployees();
            /*ArrayList<Employee> al = new ArrayList<>();
            al.add(employee2);
            dept.setEmployees(al);*/
            //Thread.dumpStack();
            em.persist(dept);

            env.commitTransactionAndClear(em);
            Assert.assertTrue(ArrayListExt.cloneCalled);
```

```    private static final class ArrayListExt extends ArrayList {

        private static final long serialVersionUID = 7644682526364234652L;

        public static boolean cloneCalled = false;
        
        @Override
        public ArrayListExt clone() {
            cloneCalled = true;
            return (ArrayListExt) super.clone();
        }
    }
```

The Department employee type is also changed to List instead of Collection for the employees.  With that code you will see that the clone is not called on the ArrayListExt.

jhanders34:
I put in a new change that I think addresses your concerns.  I would like to update the getCloneMethod to use a cache so we aren't doing reflection every time to get the clone method when it isn't the actual concrete class that is provided since it won't be in these cases.  I would use a ConcurrentHashMap with weak reference keys of Classes pointing to the cloneMethod.

jhanders34:
clone method class added to the commit.

jhanders34:
@lukasj can you take a look at this PR again?  Do you want me to open a bug for the change?

jhanders34:
@lukasj anything I can do to speed this getting in?

jhanders34:
@lukasj.  Any updates you can provide?

jhanders34:
@lukasj and @dazey3 I have updated the pull request again to make the code similar to what it was before and not use super.cloneFor() to call the clone method of the Collection type that is passed to the cloneFor method in the ArrayList and Vector ContainerPolicy classes.. It now does what it did before and just makes a new ArrayList or Vector if it isn't an ArrayList or Vector.  The super cloneFor method is updated though in case in the future you want to call super.cloneFor or use the InterfaceContainerPolicy for Collection, List, Set, Map, etc interfaces being declared for ToMany relationships.  It now will handle that the passed in collection type may be different than the cached collection type or interface and will cache the clone Method object to avoid having to look it up on each call to cloneFor in this scenario.

With the code changes, there is not a way to test for this particular change because it is all internal to the cloneFor methods.  If I left it to call the super.cloneFor then I could verify that the clone method was called on the passed in collection type if that collection type was Cloneable similar to the comment I posted above.  From our testing with a Java EE benchmark (not a JPA specific benchmark), we were seeing 1 to 2% reduction in CPU because there isn't an exception generated on each commit of an Entity with a OneToMany relationship that used a List as its type and the impl was an ArrayList.  With a JPA specific benchmark the reduction in CPU cost probably is much higher.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/InterfaceContainerPolicy.java
```diff
@@ -169,26 +175,77 @@ public Method getCloneMethod() {
         return cloneMethod;
     }
 
+    private static final class ClassWeakReference extends WeakReference<Class> {
```
dazey3: Is there a specific reason we can't just use ```WeakReference<Class>``` instead of creating this new class? Im guessing it's something to do with overriding equals, but the Object.equals implementation should work, right?

jhanders34: The equals on WeakReference is Object.equals which is an == check.  That will never happen which means we will never find anything in the ConcurrentHashMap.

jhanders34:
As you can see I have attached an issue to this pull request now.  The issue has additional information about the impact of this change to performance.

dazey3:
@lukasj It seems we don't run any performance tests as part of our Travis build. Should we? Do we know how these tests work? I am finding 'eclipselink.jpa.test' has a decent number of performance tests, but no one is calling those targets (test-performance test-performance2 test-performance2-nocache ...) However, I am also seeing the base antbuild.xml contains 'test-perf' as well as 'test-jpa-perf', which is for the 'eclipselink.perf.test' project?? Why do we have performance tests all over the place? Have you run these before?

lukasj:
@dazey3 problem with existing perf tests is that they require data storage and are still heavily tight to Oracle DB; basically each run stores some data there and there's some UI app (testbrowser?) allowing one to query the DB for those data. There were few attempts to migrate it somehow to some mysql (which we could get on public infra, I believe) but all those working on that in the past are no longer in our team. One possibly problematic part I can see with travis is how to get data from it to some external storage and what kind of storage it should be - for these tests to be useful, we should be able to see the trend over the time

dazey3:
Interesting. What if the results were just stored in the DB the tests are running against (instead of a specific OracleDB)? If you run the tests against MySQL, then those results are stored in the MySQL DB for future reference. Assuming the DBs aren't cleaned, then this should be safe and provide accurate results, right?

One things I am noticing is a bunch of hardcoded DB urls all over the place. Example:
https://github.com/eclipse-ee4j/eclipselink/blob/17a37026e4b2e3cccc07e2beda7d926552025956/jpa/eclipselink.jpa.test/resource/eclipselinkorm/eclipselink-xml-extended-composite-advanced-model-member_2/persistence.xml#L23-L27

It seems so weird to me that these tests are using specific URLs. I can appreciate that these tests were not ever meant to be run by general devs, but it seems odd that we have these URLs checked in...

lukasj:
> Interesting. What if the results were just stored in the DB the tests are running against (instead of a specific OracleDB)? If you run the tests against MySQL, then those results are stored in the MySQL DB for future reference.

that was also possible the idea and main motivation I kept all perf tests related properties in place while updating build scripts in tests, it starts in [LoadBuildSystem](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/framework/LoadBuildSystem.java) and [LoadBuildProject](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/framework/LoadBuildProject.java)


> One things I am noticing is a bunch of hardcoded DB urls all over the place. [..] but it seems odd that we have these URLs checked in...

absolutely agree and generally speaking only public URLs belong to public place. Anyway, in some cases these URLs are meaningless as tests may be overwritten by custom Properties object sent to the provider/factory in tests.

dazey3:
@lukasj You have any objections to this?

lukasj:
Thanks! Especially for the effort you invested into proving that the change is correct.

jhanders34:
for #285

PR #81, opened 2018-04-26
Bug 503015 - NPE in XMLJavaTypeConverter.java

rfelcman:
Bug 503015 - NPE in XMLJavaTypeConverter.java

Possible source of NPE if marshaller, unmarshaller variable will be null.
This case is not covered by special test case. It's not usual to create and use instance of org.eclipse.persistence.internal.jaxb.XMLJavaTypeConverter directly. There are tests in  org.eclipse.persistence.testing.jaxb.xmladapter package.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=503015 .


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=503015

PR #84, opened 2018-04-26
Bug 533861 - JSON unmarshaller incorrectly parse nil elements

rfelcman:
Bug 533861 - JSON unmarshaller incorrectly parse nil elements

JSON unmarshaller incorrectly parse nil elements in JSON. It parse them like XML xsi:nil="true" control attribute.
New condition in org.eclipse.persistence.internal.oxm.record.UnmarshalRecordImpl to evaluate nil as a control attribute only when input is XML document.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=533861 .


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=533861

PR #85, opened 2018-04-26
Bug 533874 - Attribute and Element with same name results in element not appearing on parse

rfelcman:
Bug 533874 - Attribute and Element with same name results in element not appearing on parse

MOXy doesn't unmarshall XML/JSON document correctly if attribute and element with same name is at the same level.
Input XML Example: ...<employee name="john.smith"><name>...
Input JSON Example: ..."employee": {"@name": "john.smith","name": [...
leads into wrong mapping to entity instance.
This patch repairs mapper selection from HashTable to correctly identify attribute and element.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=533874 .


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=533874

PR #86, opened 2018-05-02
Bug 484561 - JAXB changes null to "" - when @XmlValue is used for JSON

rfelcman:
Bug 484561 - JAXB changes null to "" - when @XmlValue is used for JSON

MOXy JSON unmarshaller incorrectly unmarshalls null into field marked with @XmlValue annotation and when @XmlVariableNode is used.
Instead of null it returns empty string.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=484561 .


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484561

PR #94, opened 2018-05-10
[2.7] Eclipselink#93: Resolve issue with JPA Versoning with java.sql.Timestamp

### LINKED ISSUE #93: Bug with bulk update processing involving version field update parameter
The JPA 2.1 Specification states:

> **4.10 Bulk Update and Delete Operations**
> ...
> Bulk update maps directly to a database update operation, bypassing optimistic locking checks. Portable applications must manually update the value of the version column, if desired, and/or manually validate the value of the version column.

However, there is a problem with this when attempting to execute such a bulk update in Eclipselink.  In certain scenarios, an UPDATE block that targets the version field is ignored, and instead replaces the Parameter with a Constant (generated at the time the query is generated, which gets cached and fouls up successive executions of the same query once it gets into the query cache, too.)  When this happens, the parameter passed into the query is ignored.

Consider the following entity:
``` java
@Entity
public class TemporalVersionedEntity {
	@Id private long id;
	@Version private java.sql.Timestamp updatetimestamp;
...
}
```

and the following execution logic:

``` java
private final static String qStr1 = "UPDATE TemporalVersionedEntity  SET updatetimestamp = ?3 " +
     "WHERE id = ?1 AND updatetimestamp = ?2";
em.getTransaction().begin();
final Query q = em.createQuery(qStr1);
q.setParameter(1, theId);
q.setParameter(2, theOldTimestamp);
q.setParameter(3, new java.sql.Timestamp(System.currentTimeMillis());
final int count = q.executeUpdate();
```

However, when you run the query, you don't get the time put into parameter #3, you get a database time generated by Eclipselink when first processing the Query (in this case, the time returned from the database).  Here's an example output from my unit test (which the above example is based on):

```
Running testTemporalVersionField1 ...
Bulk Update Parameter 3 = **2018-05-10 14:16:33.533**
[EL Fine]: sql: 2018-05-10 14:16:34.035--ClientSession(831654622)--Connection(319558327)--Thread(Thread[main,5,main])--UPDATE TEMPORALVERSIONEDENTITY SET UPDATETIMESTAMP = ? WHERE ((ID = ?) AND (UPDATETIMESTAMP = ?))
	bind => [**2018-05-10 14:16:33.02**, 1, 2018-05-10 14:16:32.365]
```

As you can see, the value that is actually sent to the database is not only the value not specified by the parameter, but from a time moments before parameter #3 was generated (using DerbyDB on the same machine, so system clock is the same.)

The crazy thing that confounded me before I figured out what was going on was if I changed the version field name from "updatetimestamp" to plain old "version", the test started working!  This was particularly confounding since a @Version field can be named anything.  However, it turned out to be an issue with iterating through the order of entries in a HashMap, and the keys happened to evaluate in a consistent order between running against both entities.  Here is the affected code within org.eclipse.persistence.internal.queries.ExpressionQueryMechanism.prepareUpdateAll():

``` java
    public void prepareUpdateAll() {
        ExpressionBuilder builder = ((UpdateAllQuery)getQuery()).getExpressionBuilder();
        HashMap updateClauses = ((UpdateAllQuery)getQuery()).getUpdateClauses();

        boolean updateClausesHasBeenCloned = false;
        // Add a statement to update the optimistic locking field if their is one.
        OptimisticLockingPolicy policy = getDescriptor().getOptimisticLockingPolicy();
        if (policy != null) {
            if(policy.getWriteLockField() != null) {
                Expression writeLock = builder.getField(policy.getWriteLockField());
                Expression writeLockUpdateExpression = policy.getWriteLockUpdateExpression(builder, getQuery().getSession());
                if (writeLockUpdateExpression != null) {
                    // clone it to keep user's original data intact
                    updateClauses = (HashMap)updateClauses.clone();
                    updateClausesHasBeenCloned = true;
                    updateClauses.put(writeLock, writeLockUpdateExpression);
                }
            }
        }
...
        Iterator it = updateClauses.entrySet().iterator();
        while(it.hasNext()) {
            Map.Entry entry = (Map.Entry)it.next();

            Object fieldObject = entry.getKey();
...
```

Here, we get the OptimisticLockingPolicy (OLP) for the entity in question, and since the entity has a @Version field it returns an instance of TimestampLockingPolicy.  It enters the nested if-block, producing a FieldExpression value for `Expression writeLock` and a ConstantExpression value for `Expression writeLockUpdateExpression`.  These are then added to the `updateClauses` hashMap (whether they're needed or not), which already includes the QueryKeyExpression/ParameterExpression for Parameter 3 (the parameter updating the version column in the update clause.) 

Here's a snapshot of the updateClauses HashMap after the insertion (for the failing scenario):

```
updateClauses	HashMap<K,V>  (id=125)	
	[0]	HashMap$Node<K,V>  (id=128)	
		key	QueryKeyExpression  (id=118) = Query Key updatetimestamp
		value	ParameterExpression  (id=119) = Parameter 3
	[1]	HashMap$Node<K,V>  (id=130)	
		key	FieldExpression  (id=95)  = Field TEMPORALVERSIONEDENTITY.UPDATETIMESTAMP
		value	ConstantExpression  (id=99) = Constant 2018-05-10 14:30:02.07	
```

Which is later iterated through starting on line 1871.  The order actually matters in whether the update works or not -- in the failing case the QueryExpression is iterated first (and lost when iterating the second entry in the map)  In the "passing case" where the version column is "version" the FieldExpression actually is inserted before the existing entry instead of after, allowing it to work as expected by happenstance.

jgrassel:
fixes #93 
Signed-off-by: Joe Grassel <jgrassel@us.ibm.com>

My solution to resolving the problem is to first check if the database field referenced by the writeLock expression is already part of an existing expression in the updateClause.  For example, during a debug session, I noticed while examining the memory state of the objects available during the method invocation:

```
writeLock	FieldExpression  (id=90)	
	field	DatabaseField  (id=87)	
updateClauses	HashMap<K,V>  (id=77)	
	[0]	HashMap$Node<K,V>  (id=147)	
		key	QueryKeyExpression  (id=152)	
			baseExpression	ExpressionBuilder  (id=71)	
				derivedFields	ArrayList<E>  (id=168)	
					[0]	FieldExpression  (id=90)
						field	DatabaseField  (id=87)	
```	

And I thought a potential solution would be to identify the DatabaseField that is member of the writeLock FieldExpression, and look for it in any derivedFields in the updateClauses.  In practice, this worked (the tests passed), but I would like to hear from the community on whether this is a solid fix, or if there is a better approach to resolving the issue.

The approach I took attempts to prevent the writeLock FieldExpression from being admitted to the updateClauses map, rather than trying to determine that its removal is necessary in the subsequent processing that iterates through the contents of updateClauses.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2018 Oracle, IBM and/or its affiliates. All rights reserved.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
@@ -12,6 +12,8 @@
  *     Thomas Spiegl - fix for bug 324406
  *     10/15/2010-2.2 Guy Pelletier
  *       - 322008: Improve usability of additional criteria applied to queries at the session/EM
+ *     05/10/2018-2.7 Joe Grassel
+ *       - Github#93: Bug with bulk update processing involving version field update parameter
  ******************************************************************************/
 package org.eclipse.persistence.internal.queries;
 
@@ -1833,6 +1835,32 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
         }
     }
 
+    private boolean isFieldInUpdate(Expression writeLock, HashMap updateClauses) {
+    	if (!(writeLock instanceof FieldExpression)) {
+    		return false;
+    	}
+    	
+    	final FieldExpression fe = (FieldExpression) writeLock;
+    	final DatabaseField targetField = fe.getField();
+    	
+    	final Set keys = updateClauses.keySet();
+    	for (Object key : keys) {
+    		if (!(key instanceof QueryKeyExpression)) {
+    			continue;
+    		}
+    		
+    		QueryKeyExpression qke = (QueryKeyExpression) key;
+    		ExpressionBuilder eb = qke.getBuilder();
+    		Expression ex = eb.existingDerivedField(targetField);
+    		if (ex != null) {
+    			// Found an expression in the update clause that touches this database field.
+    			return true;
+    		}
+    	}
+    	
+    	return false;
+    }
+    
     /**
      * Pre-build the SQL statement from the expressions.
      */
@@ -1846,12 +1874,18 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
         if (policy != null) {
             if(policy.getWriteLockField() != null) {
                 Expression writeLock = builder.getField(policy.getWriteLockField());
-                Expression writeLockUpdateExpression = policy.getWriteLockUpdateExpression(builder, getQuery().getSession());
-                if (writeLockUpdateExpression != null) {
-                    // clone it to keep user's original data intact
-                    updateClauses = (HashMap)updateClauses.clone();
-                    updateClausesHasBeenCloned = true;
-                    updateClauses.put(writeLock, writeLockUpdateExpression);
+                // Note: The spec allows for version fields to be updated in bulk updates.  Adding the writeLockUpdateExpression when there is already
+                // a QueryKeyExpression associated with the version column will result in a scenario where one wins out by virtue of order of iteration
+                // of updateClauses's entrySet.  So we need to check the updateClause to see if the database fields in the writeLock expression are
+                // already targeted for update.
+                if (!isFieldInUpdate(writeLock, updateClauses)) {
+                	Expression writeLockUpdateExpression = policy.getWriteLockUpdateExpression(builder, getQuery().getSession());
+                    if (writeLockUpdateExpression != null) {
+                         // clone it to keep user's original data intact
+                        updateClauses = (HashMap)updateClauses.clone();
+                        updateClausesHasBeenCloned = true;
+                        updateClauses.put(writeLock, writeLockUpdateExpression);
+                    }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


jgrassel:
Interesting.  It looks like I am running afoul of the JTAOptimisticLockExceptionTest test case:

```
Lock exception should have been thrown 

junit.framework.AssertionFailedError: Lock exception should have been thrown
at org.eclipse.persistence.testing.tests.jpa.jpql.JUnitJPQLValidationTestSuite.JTAOptimisticLockExceptionTest(JUnitJPQLValidationTestSuite.java:956)
at org.eclipse.persistence.testing.framework.junit.JUnitTestCase.runBare(JUnitTestCase.java:655)
at org.eclipse.persistence.testing.framework.TestExecutor.execute(TestExecutor.java:248)
at org.eclipse.persistence.testing.framework.TestModel.execute(TestModel.java:211)
at org.eclipse.persistence.testing.framework.TestCollection.run(TestCollection.java:313)
```

The test case is as follows:

```
        EntityManager em = createEntityManager();
        try {
            beginTransaction(em);
            try {
                Employee emp = (Employee) em.createQuery("SELECT OBJECT(emp) FROM Employee emp WHERE emp.firstName='Bob' ").getSingleResult();
                em.createQuery("Update Employee set lastName = 'test-bad' WHERE firstName='Bob' ").executeUpdate();
                emp.setLastName("test");
                commitTransaction(em);
            } catch (RuntimeException ex) {
                // Truncated, OptimisticLockException is thrown by the test and is looked for here
            }
            fail("Lock exception should have been thrown");
```

However, the JPA 2.1 Specification states:

> **4.10 Bulk Update and Delete Operations**
> ...
> Bulk update maps directly to a database update operation, bypassing optimistic locking checks. Portable applications must manually update the value of the version column, if desired, and/or manually validate the value of the version column.

To me, this suggests that the above test case is invalid, because it is testing that the version column will be automatically updated by the bulk update.

jgrassel:
Well, the initial strategy was not sound.  But identified a simple change to [ExpressionQueryMechanism](https://github.com/jgrassel/eclipselink/blob/TimestampVersionBug_2.7/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java#L1838) which looks more promising.

```java
    		QueryKeyExpression qke = (QueryKeyExpression) key;   		
    		DatabaseField qkField = getDescriptor().getObjectBuilder().getFieldForQueryKeyName(qke.getName());
    		if (qkField == targetField) {
    			return true;
    		}
```
Instead of navigating deeply through the QueryKeyExpression data structures for a DatabaseField, I modified the algorithm to invoke the objectBuilder's getFieldForQueryKeyName() method (which is done later on line 1932 for the same effect(), which seems to be a safer alternative.  It passed my TestVersioning bucket (which has been updated for more tests), waiting to see the Travis results for the rest.

PR #99, opened 2018-05-18
Update core.test script to allow easier testing against modules

lukasj:
This also removes split package between core and core.test and updates travis to use Ant 1.10.2 for the build

### DIFF
--- a/foundation/eclipselink.core.test/antbuild.xml
+++ b/foundation/eclipselink.core.test/antbuild.xml
@@ -37,82 +37,400 @@
         - clean   : cleans up all generated files and dirs
         - compile : compiles all tests
         - package : packages test jar
-        - process.resource : translates login info in resources into run directory
+        - process-resource : translates login info in resources into run directory
         - test : runs LRG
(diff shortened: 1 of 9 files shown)


foundation/eclipselink.core.test/antbuild.xml
```diff
@@ -37,267 +37,449 @@
...
             <pathelement path="${extensions.depend.dir}/${jgroups.jar}"/>
         </path>
-        <path id="core_test.run.path">
+        <path id="core_test.run.path" cache="true">
             <path refid="core_test.compile.path"/>
             <path refid="core_test.runtime.only.path"/>
         </path>
     </target>
 
+    <target name="init-test-macro" depends="init">
```
Tomas-Kraus: Is it really necessary to put target around macro? Why nto to call it directly?

lukasj: Proper dependecies is one point. Other is that for 'clean build' one does not need the macro at all, so this may also improve efficiency of the build in terms of time and memory.

Issue #63, opened 2018-04-09
Recursive annotations triggers a StackOverflowError


marvelous:
When an annotation class annotated with itself gets scanned by EclipseLink (`<exclude-unlisted-classes>false</exclude-unlisted-classes>`), it enters into a recursive loop which triggers a `StackOverflowError` where `org.eclipse.persistence.internal.jpa.metadata.accessors.objects.MetadataAnnotatedElement.getAnnotation(MetadataAnnotatedElement.java:215)` calls itself.

```
@Self
public @interface Self {}
```

It happens in payara-micro 5 (EclipseLink 2.7.1.qualifier), but it didn't happen in payara-micro 4. A sample WAR can be found in payara/Payara#2608.

marvelous:
Here is a Java SE sample: [sscce.zip](https://github.com/eclipse-ee4j/eclipselink/files/1893823/sscce.zip)

rieonke:
so can we just ignore the annotations which self annotated?
I ignored them at https://github.com/eclipse-ee4j/eclipselink/blob/a985f91916b85540209553514cafd9f88718c8c6/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java#L373-L385

```java
        @Override
        public void visitEnd() {
            if (this.element != null && this.annotation != null && !this.annotation.getName().equals(this.element.getName())) {
                if (this.annotation.isMeta()) {
                    this.element.addMetaAnnotation(this.annotation);
                } else {
                    this.element.addAnnotation(this.annotation);
                }
            }
        }
```
then it works fine.

@lukasj

marvelous:
I doubt it will be enough to catch mutual recursion like

```
@Self2
public @interface Self1 {}
```
```
@Self1
public @interface Self2 {}
```

cristhiank:
I created a bug in Bugzilla also because it's not clear if the project uses Github as the bug tracker: https://bugs.eclipse.org/bugs/show_bug.cgi?id=535250

Tomas-Kraus:
Well, that's a question. There are 2 options how to handle it:
1. break the cycle while reading classes (fix proposed here)
2. break the cycle in the MetadataAnnotation getAnnotation(String annotation) method which is causing stack overflow

(1.) means that the last class of the cycle will be missing in the metamodel and I think that it's not good even when we know that something is going wrong here
[...]

PR #108, opened 2018-05-27
proposed solution for #63

cristhiank:
Proposed solution for #63  and https://github.com/payara/Payara/issues/2608
see https://github.com/eclipse-ee4j/eclipselink/issues/63

It includes an explicit exclusion for kotlin annotations that are widely used when developing in kotlin.

Included test app written in Kotlin

[eclipselink-bug-kotlin.sources.zip](https://github.com/eclipse-ee4j/eclipselink/files/2042297/eclipselink-bug-kotlin.sources.zip)

[eclipselink-bug-kotlin-1.0-SNAPSHOT.war.zip](https://github.com/eclipse-ee4j/eclipselink/files/2042298/eclipselink-bug-kotlin-1.0-SNAPSHOT.war.zip) Remove .zip extension to get the war


jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java
```diff
@@ -372,14 +376,28 @@ public AnnotationVisitor visitArray(String name) {
...
+                if ( this.annotation.isMeta() ) {
+                    this.element.addMetaAnnotation( this.annotation );
                 } else {
-                    this.element.addAnnotation(this.annotation);
+                    this.element.addAnnotation( this.annotation );
                 }
             }
         }
+        private boolean isRecursive() {
+            if(this.element == null || this.annotation == null)
```
Tomas-Kraus: _this.element == null_ is already being checked before this method is called
null value of _this.annotation_ would cause NPE in this.annotation.isMeta() so this is expected always != null

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java
```diff
@@ -372,14 +376,28 @@ public AnnotationVisitor visitArray(String name) {
...
+                return false;
+            if(this.annotation.getName().equals( this.element.getName())) {
+                //simple recursion (annotation annotated with itself)
+                return true;
+            }
+            MetadataClass aClass = this.element.m_factory.m_metadataClasses.get( this.annotation.m_name );
+            if ( aClass != null && !aClass.m_metaAnnotations.isEmpty() ) {
```
Tomas-Kraus: I'm not sure if isEmpty() check makes it much faster. It can save hash(key) computing for empty map, but on the other hand it's being called always.

PR #114, opened 2018-06-05
Update jpa.test script to allow easier testing against modules

lukasj:
This is first part of changes affecting mainly SE srg/lrg cases. Changes and updates to server side tests will come later but server-side testing has been added to travis now.


jpa/eclipselink.jpa.test/antbuild.properties
```diff
@@ -45,53 +56,41 @@ transaction.jar=javax.transaction.api_1.3.0.jar
 restfulws.jar=javax.ws.rs_2.0.1.v201504171603.jar
 jaxb-api.jar=javax.xml.bind_2.2.12.v201410011542.jar
 commonj.sdo.jar=commonj.sdo_2.1.1.v201112051852.jar
+persistence.jar=javax.persistence_2.0.6.v201308211412.jar
```
Tomas-Kraus: Looks like some kind of default that duplicates 2.0. But I don't see any usage in the diff.

lukasj: usages are commented out. I plan to either remove this in a later change (once I go through commented out stuff in order to remove/fix it) or use it as a fallback for default jpa api used by the build

PR #115, opened 2018-06-05
[2.7] Update core.test script to allow easier testing against modules

lukasj:
backport of #99 to 2.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 env:
   global:
-    - ANT_HOME=/usr/share/ant
+    - ANT_HOME=$HOME/apache-ant-1.10.2
     - M2_HOME=/usr/local/maven-3.5.2
   matrix:
     - TEST_TARGET=test-core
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ addons:
 
 env:
   global:
-    - ANT_HOME=/usr/share/ant
+    - ANT_HOME=$HOME/apache-ant-1.10.2
```
Tomas-Kraus: This is inconsistent with M2_HOME: 1st points to $HOME/..., 2nd to /usr/local/...
I would expect those utilities with the same prefix. But I did not see how it looks in Travis CI image so if it works fine ....it should be ok.

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/TestRunModel.java
```diff
@@ -13,13 +13,17 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Vector;
 
 import org.eclipse.persistence.internal.helper.Helper;
 import org.eclipse.persistence.sessions.DatabaseLogin;
-import org.eclipse.persistence.testing.framework.*;
+import org.eclipse.persistence.testing.framework.TestModel;
+
+import junit.framework.Test;
```
Tomas-Kraus: Imports only change? Why?

.travis.yml
lukasj: build uses default version of Maven (the one from `/usr/local/...`) and custom version of ant (explicitly downloaded and unzipped to`$HOME/...`); ant in `/usr/local/...` is version 1.9.6, or sth like that, and newer version is not available in trusty repo, so it can't be updated through `apt` (or at least not using default repos)

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/TestRunModel.java
lukasj: `git cherry-pick` found out that line having `AdvancedProcessingTestModel` has been already removed from the file in 2.7 by some earlier commit

Tomas-Kraus: Thanks for the response. Changes are already approved so go on with merging it.

PR #132, opened 2018-06-13
Fix classpaths for jpa tests

lukasj:
jpa tests in 3.0 nightly are failing with:
```
java.lang.NoClassDefFoundError: javax/json/JsonException
at org.eclipse.persistence.internal.oxm.record.SAXUnmarshaller.getNewXMLReader(SAXUnmarshaller.java:209)
...
```

this is to fix that


lukasj:
@rfelcman Can you check this, please?

rfelcman:
@lukasj I checked it too, but Tomas had trouble to assign me as a reviewer. Looks good.

PR #133, opened 2018-06-14
Bug 535250 - Recursive annotations triggers a StackOverflowError

Tomas-Kraus:
There are 2 commits in this PR:
 * Eclipse projects classpath fix to resolve problem with new JSON-P jars
 * StackOverflowError issue in meta-annotations cycle processing

This StackOverflowError fix changes last meta-annotation (before cycle is closed) to regular annotation. It's 1st option of possible fixes.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
@@ -219,7 +219,7 @@ public class MetadataAnnotatedElement extends MetadataAccessibleObject {
         if (metadataAnnotation == null) {
             for (MetadataAnnotation a: m_metaAnnotations.values()) {
                 if (names != null) {
-                    if (names.contains(a.getName())) {
+                    if (names.contains(a.getName()) && annotation.equals(a.getName())) {
                         return a;
                     }
                 } else {


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=535250

Tomas-Kraus:
Please do not merge it yet. I would like to hear whether you prefer fix in visitor visitEnd() method or in getAnnotation() method. Both fixes did break the cycle but fix in visitEnd() may cause getAnnotation() method to return unreliable results, because metadata do not contain full information (one meta-annotation was converted to regular one).

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
```diff
@@ -202,17 +204,29 @@ public MetadataAnnotation getAnnotation(Class annotation) {
...
         if (m_annotations == null && m_metaAnnotations == null) {
             return null;
         }
-
         MetadataAnnotation metadataAnnotation = m_annotations.get(annotation);
         if (metadataAnnotation == null) {
             for (MetadataAnnotation a: m_metaAnnotations.values()) {
-                MetadataAnnotation ma = m_factory.getMetadataClass(a.getName()).getAnnotation(annotation);
+                if (names != null) {
+                    if (names.contains(a.getName())) {
```
cristhiank: I think it's missing `annotation.equals(a.getName())` here because the `annotation` parameter is not always equals to `a.getName()`.
You can see an example of this in the following line https://github.com/eclipse-ee4j/eclipselink/blob/f7918de2097405feb7fbe5a5a9232cee00c11002/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/MetadataAccessor.java#L638

Tomas-Kraus: Yes, it makes sense. Will verify tomorrow.

foundation/org.eclipse.persistence.core/.classpath
```diff
@@ -5,7 +5,8 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="var" path="ECLIPSELINK_HOME/plugins/cdi.api_1.0.0.v201105160744.jar"/>
 	<classpathentry kind="var" path="ECLIPSELINK_HOME/plugins/javax.jms_1.1.0.v200906010428.jar"/>
-	<classpathentry kind="var" path="ECLIPSELINK_HOME/plugins/org.glassfish.javax.json_1.0.4.v201311181159.jar" sourcepath="/ECLIPSELINK_HOME/plugins/org.glassfish.javax.json.source_1.0.4.v201311181159.jar"/>
+	<classpathentry kind="var" path="ECLIPSELINK_HOME/plugins/javax.json-api_1.1.2.jar"/>
+	<classpathentry kind="var" path="ECLIPSELINK_HOME/plugins/javax.json_1.1.2.jar"/>
```
lukasj: core does not (and should not) depend on any particular implementation. BTW: this kind of things is better done in an extra commit

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
```diff
@@ -202,17 +204,29 @@ public MetadataAnnotation getAnnotation(Class annotation) {
     /**
      * INTERNAL:
      * Return the annotated element for this accessor. Note: This method does
-     * not check against a metadata complete.
+     * not check against a meta-data complete.
+     * @param annotation annotation name
      */
-    public MetadataAnnotation getAnnotation(String annotation) {
+    public MetadataAnnotation getAnnotation(final String annotation) {
+        return getAnnotation(annotation, (Set<String>)null);
```
lukasj: why not pass new HashSet here and replace null check with isEmpty() later?

Tomas-Kraus: I don't see significant difference in current code or your proposal. Some check still needs to be there (and I think null check may be a bit faster than calling a method of Hashset). I don't care much about this. If you think that your option will look better, I'll put it there.

Tomas-Kraus:
Replaced by PR 136 with final version of the fix.

PR #137, opened 2018-06-19
Bug 535250 - Recursive annotations triggers a StackOverflowError

Tomas-Kraus:
2.7 backport. See PR 136 for details.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
@@ -66,8 +66,10 @@ import static org.eclipse.persistence.internal.jpa.metadata.MetadataConstants.JP
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.persistence.annotations.Array;
 import org.eclipse.persistence.annotations.BasicCollection;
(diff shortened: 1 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=535250

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=535250

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAnnotatedElement.java
```diff
@@ -202,17 +204,32 @@ public MetadataAnnotation getAnnotation(Class annotation) {
...
+    protected final MetadataAnnotation getAnnotation(final String annotation, final Set<String> names) {
         if (m_annotations == null && m_metaAnnotations == null) {
             return null;
         }
-
         MetadataAnnotation metadataAnnotation = m_annotations.get(annotation);
         if (metadataAnnotation == null) {
             for (MetadataAnnotation a: m_metaAnnotations.values()) {
-                MetadataAnnotation ma = m_factory.getMetadataClass(a.getName()).getAnnotation(annotation);
+                if (names.contains(a.getName()) && annotation.equals(a.getName())) {
```
cristhiank: @Tomas-Kraus I think it still won't work. Could you add a test case where you invoke getAnnotation with an annotation name that is not present in the model for example "javax.persistence.Access", the system will loop recursively again because it won't match this "if" test. This "if" is only matched if the annotation is present. I would rewrite it with something like: 
`
if(annotation.equals(a.getName())return a; //existent recursive
if(names.contains(a.getName())continue; //non-existent recursive checked in past invocations
`

Tomas-Kraus: This is 2.7 backport, let's resolve it in master first.

PR #140, opened 2018-06-19
Bug 536067: Update license to EPL 2.0

lukasj:
This PR updates the license in sources from EPL/EDL 1.0 to EPL 2.0; license header has been added where possible if it was missing

Tests, mostly moxy and sdo, were updated to take license header in golden files into account.

Eclipse plugin/features related sources (readme/feature.properties/about) were updated as well.

@waynebeaton should we run this change through ee4j pmc or are we good to go without that Thanks!


lukasj:
@bshannon my bad, I overlooked EDL part :-( should be back. Thanks for catching this!

lukasj:
@bshannon :

re about.html - refers to the license of JPA 2.2 API

re  license.html - will fix

re features/common/license.html - it should be SUA per https://www.eclipse.org/projects/handbook/#legaldoc-plugins-features

re plugins/* - in general majority of these are test compile and test runtime dependencies, bigger part of the build is ant based; those with oracle in the name are binaries compiled against non-public dependencies (ie ojdbc driver etc). There are only 2 sub folders relevant to license update - o.e.p.antlr, o.e.p.asm - both contain repackaged dependencies and only one file there has been created by us

generally there seem two be 2 patterns where copyright tool reports wrong copyright (at least  version 1.50-SNAPSHOT):
*  missing "All rights reserved"
*  if there is a text between "...and/or its affiliates" and "All rights reserved." parts on the first line
will adress both

lukasj:
I removed obsolete sources from 'plugins' folder; and addressed the rest

PR #143, opened 2018-06-21
Table existence check per database platform.

Tomas-Kraus:
Optimized for MySQL.
Unfortunately platforms need control not just on the statement itself, but also on the way how it's executed and evaluated.


Tomas-Kraus:
Initial review, please do not merge yet.

PR #151, opened 2018-06-25
[master][no bug]: fix InvalidPathException: Illegal char exception during ant…

dazey3:
Running "ant -f antbuild.xml clean" fails with the following exception:

BUILD FAILED
...\eclipselink.runtime\antbuild.xml:477: java.nio.file.InvalidPathException: Illegal char <*> at index 43: ...\eclipselink.runtime\*.zip
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:194)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:165)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:89)

It doesn't say that you can use wildcards in the file attribute of the [delete task](https://ant.apache.org/manual/Tasks/delete.html)

### DIFF
--- a/antbuild.xml
+++ b/antbuild.xml
@@ -474,7 +474,9 @@
         <!-- ant antfile="antbuild.xml" dir="${eclipselink.util.rename}"            target="clean"/ -->
         <ant antfile="antbuild.xml" dir="${eclipselink.util.sigcompare}"        target="clean"/>
         <ant antfile="build.xml"    dir="${eclipselink.util.workbench}"         target="clean"/>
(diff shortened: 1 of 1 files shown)


dazey3:
The build failed for the reason: 

    The command "curl -sSL "http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x8C718D3B5072E1F5" | sudo -E apt-key add -" failed and exited with 2 during .

Since all the other tests + the 2.7 tests passed and this change is very minor, I'm going to merge anyway.

PR #162, opened 2018-07-02
[nobug] Change of source info for org.eclipse.persistence.Version

rfelcman:
[nobug] Change of source info for org.eclipse.persistence.Version

Build info is now generated and stored into version.properties file during a build.
Class org.eclipse.persistence.Version reads build info from version.properties file.


dazey3:
Could we get rid of the Version.java class (or reduce what it's used for) and just ship the version.properties file in the bundle's META-INF dir? OpenJPA does something very similar to include the version information in their shipped bundle.
Alternatively, if we want to keep Version.java so that users can run the main() and output the version details, we can just have the main() read the version.properties file from META-INF and output it. That'd drastically reduce what Version.java needs to store and do, yes?

lukasj:
Version being set as Main-Class in eclipselink.jar's manifest is just one use-case. The other is that build info from this class is being used in log messages (and in the end for debugging on customer's side...)

rfelcman:
This fix replace part of previous build system where Version.java class was used as a template and Ant during a build replaced there some substitution strings/variables before compilation.
In this case generated file is version.properties and source code of Version.java class will be untouched.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/Version.java
```diff
@@ -91,41 +89,67 @@ public static String getBuildNumber() {
...
+            System.out.println(getVersionString());
+            System.out.print("   Build Qualifier: ");
+            System.out.println(getQualifier());
+            System.out.print("   Build Date:      ");
+            System.out.println(getBuildDate());
+            System.out.print("   Build Time:      ");
+            System.out.println(getBuildTime());
+            System.out.print("   Build Revision:  ");
+            System.out.println(getBuildRevision());
+        }catch (Exception e) {
```
lukasj: be either specific or use Throwable

dazey3:
Forgive me if the maven update portion addresses this (not quite sure what copy-resources is doing), but will we be shipping the `version.properties` file as part of `org.eclipse.persistence.core`? I figure we have to or loadProperties() wont work.

rfelcman:
Yes `version.properties` file will be distributed as a part of `org.ecliplse.persistence.core` module. Maven during build of `org.ecliplse.persistence.core` module will prepare timestamp properties (buildnumber-maven-plugin) and maven-resources-plugin will copy this one to the output (.../target/classes..). During copy phase it will replace all substitution variables "${....}" in origin `version.properties` file by current values. The feature is enabled by `<filtering>true</filtering>` in the `pom.xml` in maven-resources-plugin configuration part.

dazey3:
@rfelcman Ok, so then all this change is really doing is changing the current process of: 
1) copy the values into the java class

to the process:
1) copy the values into a properties file
2) have the java class just read the properties file

I am assuming we have to do it this way cause of maven or the build process?

rfelcman:
@dazey3 Yes this is change of process to:
    1. copy the values into a properties file
    2. have the java class just read the properties file
We want replace `Version.java` class generation. This class/java source was generated before this patch. I think it's more clear to generate some properties file (like `version.properties`), then java sources. Maven is there (core module). I just added there some plugins to generate `version.properties` file.
Side note: `.gitignore` new `version.properties` file in sources is used as a template so we have to keep it in the git repo.

foundation/org.eclipse.persistence.core/version.properties
```diff
@@ -0,0 +1,6 @@
+version=${release.version}
```
lukasj: license is missing

foundation/org.eclipse.persistence.core/pom.xml
```diff
@@ -37,53 +37,66 @@
...
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.1.0</version>
+                <executions>
+                    <execution>
+                        <id>copy-resources</id>
+                        <phase>process-resources</phase>
                         <goals>
-                            <goal>run</goal>
+                            <goal>copy-resources</goal>
```
lukasj: Is this whole plugin config required? I'd assume version.properties file will be put under $resources folder and resources configuration would be updated to enable filtering on this particular resource folder (see https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html).

PR #182, opened 2018-07-13
Bug 425150 - Fixed bug in query result evaluation.

Tomas-Kraus:
2nd version of the fix - without a new interface and with copy-pasted query evaluation code in child methods.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -3531,7 +3531,7 @@ public class DatabasePlatform extends DatasourcePlatform {
      * @param table database table meta-data
      * @return query to check whether given table exists
      */
-    protected DataReadQuery getQuery(final TableDefinition table) {
+    protected DataReadQuery getTableExistsQuery(final TableDefinition table) {
          String column = null;
          for (FieldDefinition field : table.getFields()) {
              if (column == null) {
@@ -3557,7 +3557,7 @@ public class DatabasePlatform extends DatasourcePlatform {
      public boolean checkTableExists(final DatabaseSessionImpl session, final TableDefinition table) {
          try {
              session.setLoggingOff(true);
-             session.executeQuery(getQuery(table));
+             session.executeQuery(getTableExistsQuery(table));
              return true;
          } catch (Exception notFound) {
              return false;
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -269,6 +269,11 @@ public class DatabasePlatform extends DatasourcePlatform {
     protected Boolean useJDBCStoredProcedureSyntax;
     protected String driverName;
 
+    /** Table existence check trigger.
+     *  Returned result set is checked on not being empty when {@code true}
+     *  or {@code true} value is always returned when no exception is thrown when {@code false}. */
+    protected boolean shouldCheckResultTableExistsQuery;
+
     /**
      * Creates an instance of default database platform.
      */
@@ -295,6 +300,7 @@ public class DatabasePlatform extends DatasourcePlatform {
         this.endDelimiter = "\"";
         this.useJDBCStoredProcedureSyntax = null;
         this.storedProcedureTerminationToken = ";";
+        this.shouldCheckResultTableExistsQuery = false;
     }
 
     /**
@@ -3524,6 +3530,7 @@ public class DatabasePlatform extends DatasourcePlatform {
         throw new UnsupportedOperationException("Connection user name is not supported.");
     }
 
+    // Value of shouldCheckResultTableExistsQuery must be false.
     /**
      * INTERNAL:
      * Returns query to check whether given table exists.
@@ -3548,17 +3555,22 @@ public class DatabasePlatform extends DatasourcePlatform {
     }
 
     /**
-     * INTERNAL: Executes and evaluates query to check whether given table
-     * exists.
+     * INTERNAL:
+     * Executes and evaluates query to check whether given table exists.
+     * Query result evaluation depends on {@link #shouldCheckResultTableExistsQuery} value:
+     * Returned value depends on returned vector being non empty when {@code true} and
+     * returned value is always {@code true} when {@code false}.
      * @param session current database session
      * @param table database table meta-data
+     * @param suppressLogging whether to suppress logging during query execution
      * @return value of {@code true} if given table exists or {@code false} otherwise
      */
-    public boolean checkTableExists(final DatabaseSessionImpl session, final TableDefinition table) {
+    public boolean checkTableExists(final DatabaseSessionImpl session,
+            final TableDefinition table, final boolean suppressLogging) {
         try {
-            session.setLoggingOff(true);
-            session.executeQuery(getTableExistsQuery(table));
-            return true;
+            session.setLoggingOff(suppressLogging);
+            final Vector result = (Vector)session.executeQuery(getTableExistsQuery(table));
+            return shouldCheckResultTableExistsQuery ? !result.isEmpty() : true;
         } catch (Exception notFound) {
             return false;
         }
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -3539,16 +3539,7 @@ public class DatabasePlatform extends DatasourcePlatform {
      * @return query to check whether given table exists
      */
     protected DataReadQuery getTableExistsQuery(final TableDefinition table) {
-        String column = null;
-        for (FieldDefinition field : table.getFields()) {
-            if (column == null) {
-                column = field.getName();
-            } else if (field.isPrimaryKey()) {
-                column = field.getName();
-                break;
-            }
-        }
-        final String sql = "SELECT " + column + " FROM " + table.getFullName();
+        final String sql = "SELECT 1 FROM " + table.getFullName();
         final DataReadQuery query = new DataReadQuery(sql);
         query.setMaxRows(1);
         return query;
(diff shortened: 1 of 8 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=425150

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=425150

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -3502,46 +3511,69 @@ public void writeAddColumnClause(Writer writer, AbstractSession session, TableDe
...
+     * Query result evaluation depends on {@link #shouldCheckResultTableExistsQuery} value:
+     * Returned value depends on returned vector being non empty when {@code true} and
+     * returned value is always {@code true} when {@code false}.
+     * @param session current database session
+     * @param table database table meta-data
+     * @param suppressLogging whether to suppress logging during query execution
+     * @return value of {@code true} if given table exists or {@code false} otherwise
+     */
+    public boolean checkTableExists(final DatabaseSessionImpl session,
+            final TableDefinition table, final boolean suppressLogging) {
```
dazey3: I see that we are now adding a 'suppressLogging' boolean so that we can pass the value down the inheritance. Is there a usecase in which we would want suppressLogging to be false?
Also, I'm not a fan of breaking it up. I think either we keep the logging switch logic at the `SchemaManager` level, or we move it all (try-finally too) into `DatabasePlatform`

svendiedrichsen:
@Tomas-Kraus Having specific checks for MySql and Oracle is fine but what about the other platforms Eclipselink supports? 
i.e. Postgres: https://stackoverflow.com/questions/20582500/how-to-check-if-a-table-exists-in-a-given-schema
i.e. DB2: https://stackoverflow.com/questions/12888463/how-to-find-db2-tables-that-do-not-exist
and others ...

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -269,6 +269,14 @@
     protected Boolean useJDBCStoredProcedureSyntax;
     protected String driverName;
 
+    /** Table existence check trigger.
+     *  Returned result set is checked on not being empty when {@code true}
+     *  or {@code true} value is always returned when no exception is thrown when {@code false}. */
+    protected boolean shouldCheckResultTableExistsQuery;
```
dazey3: I see what you are trying to do here with the boolean. We want to be able to address both prospects of "try/fail" & "check result length". But I think we should leave that detail (the "how do we actually determine the table exists" detail) up to the individual platform classes. I think it really is as straight-forward as having just one method `public boolean checkTableExists(final DatabaseSessionImpl session, final TableDefinition table)` that the individual platform classes can override. I understand the consequence of code duplication, but I think the amount of copy/paste is negligible:

DatabasePlatform:
``` java
public boolean checkTableExists(final DatabaseSessionImpl session, final TableDefinition table) {
    final DataReadQuery query = new DataReadQuery("SELECT * FROM " + table.getFullName());
    try {
        session.executeQuery(query);
    } catch (Exception notFound) {
        return false;
    }
    return true;
}
```
OraclePlatform:

``` java
public boolean checkTableExists(final DatabaseSessionImpl session, final TableDefinition table) {
    final DataReadQuery query = new DataReadQuery("SELECT table_name FROM user_tables WHERE table_name='" + table.getFullName() + "'");
    query.setMaxRows(1);
    final Vector<?> result = new Vector<>();
    try {
        result = (Vector)session.executeQuery(query);
    } catch (Exception notFound) {
        return false;
    }
    return !result.isEmpty();
}
```

Reusing code is certainly admirable! But I'd personally prefer a simpler design that boasts readability/maintainability and is less complex :/

dazey3:
@svendiedrichsen The platforms that do not override will be caught by the platform agnostic implementation in DatabasePlatform. I believe the thought is that supporters of the other platforms can look at providing implementations to support those platforms if they choose.
It isn't done maliciously; the issue is having knowledge of these other platforms and access for testing. For instance DB2, we have DB2, Derby, DB2 on i and z. Curiously, I and Z both have different lookup tables and the documentation is very scarce. I will be updating these platforms when this initial implementation is committed

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: @dazey3: I see your point. The problem is, that each of us (me, Lukas and you) have bit different opinion about how to implement this. :) I would use that interface from previous PR which is from Java point of view optimal solution. Lukas prefers this version with if and you what you proposed.
Anyway, current code is result of compromise and resulting code is still quite simple. I would like to keep it. If overriding class needs its own way, it can still override both methods.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -3502,46 +3511,69 @@ public void writeAddColumnClause(Writer writer, AbstractSession session, TableDe
...
+    protected DataReadQuery getTableExistsQuery(final TableDefinition table) {
+        String column = null;
+        for (FieldDefinition field : table.getFields()) {
+            if (column == null) {
+                column = field.getName();
+            } else if (field.isPrimaryKey()) {
+                column = field.getName();
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
Tomas-Kraus: That's a question. It will raise query result data payload and slow it down a bit. I personally would replace column with something that will not go trough table data at all. For example "SELECT 1 FROM <table>".  It should also work on any DB.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: suppressLogging false may find some usage in junit tests where you want to see everything.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
dazey3: Well, I still disagree that this boolean implementation is clean, but if I am outvoted 2:1, that's fine by me. I don't argue this code most likely works, just that it seems a bit complex for what I saw as an otherwise simple task. We have lots of boolean flags in the code all over the place. I'd personally rather limit creating more. I'll approve assuming Lukas agrees.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
dazey3: In your unit test, if you'd be testing this impl, then you'd call the Platform classes implementation of `checkTableExists` then, right? Since `SchemaManager's` implementation is literally: 
1) turn off logging, 
2) call platform impl, 
3) turn back on logging.

I personally am not a fan of adding code to the product that is only meant to address testing usage, it clutters things up, but that's just me.
I hope you don't think I'm being a hard-a** about this stuff :/ I know most of this is personal preference.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: @dazey3, @lukasj: Well, I personally don't much care whether this will be there or not. It's another feature Lukas asked me to add. So you two make some agreement and let me know the result. :)

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: @dazey3, @lukasj: Well, I personally don't much care whether this will be there or not. It's another feature Lukas asked me to add. So you two make some agreement and let me know the result. :)

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: Well, each of the version has some good and bad things:
 * Interface (me):
   - -More complex code around the interface, worse human readibility
   - -Platform code contains class definition
   - +No copy-pastes
   - +Proper java pattern, allows lambdas to implement platform specific things
 * Boolean flag (Lukas):
   - -Additional boolean flag and condition around return statement
   - +Still good human readibility
   - +No copy-pastes
 * Pure method overrides (Will):
   - -Each overriding method copy-pastes query result evaluation code change
   - +Probably best human readibility

I already gave up my solution so you two should make an agreement which way to go. :)

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -3502,46 +3511,60 @@ public void writeAddColumnClause(Writer writer, AbstractSession session, TableDe
...
+    // Value of shouldCheckResultTableExistsQuery must be false.
+    /**
+     * INTERNAL:
+     * Returns query to check whether given table exists.
+     * Query execution throws an exception when no such table exists.
+     * @param table database table meta-data
+     * @return query to check whether given table exists
+     */
+    protected DataReadQuery getTableExistsQuery(final TableDefinition table) {
+        final String sql = "SELECT 1 FROM " + table.getFullName();
```
lukasj: If I get it right, then this query:
* returns exactly 1 row if table exists
* throws error if it does not

Correct? If so, why this is not reflected in javadoc? Does it make sense to have extra `shouldCheckResultTableExistsQuery` option and condition in `checkTableExists` to enforce `!result.isEmpty()` check?

Tomas-Kraus: OK, 2:1 for simple stupid version. Pushing in. :)

PR #184, opened 2018-07-18
[nobug][master] Enable line numbers for the ASM bundle

dazey3:
The **org.eclipse.persistence.asm** bundle does not have line numbers enabled. Other than a small size difference, it helps with debugging to link the source with the compiled bundle we provide.


dazey3:
@lukasj 
I think if 534001 is going to be fixed, that should be covered by it's own PR that changes all the manifests at once. Also, would we get rid of `Bundle-RequiredExecutionEnvironment`? I feel like we need it for backwards compatibility. Though, perhaps as a 3.0 change that shouldn't be a problem.

Also, I regenerated the new binaries (now with line numbers) and updated the PR, let me know if I messed it up xD

dazey3:
@lukasj Wait, what IDE files? I don't see anything in `eclipselink.runtime\plugins\org.eclipse.persistence.asm\.classpath` to update...

dazey3:
@lukasj Good catch, thanks. I added that change.

dazey3:
#192 looks like it kind of makes my change moot. Do we just want to have @Cousjava include my change to `plugins/org.eclipse.persistence.asm/antbuild.xml` in #192?

lukasj:
if you can wait or found your way around this for the time being (CQ usually takes about a week), then I'd say let's close this and wait for #192

PR #185, opened 2018-07-19
Bug 535998 WLS 12.2.1.2 DOES NOT PROPERLY IMPLEMENT JSR 303 ISSUE

sureshbn:

### DIFF
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
@@ -51,6 +51,7 @@ public class BeanValidationJunitTest extends JUnitTestCase {
             suite.addTest(new BeanValidationJunitTest("testRemoveWithInvalidData"));
             suite.addTest(new BeanValidationJunitTest("testTraversableResolverPreventsLoadingOfLazyRelationships"));
             suite.addTest(new BeanValidationJunitTest("testTraversableResolverPreventsTraversingRelationshipMultipleTimes"));
+            suite.addTest(new BeanValidationJunitTest("testPesistWithInvalidAndModifiedBeanData"));
         }
         return suite;
     }
@@ -342,6 +343,32 @@ public class BeanValidationJunitTest extends JUnitTestCase {
         assertTrue( "Lazy field should not be instantiated because of validation", !isInstantiated(employee, "projects", project) );
         assertTrue( "Lazy field should not be instantiated because of validation", !isInstantiated(employee, "managedProject", project) );
     }
+    
+    public void testPesistWithInvalidAndModifiedBeanData() {
+        
+        EntityManager em = createEntityManager();
+        boolean gotConstraintViolations = false;
+        String invalidName = getFilledStringOfLength(Employee.NAME_MAX_SIZSE + 1);
+        String validName = getFilledStringOfLength(Employee.NAME_MAX_SIZSE - 1);
+        
+        // Persist an object with valid value
+        beginTransaction(em);
+        Employee e1 = new Employee(100, validName, validName, 1337);
+        em.persist(e1);
+
+        // (b) now we demonstrate how eclipse link fails to invoke the JSR validations on entity
+        // when we modify the entity during the transaction with invalid value
+        try {
+            e1.setName(invalidName);
+            e1.setSurname(invalidName);
+            em.persist(e1);
+        }catch (ConstraintViolationException e) {
+            assertTrue("Transaction not marked for roll back when ConstraintViolation is thrown", getRollbackOnly(em)) ;
+            gotConstraintViolations =true;
+        }
+        assertTrue("Did not get Constraint Violation while persisting invalid data ", gotConstraintViolations);
+   }
+    
 
     /**
      * Strategy:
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=535998

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=535998

Tomas-Kraus (commented):
Change looks good.
Just see that you are not using formating rules from project-admin/EclipseLink-Eclipse-Format.xml
It would be good to use them.

sureshbn:
Applied the formatter [EclipseLink-Eclipse-Format.xml] as specified . In this particular case we are not seeing triggering of events anywhere while updating the entity  - Thanks

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -342,6 +343,32 @@ public void testTraversableResolverPreventsLoadingOfLazyRelationships() {
...
+        beginTransaction(em);
+        Employee e1 = new Employee(100, validName, validName, 1337);
+        em.persist(e1);
+
+        // (b) now we demonstrate how eclipse link fails to invoke the JSR validations on entity
+        // when we modify the entity during the transaction with invalid value
+        try {
+            e1.setName(invalidName);
+            e1.setSurname(invalidName);
+            em.persist(e1);
```
dazey3: No ```em.merge(e1)```? The entity is already managed at this point. I ran a quick local test and I don't see EclipseLink calling @PrePersist twice when there are multiple persists for the same managed entity. I can verify that with EclipseLink's implementation as well. Take a look at `org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.registerNotRegisteredNewObjectForPersist`:

``` java
   /**
     * INTERNAL:
     * Called only by registerNewObjectForPersist method,
     * and only if newObject is not already registered.
     * Could be overridden in subclasses.
     */
```
If the object being persisted is already registered, we don't execute the PrePersistEvent.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -342,6 +343,32 @@ public void testTraversableResolverPreventsLoadingOfLazyRelationships() {
         assertTrue( "Lazy field should not be instantiated because of validation", !isInstantiated(employee, "projects", project) );
         assertTrue( "Lazy field should not be instantiated because of validation", !isInstantiated(employee, "managedProject", project) );
     }
+    
+    public void testPesistWithInvalidAndModifiedBeanData() {
```
dazey3: Is this a different test than `BeanValidationJunitTest.testUpdateWithInvalidData()`? They seem to be trying to test the same thing, but `testUpdateWithInvalidData` doesn't seem to be using callbacks and is instead expecting validation on transaction commit. Is this right?

https://github.com/eclipse-ee4j/eclipselink/blob/b87d11af0889564f8da540bec3cdf98c14cd9894/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java#L252-L258

According to 3.6 of the spec:

    These lifecycle validation events occur immediately after the point at which all the 
    PrePersist, PreUpdate, and PreRemove lifecycle callback method invocations respectively 
    have been complete

I'm wondering if this means `testUpdateWithInvalidData` is wrong.

lukasj: I saw a javadoc somewhere saying that current behaviour is expected and intended by the spec - prepersist is called on the first persist call and not during transaction commit, so subsequent updates to the entity, is is done here are not validated as some people would expect. I just don't know, where I saw that comment :-/

lukasj: here it is: https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java#L74

dazey3: @lukasj I see! I think there is a misconception then that validation occurs on transaction commit. After reading the spec, it seems pretty clear that validation occurs on the EntityManager operation invocations. However, I searched around online for some examples and it seems transaction commit is being expected by users. Given the comment you found, it certainly sounds like the current spec design has issues that we may want to address in the future.

For example, here is a [Hibernate example](https://www.thoughts-on-java.org/automatically-validate-entities-with-hibernate-validator/) I found. It seems the example they are using is expecting transaction commit. Perhaps Hibernate is not adhering to the spec? Idk.

sureshbn:
The conclusion here is : the spec need to evolve/clarify for this case and we need to wait up to that time ? , Please help to confirm

arjantijms:
More than 8 years old, let's close. If anyone ever wants to look at it, please open new PR for this properly rebased.

PR #194, opened 2018-07-25
Bug 536947 – XmlDiscriminatorNode with JSON in MoxyBindings File Is not applied to Root Elements

rfelcman:
Bug 536947 – XmlDiscriminatorNode with JSON in MoxyBindings File Is not applied to Root Elements

MOXy before this fix always applied to the JSON root element attribute name ‘type’ if XmlDiscriminatorNode was enabled (via `XmlDiscriminatorNode` annotation or XML binding configuration file). There was same strategy as in XML with
 `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="car"` . There was inconsistency with rest of JSON sub elements where custom name, specified in annotation or binding configuration file, is applied.
After this fix custom name of type attribute is applied to the JSON root element too instead of hardcoded ‘type'.
There were incorrect test/resource files ‘vehicle-write.json’ for test classes `org.eclipse.persistence.testing.jaxb.annotations.xmldiscriminator.XmlDiscriminatorTestCases`
and `org.eclipse.persistence.testing.jaxb.externalizedmetadata.xmldiscriminator.XmlDiscriminatorTestCases`.
 This is fix for this bug and test classes too.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=536947


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=536947

PR #197, opened 2018-08-01
Bug 537554 WRONG SCHEMA GENERATED - WSDL CAN NOT OPEN IN SOAPUI

sureshbn:

### DIFF
--- a/moxy/org.eclipse.persistence.moxy/src/org/eclipse/persistence/jaxb/compiler/SchemaGenerator.java
+++ b/moxy/org.eclipse.persistence.moxy/src/org/eclipse/persistence/jaxb/compiler/SchemaGenerator.java
@@ -649,7 +649,10 @@ public class SchemaGenerator {
     }
 
     private Schema getSchemaForNamespace(String namespace) {
-        return getSchemaForNamespace(namespace, null);
+        if (namespace != null && !namespace.equals(EMPTY_STRING)) {
+            return getSchemaForNamespace(namespace, null);
+        }
+        return null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dazey3:
@sureshbn Also, it would be a good idea to include a test with your change. Perhaps then we can see what usecase this defect is covering. Thanks!

sureshbn:
Corrected the condition  and test execution good now.
Seems we have a complex scenario : issue comes in integration with WLS-Webservice-Eclipselink . 
Have a test case which have dependency on additional 3-5 JAR's
[com.oracle.webservices.wls.wls-ws-metainf-services-impl.jar
com.sun.xml.bind.jaxb-core.jar
com.sun.xml.bind.jaxb-impl.jar
com.sun.xml.ws.jaxws-eclipselink-plugin.jar
com.sun.xml.ws.jaxws-rt.jar] 
whether we can include a JUNIT with these dependency JAR's ? , If so where the dependency JAR need to be placed ? any sample/pointer we have ?  whether we can avoid test-case for these cases ,if it is not break the existing?

lukasj:
@sureshbn If I get it right: jaxws-eclipselink-plugin is a wrapper around moxy to plug it into jaxws-rt. Why exactly this should be made part of the repo? Jaxb-core/impl are moxy dependencies to some extent and therefore should be already in the repo. Did you check the repo to see if they aren't there already? metainf-services-impl jar should be just some helper jar with meta-inf/services definition files.

Not adding a test is not an option in this case.

lukasj:
@rfelcman since this is in MOXy, can you help Suresh with this, please? Thanks!

rfelcman:
See comment at https://bugs.eclipse.org/bugs/show_bug.cgi?id=537554#c4

lukasj:
can be reopened once new test is added, if needed

Issue #212, opened 2018-08-08
Bug 537795: CASE THEN and ELSE scalar expression Constants should not be casted to CASE operand type


jgrassel:
Issue for https://bugs.eclipse.org/bugs/show_bug.cgi?id=537795

Given a query:

TypedQuery<EntityTbl01> query = em.createQuery(""
                + "SELECT t FROM EntityTbl01 t "
                    + "WHERE t.itemString1 = ( "
                        + "CASE "
                            + "WHEN t.itemInteger1 = 1000 THEN '047010' "
                            + "WHEN t.itemInteger1 = 100 THEN '023010' "
                            + "ELSE '033020' "
                        + "END )", EntityTbl01.class);

the query processing involved with converting the CASE scalar_expressions to statement parameters typecasts the constant value to the case_operation type.  For THEN/ELSE scalar_expressions of an incompatible type this can lead to a failure.

PR #213, opened 2018-08-08
[2.6_WAS] Bug 537795

jgrassel:
for #212

… be casted to CASE operand type.

Signed-off-by: Joe Grassel <jgrassel@us.ibm.com>


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/ExpressionOperator.java
```diff
@@ -2192,7 +2194,9 @@ public void printCollection(Vector items, ExpressionSQLPrinter printer) {
                 printer.printString(alias.getNameDelimited(printer.getPlatform()));
             } else if ((this.selector == Count) && (item.isExpressionBuilder())) {
                 printer.printString("*");
-            } else {
+            } else if ((this.selector == Case) && (item.isConstantExpression())) {
```
dazey3: I dont think this is the solution. Both the THEN and CLAUSE items are ConstantExpressions. We want to call ```item.printSQL(printer)``` for the THEN clauses so that they use the baseExpression.

jgrassel:
Cancelling, going with the fix in pull request 214.

PR #220, opened 2018-08-20
Bug 326728 - Incorrect persistence root calculation for WARs

nigredo-tori:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=326728

Changes the root calculation algorithm for WAR and JAR archives to produce the
result that correlates with the `descriptorLocation` parameter. The new algorithm
follows the link between `persistence.xml` and the persistence root rather
then the rigid list of possible cases in the JPA specification (8.2), which allows it
to handle some exotic cases. For example, given the input
`jar:file:/foo.jar!/bar/META-INF/persistence.xml`, the algorithm will produce root
`jar:file:/foo.jar!/bar`. This is useful for embedded servers (mainly Jetty), where
we wish to maintain separation between the server and application classes.

For `zip:...!/...` URLs the resulting root in such case will have the `jar:file:`
scheme. This allows us to leverage the existing JAR file tooling.

This is compatible with the 2.7.x branch, so it can be backported after merge.

### DIFF
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
@@ -72,6 +72,34 @@ public class PersistenceUnitProcessorTest extends JUnitTestCase {
                         "META-INF/persistence.xml"
                 ).toString()
         );
+
+        Assert.assertEquals(
+            // The protocol changes - we view ZIPs as JARs.
+            "jar:file:/foo/bar.war!/WEB-INF/classes",
+            PersistenceUnitProcessor.computePURootURL(
+                new URL("zip", "", -1, "/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml", dummyZipHandler), 
+                "META-INF/persistence.xml"
+            ).toString()
+        );
+    }
+
+    public void testComputePURootURLForJarFile() throws Exception {
+
+        Assert.assertEquals(
+            new URL("file:/foo/bar.jar"),
+            PersistenceUnitProcessor.computePURootURL(
+                new URL("jar:file:/foo/bar.jar!/META-INF/persistence.xml"),
+                "META-INF/persistence.xml"
+            )
+        );
+
+        Assert.assertEquals(
+            new URL("jar:file:/foo/bar.war!/WEB-INF/classes"),
+            PersistenceUnitProcessor.computePURootURL(
+                new URL("jar:file:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml"),
+                "META-INF/persistence.xml"
+            )
+        );
     }
 
     public void testGetArchiveFactory() {
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
@@ -183,19 +183,61 @@ public class PersistenceUnitProcessor {
             // 210280: any file url will be assumed to always reference a file (not a directory)
             result = new URL(pxmlURL, path.toString()); // NOI18N
         } else if("jar".equals(protocol)) { // NOI18N
+            // JPA spec limits the possible scenarios in Java EE environments (8.2).
+            // However it makes sense to be a little more flexible, supporting
+            // any root URLs that resolve descriptorLocation to pxmlURL.
+            // For example, jar:file:/foo.jar!/bar/META-INF/persistence.xml
+            // will have root in jar:file:/foo.jar!/bar.
+
             // e.g. jar:file:/tmp/a_ear/b.jar!/META-INF/persistence.xml
             JarURLConnection conn =
                     JarURLConnection.class.cast(pxmlURL.openConnection());
-            result = conn.getJarFileURL();
-        } else if("zip".equals(protocol)) { // NOI18N
-            // e.g. zip:/tmp/a_ear/b.jar!/META-INF/persistence.xml
-            // stolen from java.net.JarURLConnection.parseSpecs method
-            String spec = pxmlURL.getFile();
-            int separator = spec.lastIndexOf("!/");
-            if (separator == -1) {
-                separator = spec.length() - 1;
+            // Entry path, e.g. META-INF/persistence.xml. Empty string
+            // corresponds to the archive root.
+            String entry = conn.getEntryName();
+            if (entry == null) entry = "";
+            // Go up to the directory, corresponding to root of descriptorLocation
+            // (twice for META-INF/persistence.xml).
+            for(int i = 0; i <= descriptorDepth; i++) {
+                int entrySeparator = entry.lastIndexOf("/");
+                entry = entry.substring(0, Math.max(0, entrySeparator));
             }
-            result = new File(spec.substring(0, separator++)).toURL();
+
+            result = entry.isEmpty()
+                ? conn.getJarFileURL()
+                : new URL("jar:" + conn.getJarFileURL() + "!/" + entry);
+        } else if("zip".equals(protocol)) { // NOI18N
+            // More permissive than the JPA spec - see the JAR comment above.
+
+            // e.g. /tmp/a_ear/b.jar!/META-INF/persistence.xml
+            String input = pxmlURL.getFile();
+            int separator = input.lastIndexOf("!/");
+            // File part of the input, e.g. zip:/tmp/a_ear/b.jar
+            String file;
+            // Entry path, e.g. META-INF/persistence.xml. Empty string
+            // corresponds to the archive root.
+            String entry;
+            if (separator == -1) {
+                file = input;
+                entry = "";
+            } else {
+                file = input.substring(0, separator);
+                entry = input.substring(separator + 2);
+            }
+            // Go up to the directory, corresponding to root of descriptorLocation
+            // (twice for META-INF/persistence.xml).
+            for(int i = 0; i <= descriptorDepth; i++) {
+                int entrySeparator = entry.lastIndexOf("/");
+                entry = entry.substring(0, Math.max(0, entrySeparator));
+            }
+
+            if (entry.isEmpty()) {
+                result = new File(file).toURL();
+            } else {
+                // JARs and ZIPs are virtually the same thing, but we have a lot
+                // better tooling for JARs.
+                result = new URL("jar:file:" + file + "!/" + entry);
+            }
         } else if("wsjar".equals(protocol)) { // NOI18N
             // e.g. wsjar:file:/tmp/a_ear/b.jar!/META-INF/persistence.xml
             // but WS gives use jar:file:..., so we need to match it.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=326728

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -183,19 +183,61 @@ public static URL computePURootURL(URL pxmlURL, String descriptorLocation) throw
             // 210280: any file url will be assumed to always reference a file (not a directory)
             result = new URL(pxmlURL, path.toString()); // NOI18N
         } else if("jar".equals(protocol)) { // NOI18N
+            // JPA spec limits the possible scenarios in Java EE environments (8.2).
+            // However it makes sense to be a little more flexible, supporting
+            // any root URLs that resolve descriptorLocation to pxmlURL.
+            // For example, jar:file:/foo.jar!/bar/META-INF/persistence.xml
+            // will have root in jar:file:/foo.jar!/bar.
```
dazey3: As the reference implementation, I don't think I want to support any persistence.xml location not explicitly declared as part of section 8.2 of the JPA spec... I can understand wanting to be more flexible, but there is also the argument that ```jar:file:/foo.jar!/bar/META-INF/persistence.xml``` is just wrong according to the java jar specification. ```META-INF``` must be at the root of your jar. Users should understand the correct way of doing things rather than us trying to just accept anything.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
```diff
@@ -72,6 +72,34 @@ protected URLConnection openConnection(URL u) throws IOException {
...
+            )
+        );
+
+        Assert.assertEquals(
+            new URL("jar:file:/foo/bar.war!/WEB-INF/classes"),
+            PersistenceUnitProcessor.computePURootURL(
+                new URL("jar:file:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml"),
+                "META-INF/persistence.xml"
+            )
+        );
```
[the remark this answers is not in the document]
dazey3: Perhaps you should take a look at the implementation for ```wsjar```:
https://github.com/eclipse-ee4j/eclipselink/blob/ea5c77487819d44b821bee545651d85a8cd93535/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java#L199-L219

For WebSphere, we rewrap our ```jar``` protocol with wsjar. This is why we added a special case here for just this protocol, but it seems the implementation tried to correctly take into account this usecase:

``` java
                // If this doesn't reference a war file with a properly located persistence.xml,
                // then chop off everything after the "!/" marker and assume it is a normal jar.
                // Else, if the wsjar URL references a file with a ".war" extension, and its entry
                // starts with WEB-INF/classes/, then the calculated persistence unit root should
                // be wsjar:path/to/a.war!/WEB-INF/classes/ as per JPA 2.1 Spec section 8.2 "Persistence Unit Packaging".
                // Filter out invalid scenarios such as wsjar:file:/a/path/to/my.war!/foo/WEB-INF/classes/META-INF/persistence.xml
```
This may explain why I have not seen this as a problem :/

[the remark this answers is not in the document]
nigredo-tori: > I'm surprised we do not correctly calculate the root for this.

> This may explain why I have not seen this as a problem :/

We've been dealing with Tomcat the whole time, which explodes jars by default - I'm guessing other web containers are doing the same thing?.. I've only stumbled upon this issue once I attempted to use embedded Jetty for this (and package everything into a jar - running from a directory is OK).

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: Ah, I forgot EclipseLink is a reference implementation. Then, indeed, extra flexibility is evil. I'll clean the comments/tests up tomorrow. That said, the logic itself shouldn't change, IMO - we still can have a persistence root in a directory inside a JAR. Validating the location of this root would be nice, but that's not the point of this PR.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -182,41 +182,39 @@ public static URL computePURootURL(URL pxmlURL, String descriptorLocation) throw
...
-                separator = spec.length() - 1;
-            }
-            result = new File(spec.substring(0, separator++)).toURL();
-        } else if("wsjar".equals(protocol)) { // NOI18N
-            // e.g. wsjar:file:/tmp/a_ear/b.jar!/META-INF/persistence.xml
-            // but WS gives use jar:file:..., so we need to match it.
-            String spec = pxmlURL.getFile();
+        } else if("zip".equals(protocol) ||
+                  "jar".equals(protocol) ||
+                  "wsjar".equals(protocol)) {
```
dazey3: Id rather you not mess with ```wsjar``` please. I referenced that implementation because it seemed to already account for the usecases we want covered by ```jar```, such as "path/to/a.war!/WEB-INF/classes/ as per JPA 2.1 Spec section 8.2". But unless you added a new, valid test for ```wsjar``` and it is failing, I'd rather you just leave that implementation alone. Call me paranoid, but we haven't had any issues that I know of so Id rather keep it that way :/

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -182,41 +182,39 @@ public static URL computePURootURL(URL pxmlURL, String descriptorLocation) throw
...
+            // Example URLs and expected results:
+            // zip:/foo/bar.jar!/META-INF/persistence.xml -> file:/foo/bar.jar
+            // zip:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml -> jar:file:/foo/bar.jar!/WEB-INF/classes
+            // jar:file:/foo/bar.jar!/META-INF/persistence.xml -> file:/foo/bar.jar
+            // jar:file:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml -> jar:file:/foo/bar.jar!/WEB-INF/classes
+            // wsjar:file:/foo/bar.jar!/META-INF/persistence.xml -> file:/foo/bar.jar
+            // wsjar:file:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml -> wsjar:file:/foo/bar.jar!/WEB-INF/classes
```
dazey3: It would be nice if we had these as tests in PersistenceUnitProcessorTest :)

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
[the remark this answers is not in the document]
dazey3: > Should I just reuse this (restrictive) logic? The way I see it, zip, jar and wsjar should all work the same.
Not sure what you mean by "restrictive"... are you referring to the fact that we only want to support specification supported URLs?

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
dazey3: Also negative tests for invalid scenarios like ```wsjar:file:/a/path/to/my.war!/foo/WEB-INF/classes/META-INF/persistence.xml``` curious how that is handled.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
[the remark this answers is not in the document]
nigredo-tori: > Not sure what you mean by "restrictive"... are you referring to the fact that we only want to support specification supported URLs or the fact that we have each protocol with separate logic?

By "restrictive" I meant the fact that we go out of our way to make sure that any cases not conforming to the specification (like `jar:file:foo.jar!/bar/META-INF/persistence.xml`) resolve to the top of the archive. The thing that irks me most about this approach is that resolution works differently depending on the packaging:
```
jar:file:/foo.jar!/bar/META-INF/persistence.xml -> file:/foo.jar
file:/foo/bar/META-INF/persistence.xml          -> file:/foo/bar
```
This can lead to all sorts of subtle and confusing bugs, and this is what bit me in the first place. Finding some configuration that maps the same way in both cases works (somewhat) in my case, but I still find this behavior surprising. Look at it this way: since these URLs don't conform to the spec, we have a choice to either raise an error, or handle them in the most consistent and predictable way possible. As I see it, the existing approach of using the archive root fails to do this.

However since we're keeping this logic for `wsjar` anyway, I guess it does make a modicum of sense to do the same thing for other archives as well. At least for my case fixing `WEB-INF/classes` should be good enough.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: Sure thing. Promoted all of these to test cases.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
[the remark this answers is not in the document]
dazey3: >By "restrictive" I meant the fact that we go out of our way to make sure that any cases not conforming to the specification (like jar:file:foo.jar!/bar/META-INF/persistence.xml) resolve to the top of the archive.

Maybe we should throw some kind of warning if the user has their persistence.xml at a bad location. We can then proceed forward, but the warning will have let the user know we are operating off a bad location.

> we have a choice to either raise an error, or handle them in the most consistent and predictable way possible. As I see it, the existing approach of using the archive root fails to do this.

I think I'd rather raise an error. So, if I understand your situation correctly, the fact that the archive root is being returned is causing the issue later in EclipseLink? We set the URL to point at the archive root, like ```path/to/foo.jar```, and this can't be handled correctly somewhere in EclipseLink?

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
```diff
@@ -35,42 +35,140 @@
...
+
+        checkPURootCustom(
+            "wsjar", "file:/foo/bar.war!/WEB-INF/classes/META-INF/persistence.xml",
+            "jar:file:/foo/bar.war!/WEB-INF/classes/"
+        );
+
+        checkPURootCustom(
+            "wsjar", "file:/foo/bar.jar!/WEB-INF/classes/META-INF/persistence.xml",
+            "jar:file:/foo/bar.jar!/"
+        );
```
dazey3: I'm fairly certain this is not a usecase covered by Section 8.2:

    In Java EE environments, the root of a persistence unit must be one of the following:
     • an EJB-JAR file
     • the WEB-INF/classes directory of a WAR file[87]
     • a jar file in the WEB-INF/lib directory of a WAR file
     • a jar file in the EAR library directory
     • an application client jar file

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
```diff
@@ -35,42 +35,140 @@
...
+
+        checkPURootCustom(
+            "wsjar", "file:/foo/bar.jar!/foo/META-INF/persistence.xml",
+            "jar:file:/foo/bar.jar!/"
         );
```
dazey3: This is also not a valid persistence.xml location. I'm not sure what we should do for this situation (my gut instinct is that the URL should be validated before it gets here), but I still think we shouldnt be testing this like it's valid

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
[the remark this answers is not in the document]
nigredo-tori: > I think I'd rather raise an error.

Personally, I'd rather have consistent behavior in this cases, but for a reference implementation an error is also OK. I'll figure something out on my end.

> So, if I understand your situation correctly, the fact that the archive root is being returned is causing the issue later in EclipseLink? We set the URL to point at the archive root, like path/to/foo.jar, and this can't be handled correctly somewhere in EclipseLink?

As I recall, the error itself was pretty simple - the "missing" `persistence.xml` resulted in a `@NamedQuery` lookup issue. Sure, there was a warning somewhere (I think), but spotting it in a log of a large-ish application is not easy, especially if you don't know what to look out for. The confusing part was that everything worked OK in the development environment (when running from an exploded directory), but failed for the JAR.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
nigredo-tori: Same as the other test case - we make sure our behavior doesn't suddenly change for cases that don't conform to the JPA spec.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
```diff
@@ -35,42 +35,140 @@
...
-        
-        Assert.assertEquals(
-                "file:/C:/Program.Files/Middleware/Oracle_Home/wlserver/samples/domains/mydomain/servers/myserver/tmp/_WL_user/eclipselink-advanced-model/y986my/eclipselink-advanced-model_ejb.jar", 
-                PersistenceUnitProcessor.computePURootURL(
-                        new URL("zip", "", -1, "/C:/Program.Files/Middleware/Oracle_Home/wlserver/samples/domains/mydomain/servers/myserver/tmp/_WL_user/eclipselink-advanced-model/y986my/eclipselink-advanced-model_ejb.jar!/META-INF/persistence.xml", dummyZipHandler), 
-                        "META-INF/persistence.xml"
-                ).toString()
+    }
+
+    public static void testComputePURootURLForZipFile() throws Exception {
```
dazey3: I dont think this needs to be static

nigredo-tori: Coming from Scala, I have to say that it doesn't have to be non-static :smile: . Nevertheless, I've changed this to fit the project style.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
dazey3: This isnt a negative test. A negative test would be validating that passing a bad URL will result in a failure. This test is verifying that the URL ```bar.jar!/WEB-INF/classes/META-INF/persistence.xml``` gives us back the expected result of ```jar:file:/foo/bar.jar!/``` The URL is invalid in the first place. We should give back like null or throw an exception or something. Giving a result back makes it look like this is acceptable.
Granted we want to preserve

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
dazey3: That's the beauty of having a spec define your expected behavior: if it isnt supported, users shouldnt be relying on it. We don't have to make sure old, bad behavior is still supported. Instead we should throw some exception or return null or something more than providing what is otherwise a valid result.

Maybe @lukasj has a different opinion? Id be interested in thoughts on this.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
nigredo-tori: Got it, will do that tomorrow.

nigredo-tori:
I added validation for archive roots - `computePURootURL` now returns `null` if it fails to determine a valid root. I've also changed the implementation for `wsjar:` again: this time the behavior for valid cases stays the same, but for invalid cases the algorithm returns `null` (same as the other archives).

Since the contract for `computePURootURL` has changed, I had to update three places where it was used:
- `PersistenceUnitProcessor.findPersistenceArchives` - just ignore the invalid roots.
- `InMemoryArchive`, `PersistenceUnitImpl` - for `computePURootURL` throw an `IllegalStateException` if failed to compute the corresponding roots. However, I'm not sure this is the right way to go...

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -719,4 +727,23 @@ public static String buildPersistenceUnitName(URL url, String puName){
...
+     * to JPA Spec (8.2).
+     */
+    private static boolean isValidRootInArchive(String file, String rootEntry) {
+        String extension = file.substring(Math.max(0, file.length() - 4));
+        boolean valid;
+        if (extension.equalsIgnoreCase(".jar")) {
+            // For a JAR, the root can only be the archive itself.
+            return rootEntry.isEmpty();
+        } else if (extension.equalsIgnoreCase(".war")) {
+            // For a WAR, the root can only be WEB-INF/classes.
```
dazey3: :/ Not exactly true... 

Spec 8.2: 

    a jar file in the WEB-INF/lib directory of a WAR file

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -719,4 +727,23 @@ public static String buildPersistenceUnitName(URL url, String puName){
        return fullPuName;
    }
 
+    /**
+     * @param file archive file URL
+     * @param entry a directory entry in the archive (or an empty string)
+     * @return true if the file-entry pair can be a persistence root according
+     * to JPA Spec (8.2).
+     */
+    private static boolean isValidRootInArchive(String file, String rootEntry) {
```
dazey3: I feel like we should just put this logic in the ```computePURootURL``` method. There we already have the protocol so ```extension``` is available, this helper method is only called once anyway, and we can throw an Exception instead of bothering with boolean checks

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: Hm, the comment is indeed a tad misleading. Any idea how I should rephrase this? Maybe, just copy&paste examples from the spec?

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
```diff
@@ -182,41 +186,44 @@ public static URL computePURootURL(URL pxmlURL, String descriptorLocation) throw
...
-                if (spec.regionMatches(true, separator - 6, ".war", 0, 4) &&
-                        spec.regionMatches(true, separator, WEBINF_CLASSES_STR, 0, WEBINF_CLASSES_LEN)) {
-                    separator += WEBINF_CLASSES_LEN;
-                }
+
+            String rootEntry = entry.substring(0, entry.length() - descriptorLocation.length());
+
+            // Since EclipseLink is a reference implementation, let's validate
+            // the produced root!
+            if (!isValidRootInArchive(file, rootEntry)) {
```
dazey3: I was thinking about this a bit more and I think throwing an exception should be better. Take a look at

https://github.com/eclipse-ee4j/eclipselink/blob/ac5913f745ebbe74d1dcd3e38ffea1baba208141/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/config/persistenceunit/PersistenceUnitImpl.java#L53-L60

What if we changed that to this instead:

``` java
        URL puURL = PersistenceUnitImpl.class.getClassLoader().getResource(persistenceFactoryResource);
        try{
            puURL = PersistenceUnitProcessor.computePURootURL(puURL, persistenceFactoryResource);
        } catch (Exception ex){
            throw PersistenceUnitLoadingException.exceptionProcessingPersistenceUnit(puURL, ex);
        }
```

The other usage already takes care of exceptions so that's covered:
https://github.com/eclipse-ee4j/eclipselink/blob/ac5913f745ebbe74d1dcd3e38ffea1baba208141/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java#L332

Then, we just need an exception to throw from ```computePURootURL()```. I think adding a new exception message here is warranted. I couldnt find an existing one that makes any sense. Take a look at:
https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/exceptions/ValidationException.java 
https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/exceptions/i18n/ValidationExceptionResource.java
to see how we can add a new exception. I'm thinking something like ```{ "7357", "The following URL is not valid {0}"}``` maybe? Then we can throw that exception from ```computePURootURL()```? Thoughts? :)

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: > I feel like we should just put this logic in the computePURootURL

In my opinion, `computePURootURL` is getting unwieldy as it is. This validation doesn't interact with the computation algorithm itself, so for me it makes sense to name and separate this part of code. This way I can keep the computation simpler as well - just cut off the descriptor, clean up and we're done.

> There we already have the protocol so extension is available,

Protocol here doesn't determine the extension. We have three protocols (`jar:`, `wsjar:`, `zip:`), and two extensions of interest (`.jar`, `.war`), and all six combinations are possible (as far as I see).

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: > Thoughts? :)

Sure, I'll get on it tomorrow.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
dazey3: Ok, you convinced me :)

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
dazey3: Ya, I think that should be fine. I think it would be good to add the same comments to the matching tests with the spec quotes.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
nigredo-tori: Done.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
dazey3: Just checking back up on this; sorry, Ive been busy. Were you planning on adding the ```WEB-INF/lib``` case as valid for ```isValidRootInArchive()```?

nigredo-tori: This gets kind of weird. The spec says:
```
a jar file in the WEB-INF/lib directory of a WAR file
```
which means that `pxmlURL` in this case must be an entry inside of an entry. I know that this isn't supported in the `jar:` URLs. Could you give me an example of such URLs for other schemes?

In any case, I'm not sure it makes sense to forbid all such nesting archives except the ones explicitly permitted by the spec.

dazey3: I'm guessing this would be an exploded war? So I think the URL would look like this? : ```jar:file:/foo/bar.war/WEB-INF/lib/jpa.jar!``` I'm actually not sure. Perhaps this isn't a special case for ```war``` in this method.

nigredo-tori: > I'm guessing this might come in as an exploded war?

For exploded war we can't really validate this. If the file path ends in `WEB-INF/lib/foo.jar` - it might be a lib JAR inside an exploded WAR, if not - it's just a stand-alone JAR. Both are possible, and both should be allowed, so it doesn't matter to check in this case.

The problem is what happens if the WAR is NOT exploded. At least Tomcat seems to [allow that](https://tomcat.apache.org/tomcat-8.5-doc/config/context.html) (search for `unpackWar`), though I don't know the details.

> My default is just that we should cover all the spec situations.

That is the easy part. We take the URL to the descriptor resource, strip the descriptor location postfix from it - and that's it. Some complexity arises from nuances of packaging and file system differences, but that's already done. We're now talking about the **validation** for possible situations (raising an exception if we detect something that isn't *explicitly* allowed by the spec). This is a bit of a feature creep for this PR, IMO.

> I'm open to interruption + it helps if we can see how the previous code would have worked. If the old code didnt cover this case, however we decide that looks, then at least we don't have a regression :/

1. JAR URLs don't allow nested archives anyway.
2. For ZIP/WSJAR URLs we (erroneously) used to strip the entry entirely, whereas now we just strip the `META-INF/persistence.xml` postfix (which should work . This is the point of this PR.

I guess the prudent thing to do here would be to check what URLs we get from different servers for nested archives, and **validate** based on that. However, this looks like a lot of work, and this PR has already taken far more time than I consider reasonable for this (seemingly trivial) fix. Can I limit its scope to just fixing the algorithm bug, and leave adding the validation (except the descriptor location check) for some other time?

dazey3: If we can skip the "jar inside war" situation, then I think your change is good as is and I'll approve it. Apologies for having this bug go further than you intended. I want to make sure this is fixed in a way that is good for all situations and that the testing is correct.

nigredo-tori: Ok, I've clarified the assumptions this method does about its arguments, and expanded the comment in question.

nigredo-tori:
I've updated the contributor lists. The copyright comments are already at year 2018.

PR #229, opened 2018-09-04
FIXES #538183 - SETTING QUERYHINTS.CURSOR ON A NAMEDQUERY THROWS QUERYEXCEPTION

rtummuru:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=538183
Executing a named-native query that uses the hint eclipselink.cursor fails
with an exception message of "Query has to be reinitialized with a cursor
stream policy."
 
It seems this happens because SQLCall.returnType in
EJBQueryImpl.databaseQuery.queryMechanism.call is still set to
RETURN_MANY_ROWS even though it should have been RETURN_CURSOR after setting
the above hint.
 
As a workaround <hint name="eclipselink.cursor" value="true"/> can be set for
the query in orm.xml (or be annotated equivalently). But this assumes the
query will only ever be used in combination with a cursored result.
 
Forcing Eclipselink to prepare the query again also fixes the problem but
couples the calling code to implementation details of Eclipselink:
((EJBQueryImpl)jpaQuery).getDatabaseQuery().setIsPrepared(false);


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=538183

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/advanced/MyTestEntity.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
```
lukasj: This is apparently new file, created in 2018 by Oracle, no need to say 1998 nor IBM...

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/QueryHintsHandler.java
```diff
@@ -355,6 +358,7 @@ static DatabaseQuery apply(String hintName, boolean shouldUseDefault, Object hin
                 query.setProperty(QUERY_HINT_PROPERTY, existingHints);
             }
             existingHints.put(hintName, hintValue);
+            query.setHintString(hintName);
```
lukasj: This is not completely wrong statement on the wrong place - why are you changing behaviour for all hints when Cursor hint is the one affected by this bug? Wouldn't 'CursorHint.applyToDatabaseQuery' be better place for this change? Why is 'query.setHintString(..)' call better then direct 'query.setIsPrepared(false)'?

rtummuru: If there is any Hint set, this is were we are setting it to true or false, but we are not setting the name of the hint including cursors.  (with my limited knowledge)setIsPrepared(false) may trigger to prepare the stmt again and may take one more cycle.

lukasj: Are you sure that Hints can have only boolean values? Furthermore the method calls 'setIsPrepared(false)' under the hood - that is the reason, why new test is passing. Have you checked what is the purpose of the  'query.setHintString(..)' method, its javadoc/what this method does and how exactly it is implemented?

According to the [javadoc](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/queries/DatabaseQuery.java#L2152-L2174), the method is not there to set the name of the hint but to provide a hint to the database through additional comment in generated SQL statement.

rtummuru:
closed to redu changes suggested.

rtummuru:
removed branch

Issue #233, opened 2018-09-12
Bug #391279: Provide support for unidirectional OneToMany with nullable=false


dazey3:
Linking to [BZ #391279](https://bugs.eclipse.org/bugs/show_bug.cgi?id=391279)

PR #234, opened 2018-09-12
Bug 391279: Add support for Unidirectional OneToMany with non-nullable value

dazey3:
for #233 

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
@@ -16,6 +16,8 @@
 //       - 350487: JPA 2.1 Specification defined support for Stored Procedure Calls
 //     08/24/2012-2.5 Guy Pelletier
 //       - 350487: JPA 2.1 Specification defined support for Stored Procedure Calls
+//     09/12/2018 - Will Dazey
+//       - 391279: Add support for Unidirectional OneToMany mappings with non-nullable values
 package org.eclipse.persistence.internal.queries;
 
 import java.io.Serializable;
@@ -489,6 +491,9 @@ public abstract class DatabaseQueryMechanism implements Cloneable, Serializable 
             // CR#3237
             // Store the size of the modify row so we can determine if the user has added to the row in the insert.
             int modifyRowSize = modifyRow.size();
+
+            //391279: check the commitManager for scheduled events that match this insert and may need to be merged
+            commitManager.checkForDataModificationEventMerges(writeQuery, session);
 
             // PERF: Avoid events if no listeners.
             if (eventManager.hasAnyEventListeners()) {
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java
@@ -14,9 +14,12 @@
 //     ailitchev - Uni-directional OneToMany
 //     07/19/2011-2.2.1 Guy Pelletier
 //       - 338812: ManyToMany mapping in aggregate object violate integrity constraint on deletion
+//     09/12/2018 - Will Dazey
+//       - 391279: Add support for Unidirectional OneToMany mappings with non-nullable values
 package org.eclipse.persistence.mappings;
 
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Vector;
 
 import org.eclipse.persistence.descriptors.ClassDescriptor;
@@ -27,6 +30,7 @@ import org.eclipse.persistence.exceptions.OptimisticLockException;
 import org.eclipse.persistence.internal.descriptors.CascadeLockingPolicy;
 import org.eclipse.persistence.internal.helper.ConversionManager;
 import org.eclipse.persistence.internal.helper.DatabaseField;
+import org.eclipse.persistence.internal.queries.ContainerPolicy;
 import org.eclipse.persistence.internal.sessions.AbstractRecord;
 import org.eclipse.persistence.internal.sessions.AbstractSession;
 import org.eclipse.persistence.internal.sessions.ChangeRecord;
@@ -39,6 +43,7 @@ import org.eclipse.persistence.queries.ObjectLevelModifyQuery;
 import org.eclipse.persistence.queries.ObjectLevelReadQuery;
 import org.eclipse.persistence.queries.ReadAllQuery;
 import org.eclipse.persistence.queries.ReadQuery;
+import org.eclipse.persistence.queries.WriteObjectQuery;
 import org.eclipse.persistence.sessions.DatabaseRecord;
 
 /**
@@ -81,13 +86,11 @@ public class UnidirectionalOneToManyMapping extends OneToManyMapping {
     @Override
     protected AbstractRecord buildKeyRowForTargetUpdate(ObjectLevelModifyQuery query){
        AbstractRecord keyRow = new DatabaseRecord();
+
        // Extract primary key and value from the source.
-       int size = sourceKeyFields.size();
-       for (int index = 0; index < size; index++) {
-           DatabaseField sourceKey = sourceKeyFields.get(index);
-           DatabaseField targetForeignKey = targetForeignKeyFields.get(index);
-           Object sourceKeyValue = query.getTranslationRow().get(sourceKey);
-           keyRow.put(targetForeignKey, sourceKeyValue);
+       for(Map.Entry<DatabaseField, DatabaseField> entry : this.sourceKeysToTargetForeignKeys.entrySet()) {
+           Object sourceKeyValue = query.getTranslationRow().get(entry.getKey());
+           keyRow.put(entry.getValue(), sourceKeyValue);
        }
        return keyRow;
    }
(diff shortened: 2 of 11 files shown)


eclipsewebmaster:
Issue tracker reference:

eclipsewebmaster:
Issue tracker reference:

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/mapping/model/PostB.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.persistence.EmbeddedId;
+import javax.persistence.Entity;
+import javax.persistence.JoinColumn;
+import javax.persistence.JoinColumns;
+import javax.persistence.OneToMany;
+
+@Entity
+public class PostB {
+	@EmbeddedId 
+//	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "POST_SQ_A")
```
jgrassel: This commented-out code necessary, or can it be removed?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
```diff
@@ -490,6 +492,9 @@ public void insertObjectForWrite() {
             // Store the size of the modify row so we can determine if the user has added to the row in the insert.
             int modifyRowSize = modifyRow.size();
 
+            //391279: check the commitManager for scheduled events that match this insert and may need to be merged
+            commitManager.checkForDataModificationEventMerges(writeQuery, session);
```
lukasj: this does not look right to me. Isn't there a way to capture the id of the owner somewhere around the place where the 'postInsert' event is being created and in case 'id' can be extracted (possibly from some objectchangeset) just add it to the row instead of fixing the wrong event further down the road?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java
```diff
@@ -390,4 +392,42 @@ public boolean shouldIncrementTargetLockValueOnDeleteSource() {
...
+    public boolean mergeDataModificationEvent(WriteObjectQuery query, Object[] event, AbstractSession session) throws DatabaseException, DescriptorException {
+        if (event[0] == PostInsert) {
+        } else if (event[0] == ObjectRemoved) {
+        } else if (event[0] == ObjectAdded) {
+            WriteObjectQuery wq = (WriteObjectQuery)event[1];
+
+            AbstractRecord queryModifyRow = query.getModifyRow();
+            AbstractRecord updateTransRow = buildKeyRowForTargetUpdate(wq);
+
+            // Look for any rows that are non-nullable, we need to add them to the query
```
lukasj: Isn't 'nullable' defined for DDL generation only?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java
```diff
@@ -390,4 +392,42 @@ public boolean shouldIncrementTargetLockValueOnDeleteSource() {
...
+            for(int pos = 0; pos < updateTransRow.getFields().size(); pos++) {
+                DatabaseField fkField = updateTransRow.getFields().get(pos);
+                if(fkField != null && !fkField.isNullable()) {
```
GordonYorke: Dusting off the old EclipseLink knowledge but is this flag set when the database is pre-existing?  I don't remember a feature in core that read in the schemas.  The code should just include all FK fields where this mapping is the owner.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
dazey3: I do agree. It is unusual to have to let the data modification event get create, only to go back and remove it later, but the problem I was hitting was the order of operations. 

The data modification events actually get added during 'postUpdate' for the UnidirectionalOneToMany side (In my examples, the Post class). During that time, we do indeed generate the 'id' for the other side (the Comment class) and it's that object that gets added to the event! However, we are still in 'postUpdate' for the UnidirectionalOneToMany mapping. We haven't even started the INSERT for the Comment object yet.

That code seems to be controlled here:

https://github.com/eclipse-ee4j/eclipselink/blob/3e83f76f7d4e7275d9d277a4375083ed0bd437d8/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/CommitManager.java#L136-L142

See how we are looping through classes? One iteration will hit the 'postUpdate' for the Post class (which generates the data modification events), but a later iteration will actually INSERT the Comment objects. It's during THAT INSERT that I am checking for any data modification events, created in the past, that should be merged into the inserting query. 

The problem is that the data modification event is created so that we remember to UPDATE COMMENT after all the classes are processed (which unfortunately will be too late). We need the data modification event to be created so that we know an UPDATE WAS scheduled and we instead want to merge it into the current INSERT going on. 

I did try looking into this route, but the problem is the order of operations. Here is where the event gets create; you are correct, it's in 'postInsert':

https://github.com/eclipse-ee4j/eclipselink/blob/3e83f76f7d4e7275d9d277a4375083ed0bd437d8/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java#L993-L1003

Unfortunately, it's during postInsert for the UnidirectionalOneToMany side (In the example, POST); we haven't started INSERTs for the other side (In the example, COMMENT). 

The order goes like this:

First, we INSERT INTO POST. This triggers the postInsert(), which says "Make sure you register that this INSERT 

https://github.com/eclipse-ee4j/eclipselink/blob/3e83f76f7d4e7275d9d277a4375083ed0bd437d8/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/CommitManager.java#L136-L142

First, we

dazey3: One thing maybe I should consider is moving my code to 'preInsert()' instead. I hadn't considered that before, but UnidirectionalOneToMany doesn't currently override 'preInsert()' which means I could just have the mapping listen for any INSERTS and check the commitManager at THAT time. Thoughts?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java
dazey3: This flag is set during orm processing. We look at the @JoinColumn annotation and set the flag based on the value in the annotation! If the database is pre-existing, I think it should be the responsibility of the user to have their orm configuration match their database. 

Essentially, during UPDATE for POST (owner of the UnidirectionalOneToMany mapping), we create a future event also UPDATE Comment (specifically the FKs). My change is intended to merge any non-nullable FKs into the INSERT for Comment. I'm doing it this way (instead of using all FKs) because we can just let the UPDATE for Comment occur as it already works. This change should only alter behavior for 'nullable=false' mappings.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
lukasj: what I've observed recently is that when I added backpointer mapping (leaving the ownership on the onetomany) for comparision, the difference was that instead of insert/update statements, just full inserts were used and the commit order computed by commitmanager seemed to be the same. I also don't think that the postInsert event was triggered in this case. That leads me to think that there should be a way to avoid doing updates (even for nullable=true cases eventually) Haven't had a chance to dig more into this due to different task unfortunately :-/

In any case, it may be good idea to introduce some property to control this.

dazey3: I tried adding a backpointer mapping. Unfortunately, I could not get around the fact that the mapping has no accessor and no actual field on the object. There are 101 places in EclipseLink where we use the field name (keys in maps for instance) that lead me to not being able to get around that limitation.

dazey3:
I thought of a new test that I wasn't covering and added the test. It failed, so I needed to add the *PostInsert* event coverage.

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/mapping/TestUnidirectionalOneToMany.java
```diff
@@ -0,0 +1,137 @@
...
+    @Test
+    public void testInsertOneToManyUni() {
+
+        EntityManager em = emf.createEntityManager();
+
+        em.getTransaction().begin();
+
+        PostA post = new PostA(new ComplexIdA(9));
+        em.persist(post);
+        em.flush();
```
lukasj: If this persist/flush sequence is removed and one relies on cascading to occur during the next persist, then there is still insert/update occurring - is this expected?

dazey3: Ok, I updated to reflect what it is supposed to be testing. Essentially my tests hit two situations: 
1. 'PostInsert': Post entity is being inserted WITH it's Comment relationships. This causes the data modification events to be created on Post's 'postInsert'. 
2. 'ObjectAdded': Post entity is being updated to ADD it's Comment relationships. This causes the data modification events to be created on Post's 'postUpdate'.

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/mapping/TestUnidirectionalOneToMany.java
```diff
@@ -0,0 +1,131 @@
...
+import org.eclipse.persistence.jpa.test.mapping.model.PostA;
+import org.eclipse.persistence.jpa.test.mapping.model.PostB;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(EmfRunner.class)
+public class TestUnidirectionalOneToMany {
+    @Emf(createTables = DDLGen.DROP_CREATE, 
+            classes = { CommentA.class, CommentB.class, ComplexIdA.class, ComplexIdB.class, PostA.class, PostB.class }, 
+            properties = { @Property(name = "eclipselink.cache.shared.default", value = "false") })
```
lukasj: when I add

```
@Property(name = "eclipselink.logging.parameters", value = "true"),
@Property(name = "eclipselink.logging.level.sql", value = "FINEST")
```

here, in the log, there I can see:
```
[EL Fine]: sql: 2018-09-17 13:54:42.12--ClientSession(1835848160)--Connection(460674260)--INSERT INTO COMMENTA (ID, POSTA_FK) VALUES (?, ?)
	bind => [4, 9]
[EL Fine]: sql: 2018-09-17 13:54:42.121--ClientSession(1835848160)--Connection(460674260)--UPDATE COMMENTA SET POSTA_FK = ? WHERE (ID = ?)
	bind => [9, 4]
```

I'd expect, that after this patch, the update statement won't be executed

dazey3: Ya, I'm having troubles with the 'postInsert' event. See, what happens is that 'postInsert' for Post, we create a 'PostInsert' event, but it's just meant for the Post object being inserted! For 'PostUpdate' we have one event PER Comment object, so it's pretty easy to remove from the List of data modification events. 
If I remove the whole 'PostInsert' event for the first Comment, then the rest will not know there was an event! Causing the rest of the INSERT COMMENT to fail.

dazey3: Ok @lukasj , I think I fixed this issue by reworking how 'PostInsert' works. I changed it now so that an individual 'PostInsert' event gets created for EACH Comment object. The current way it works is that only 1 'PostInsert' event gets created and during execution (still on transaction commit), we loop through all the relationship objects and UPDATE each one. 
Now, we create an event for each relationship object and each event just runs for that one object. I think this basically has the same logic, but I'd appreciate an objective look and thoughts/tests I don't have.

PR #242, opened 2018-09-22
[2.7] Bug 391279: Improve implementation

lukasj:
@dazey3 what about following approach?

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
@@ -1091,10 +1091,10 @@ public class OneToManyMapping extends CollectionMapping implements RelationalMap
                             query.getSession().executeQuery(insertQuery);
                         }
                     }
-                } else {
-                    if ((requiresDataModificationEvents() || containerPolicy.shouldUpdateForeignKeysPostInsert())) {
-                        updateTargetRowPostInsertSource(query);
-                    }
+                }
+            } else {
+                if ((requiresDataModificationEvents() || containerPolicy.shouldUpdateForeignKeysPostInsert())) {
+                    updateTargetRowPostInsertSource(query);
                 }
             }
         }
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
@@ -16,23 +16,50 @@
 //       - 338812: ManyToMany mapping in aggregate object violate integrity constraint on deletion
 package org.eclipse.persistence.mappings;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Vector;
 
-import org.eclipse.persistence.exceptions.*;
-import org.eclipse.persistence.expressions.*;
-import org.eclipse.persistence.internal.helper.*;
-import org.eclipse.persistence.internal.identitymaps.*;
-import org.eclipse.persistence.internal.queries.*;
-import org.eclipse.persistence.internal.sessions.*;
-import org.eclipse.persistence.queries.*;
-import org.eclipse.persistence.sessions.DatabaseRecord;
-import org.eclipse.persistence.internal.descriptors.CascadeLockingPolicy;
 import org.eclipse.persistence.descriptors.ClassDescriptor;
+import org.eclipse.persistence.exceptions.ConversionException;
+import org.eclipse.persistence.exceptions.DatabaseException;
+import org.eclipse.persistence.exceptions.DescriptorException;
+import org.eclipse.persistence.exceptions.OptimisticLockException;
+import org.eclipse.persistence.expressions.Expression;
+import org.eclipse.persistence.expressions.ExpressionBuilder;
+import org.eclipse.persistence.internal.descriptors.CascadeLockingPolicy;
 import org.eclipse.persistence.internal.descriptors.ObjectBuilder;
 import org.eclipse.persistence.internal.expressions.FieldExpression;
 import org.eclipse.persistence.internal.expressions.ParameterExpression;
 import org.eclipse.persistence.internal.expressions.SQLUpdateStatement;
+import org.eclipse.persistence.internal.helper.ConversionManager;
+import org.eclipse.persistence.internal.helper.DatabaseField;
+import org.eclipse.persistence.internal.helper.DatabaseTable;
+import org.eclipse.persistence.internal.identitymaps.CacheId;
+import org.eclipse.persistence.internal.identitymaps.CacheKey;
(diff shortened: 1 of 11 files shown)


eclipsewebmaster:
Issue tracker reference:

eclipsewebmaster:
Issue tracker reference:

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
```diff
@@ -982,19 +1045,56 @@ public void postInsert(WriteObjectQuery query) throws DatabaseException, Optimis
...
+                            if (listOrderField != null) {
+                                databaseRow.put(listOrderField, objectIndex++);
+                            }
+                            InsertObjectQuery insertQuery = getInsertObjectQuery(session, referenceDesc);
+                            insertQuery.setObject(object);
+                            insertQuery.setCascadePolicy(query.getCascadePolicy());
+                            insertQuery.setTranslationRow(databaseRow);
+                            insertQuery.setModifyRow(databaseRow);
+                            insertQuery.setIsPrepared(false);
+                            query.getSession().executeQuery(insertQuery);
```
dazey3: Interesting, I tried something similar to this, but I struggled with was this portion of logic: https://github.com/eclipse-ee4j/eclipselink/blob/c63079c2e0b2c9bb568698abc313bb5f84f5acde/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/CommitManager.java#L139-L142

Here, we are looping through the classes, in order (Post, Comment, ect), and execute any inserts for new objects. This means, we first INSERT POST, then create any data modification events (postInsert events for instance). Then we INSERT COMMENT objects in the second iteration. I kept hitting the issue that this sort of change would end up calling INSERTs for COMMENT twice... However, I just tried this change and I am not getting that issue like I had been -_-' 

I'm wondering if we can just update 'updateTargetRowPostInsertSource()' to work like this and then change the logic so that 'updateTargetRowPostInsertSource()' can just be called. I think the main difference might be `insertQuery.setObject(object);`? That removes the objects from the uowChangeSet??

lukasj: Or better - the object is marked as committed during the first insert, so the commit manager knows it should not commit it again

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
```diff
@@ -884,13 +918,42 @@ protected void objectAddedDuringUpdate(ObjectLevelModifyQuery query, Object obje
         if (requiresDataModificationEvents() || containerPolicy.requiresDataModificationEvents()){
             // In the uow data queries are cached until the end of the commit.
             if (query.shouldCascadeOnlyDependentParts()) {
```
dazey3: This implementation is so close to updateTargetForeignKeyPostUpdateSource_ObjectAdded() that I wish we could just have:

``` java
if (query.shouldCascadeOnlyDependentParts() && shouldDeferInsert()) {
    Object[] event = new Object[4];
    event[0] = ObjectAdded;
    event[1] = query;
    event[2] = objectAdded;
    event[3] = extraData;
    query.getSession().getCommitManager().addDataModificationEvent(this, event);
} else {
    updateTargetForeignKeyPostUpdateSource_ObjectAdded(query, objectAdded, extraData);
}
```
I don't understand the `OneToManyMapping.addTargetQuery` object enough tho

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
dazey3: I do like this implementation better than mine. I am just looking it over to see if there is anything else to note, but I think we should go with this.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
lukasj: I'd like to avoid code duplication as well but in this case it is non trivial and would very likely mean casts and ifs within the for loop - ie common parent for insert/datamodify queries is ModifyQuery which does not allow setting object on it, another thing is that `executeQuery(query, databaseRow)` apparently works differently for Insert and DataModify. Lambdas could safe us here but since we need this also in 2.6 (jdk7), that is not an option. At least not right now. We can improve this later...

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java
```diff
@@ -982,19 +1045,56 @@ public void postInsert(WriteObjectQuery query) throws DatabaseException, Optimis
...
+                            InsertObjectQuery insertQuery = getInsertObjectQuery(session, referenceDesc);
+                            insertQuery.setObject(object);
+                            insertQuery.setCascadePolicy(query.getCascadePolicy());
+                            insertQuery.setTranslationRow(databaseRow);
+                            insertQuery.setModifyRow(databaseRow);
+                            insertQuery.setIsPrepared(false);
+                            query.getSession().executeQuery(insertQuery);
+                        }
+                    }
+                } else {
```
dazey3: wait, did this get moved on purpose? This if-else is different now.

lukasj: true, if you check full method, it starts with `if (isReadOnly()) { return;}`[see](https://github.com/eclipse-ee4j/eclipselink/blob/8704f7fc17699bb6b02a5381e101a7a9f377fbc4/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/OneToManyMapping.java#L1015), so later call to `!isReadOnly()` should be always evaluated to `true`. Do you think there exists a case where this would be evaluated to `false`?

dazey3: Oh, I don't mean the `isReadOnly()` check. I think that should be fine since the method is prefaced with:

``` java
        if (isReadOnly()) {
            return;
        }
```
I mean the different if-else structure.

lukasj: ouch, fixed. Thanks!

lukasj:
This may be related to the tests related changes I did in master some time ago - classpath for running tests as well as built test jars are different in 2.7 and master. Can you file a bug for me to track this down, please?

lukasj:
I added system property to allow enforcing 'old' behaviour in case of issues in existing apps

dazey3:
I dont think we should have a property to revert behavior on this one. How EclipseLink handles Unidirectional 1-M, non-nullable values is obviously wrong atm. If this was a grey area or implementation dependent, I'd agree, but we are fixing a bug with this one. Right?

lukasj:
I noticed comments related to possible deadlocks in caches - not sure how likely they can occur but having fallback option for such case won't hurt, I think.

lukasj:
I treated your last comment as a go for this change, so we can close this... Hope you don't mind, Will.

dazey3:
Oh yes, that's fine. I forgot to approve this review. Thanks Lukas

PR #243, opened 2018-09-24
Bug 539323 - Converter with autoApply=true flag is applied to the @Id, @Version, @Enumerated, @Temporal fields

rfelcman:
Bug 539323 - Converter with autoApply=true flag is applied to the @Id, @Version, @Enumerated, @Temporal fields.

This is fix for JPA custom converter used/applied to the entities via autoApply “@Converter(autoApply=true)” annotation.
JPA specification 2.2 says:
Note that Id attributes, version attributes, relationship attributes, and attributes explicitly annotated as Enumerated or Temporal (or designated as such via XML) will not be converted.

But EclipseLink calls custom converter against these fields too.
After this fix EclipseLink will not apply custom converter with “@Converter(autoApply=true)” annotation to these fields.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=539323


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=539323

PR #248, opened 2018-09-27
Bug 411013: Entities in persistene context get validated on commit ev…

lukasj:
…en when unchanged

### DIFF
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
@@ -14,18 +14,28 @@
 //      Marcel Valovy
 package org.eclipse.persistence.testing.tests.jpa.beanvalidation;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
 
-import javax.persistence.*;
-import javax.validation.ConstraintViolationException;
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.TypedQuery;
 import javax.validation.ConstraintViolation;
+import javax.validation.ConstraintViolationException;
 
-import junit.framework.*;
-
-import org.eclipse.persistence.testing.framework.junit.JUnitTestCase;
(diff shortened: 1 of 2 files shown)


jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -355,7 +366,34 @@ public void testTraversableResolverPreventsTraversingRelationshipMultipleTimes()
         // If it ever finds an entity with such flag set, the entity has visited the validator twice. It should be flagged as error.
     }
 
-
+    //Bug #411013
+    public void testValidateChangedData() {
+        try {
+            getDatabaseSession().executeNonSelectingSQL("insert into CMP3_BV_PROJECT values (895, \"some long name\")");
+        } catch (Throwable t) {
+            //ignore
```
dazey3: If the INSERT fails (other than a duplicate key exception), doesn't that throw off the rest of the test?

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -355,7 +366,34 @@ public void testTraversableResolverPreventsTraversingRelationshipMultipleTimes()
...
+        props.put("eclipselink.weaving", "false");
+        EntityManagerFactory factory = getEntityManagerFactory(props);
+        EntityManager em = factory.createEntityManager();
+        try {
+            beginTransaction(em);
+            TypedQuery<Project> query = em.createQuery("select p from CMP3_BV_PROJECT p", Project.class);
+            for (Project p: query.getResultList()) {
+                System.out.println(p.getName());
+            }
+            commitTransaction(em);
```
dazey3: I'm not sure I understand the test. Validation is happening on transaction commit. Does the validation throw an exception?

lukasj: Yes, exception is thrown but should not be as no entity has been changed.

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
lukasj: Failed 'testSetup' would lead to the same situation..

PR #266, opened 2018-10-17
Bug411013 - 2nd attempt

lukasj:
Original fix in #248 caused regressions in core tests (as can be seen on nightly downloads page :-/ )

Problem with the original fix is that it did not take into account that user may have custom event listener bound to say 'preUpdate' event where he wants to alter to be modified object somehow. In this case, it is expected that the update event is fired even for unchanged objects and that is not happening after my original fix :-/

Better fix seems to be to bind internal bean validation 'preUpdate' check to the 'aboutToUpdateEvent' - this event is fired only for to be updated objects just before sending the update statement to the DB. In this case, it is also expected to be guaranteed that all user-defined listeners, even those altering the object being changed in ie 'preUpdateWithChangesEvent', were run and validation runs against the real, up-to-date to be updated object.

This also revealed one issue in travis test runs where it did not fail the build if there were failures/errors in core/jpa tests, that is fixed by the second commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ addons:
 
 env:
   global:
-    - ANT_HOME=$HOME/apache-ant-1.10.2
+    - ANT_HOME=$HOME/apache-ant-1.10.5
     - M2_HOME=/usr/local/maven-3.5.2
   matrix:
     - TEST_TARGET=test-core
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=411013

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=411013

PR #276, opened 2018-10-30
Fixes #273: poms should not contain optional dependencies

### LINKED ISSUE #273: org.eclipse.persistence.jpa depends on javax.json-api but doesn't use it
Hi,

I am migrating a Maven project from OpenJPA to EclipseLink JPA and I saw something which looked weird to me, so I thought I should report it. I added the following dependencies to my project's pom.xml:
```
            <dependency>
                <groupId>org.eclipse.persistence</groupId>
                <artifactId>org.eclipse.persistence.jpa</artifactId>
                <version>2.7.3</version>
            </dependency>
            <dependency>
                <groupId>org.eclipse.persistence</groupId>
                <artifactId>org.eclipse.persistence.extension</artifactId>
                <version>2.7.3</version>
            </dependency>
```
Unfortunately, due to my company's policy, I had to document all new dependencies (including transitive ones), so I went through them one by one. I saw that the `org.eclipse.persistence.jpa` artifact has a compile dependency to `javax.json-api` and I did a quick search through the code, but I couldn't find any usage of the `javax.json` package. I also tried excluding it and nothing broke. Is this dependency really necessary then? Removing it might save some time for other people in my situation.

lukasj:
fixes #273

### DIFF
--- a/uploadToNexus.xml
+++ b/uploadToNexus.xml
@@ -74,7 +74,7 @@
 
     <!-- Tool version definitions -->
     <property name="gpg.plugin.version"   value="1.4"/>
-    <property name="wagon.http.version"   value="2.4"/>
+    <property name="wagon.http.version"   value="2.6"/>
 
     <!-- The repository info of where maven artifacts will be uploaded -->
     <!--    Override value to change defaults                          -->
@@ -175,13 +175,6 @@
         <property name="sdoapi.prefix"        value="commonj.sdo"/>
         <property name="sdoapi.criteria"      value="[2.1.1,2.2.0)"/>
         <property name="sdoapi.name"          value="SDO API"/>
-        <property name="validation.prefix"    value="validation-api"/>
-        <property name="validation.criteria"  value="[1.1.0,2.2.0)"/>
-        <property name="validation.name"      value="Javax Validation"/>
-        <property name="json.prefix"          value="javax.json-api"/>
-        <property name="json.criteria"        value="[1.1.0,5.0.0)"/>
-        <property name="json.name"            value="Javax JSON"/>
-        <property name="json.version"         value="1.1.2"/>
         <property name="antlr.prefix"         value="org.eclipse.persistence.antlr"/>
         <property name="antlr.criteria"       value="[3.0.0,4.0.0)"/>
         <property name="antlr.name"           value="EclipseLink ANTLR"/>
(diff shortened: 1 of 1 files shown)


uploadToNexus.xml
```diff
@@ -327,18 +309,18 @@
         <!--   canonical modelgen processor during compilation. It has no real dependencies, nor is anything really       -->
         <!--   dependant upon it, but it is needed for this compiler functionality to work. (think of it as a compiler    -->
         <!--   add-on). It will need to be downloaded independently for this feature (modelgen) to work.                  -->
-        <property name="eclipselink.dependencies" value="${deps.head}${dep.persistence}${dep.sdoapi}${dep.validation}${dep.json}${deps.foot}"/>
+        <property name="eclipselink.dependencies" value="${deps.head}${dep.persistence}${dep.sdoapi}${deps.foot}"/>
         <!-- would need the modelgen.jar for modelgen but cannot include dep -->
         <property name="jpql.dependencies"        value="${deps.head}${dep.hermes}${deps.foot}"/>
-        <property name="core.dependencies"        value="${deps.head}${dep.asm}${dep.json}${deps.foot}"/>
```
dazey3: Core still depends on javax.json for compile, yes?

uploadToNexus.xml
```diff
@@ -327,18 +309,18 @@
...
+        <property name="core.dependencies"        value="${deps.head}${dep.asm}${deps.foot}"/>
         <property name="oracle.dependencies"      value="${deps.head}${dep.persistence}${dep.asm}${dep.core}${deps.foot}"/>
-        <property name="jpa.dependencies"         value="${deps.head}${dep.persistence}${dep.asm}${dep.antlr}${dep.json}${dep.hermes}${dep.core}${deps.foot}"/>
+        <property name="jpa.dependencies"         value="${deps.head}${dep.persistence}${dep.asm}${dep.antlr}${dep.hermes}${dep.core}${deps.foot}"/>
```
dazey3: Doesn't org.eclipse.persistence.jpa compile with javax.validation?

uploadToNexus.xml
lukasj: Yes and that dependency comes from manifest in the core project (where it is optional), not from pom.xml. This pom and the pom you see in the project sources are different and used for different things - this one is for artifacts in maven repo (see the building el wiki for how to check this locally) and the latter is for "tycho magic"

uploadToNexus.xml
lukasj: I think it does but again it is not mandatory for runtime and typically comes from the environment - if user wants beanval then he needs to add it on the classpath explicitly

PR #290, opened 2018-11-21
 [nobug] Mavenize compilation dependencies

rfelcman:
This is change of compilation dependencies.
EclipseLink compilation dependencies are generated dynamically by Maven/Tycho.
There are two submodules.
1. EclipseLink Internal Compilation Dependencies (_buildsystem/compdeps/org.eclipse.persistence.compdeps.internal/pom.xml_): to generate P2 repository with internal dependencies (ASM, ANTLR, Oracle DDL Parser, EL JPA API).
    This submodule is called from Ant target _generate-internal-compdeps_ .
2. EclipseLink CompDeps (_buildsystem/compdeps/pom.xml_): to generate P2 repository with rest/external dependencies used by EL and EL tests.
    This submodule is called from Ant target _generate-local-compdeps_ .

Ant target _generate-local-compdeps_ will copy jar files from local P2 repository to the plugins directory. Before copy operation it will delete files in plugins directory by _clean-plugins_ target.


lukasj:
needs few tweaks (ie to run by default if none is created etc) and it will be fine; let's walk through this tomorrow.

@dazey3 we're about to change the way compile time dependencies are being managed and resolved and we need to do this in 2.7 too.

The general idea here is to use maven repositories as the primary source of artifacts, local filesystem as secondary (ie to provide various jdbc drivers and/or proprietary files) and current orbit and/or update sites as the last resort to compose local-only update site which existing tycho build will be able to use with minimal changes.

lukasj:
to sum up the discussion - it is required for the change to be transparent to existing travis/hudson builds:

* existing binaries being downloaded from maven should be removed/deleted from current {., jpa, sdo, utils}/plugins folders
* (optional) dependency on local presence of eclipse-SDK should be removed
* travis configuration should be extended to run also 'ant -f antbuild.xml build-distribution'
* extended travis run has to pass and tests should show no errors

Next steps:
* adopt jakarta dependencies:
  - changes in maven coordinates are captured at https://wiki.eclipse.org/New_Maven_Coordinates
  - versions of components to be used are at https://wiki.eclipse.org/Eclipse_GlassFish_5.1_Components_Release_Tracker
  - URL to oss sonatype staging repo containing required artifacts can be found ie at https://wiki.eclipse.org/EclipseLink/Maven
  - code/tests may need to be adjusted to new APIs
  => travis has to pass
* clean up:
  - remove the hacky part around copying and playing with filenames for tests - update test scripts/properties files to look for what they need at some given location (maven repo, folder with local update site having all those jars, some other folder (?) )

I don't think everything has to be done at once and would prefer to split this work through multiple PRs as long as we don't end up with broken build after each PR

lukasj:
```
    [junit] Running org.eclipse.persistence.testing.tests.logging.slf4j.SLF4JLoggerTest
    [junit] Tests run: 4, Failures: 0, Errors: 4, Skipped: 0, Time elapsed: 0.169 sec
    [junit] Test org.eclipse.persistence.testing.tests.logging.slf4j.SLF4JLoggerTest FAILED
```

.travis.yml
```diff
@@ -30,7 +30,8 @@ env:
   global:
     - ANT_HOME=$HOME/apache-ant-1.10.5
     - M2_HOME=/usr/local/maven-3.5.2
-  matrix:
+    - MAVEN_OPTS=-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+matrix:
```
lukasj: you need to keep the formatting, number of spaces is important in yaml format

rfelcman:
There is still trouble with javadoc JDK 10 (task build-nosql-javadocs).
It ignores -Xdoclint:none switch and javadoc is not generated.

PR #298, opened 2018-11-27
[master] Bug 541391 - Cascading detach does not work at uninstantiated indirection objects

ghost:
…tion objects


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=541391

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=541391

robertpatrick:
@arjantijms there’s still a real issue here, but the current approach can lose pending updates and needs reworking before merging.

PR #2037 already fixed lazy collection detachment. A lazy `ManyToOne` annotated collection still returns a managed target when accessed after its owner is detached, so this PR is only partially superseded.

| ---- | ---- | ---- |
| Already-managed child in an uninitialized collection; detach parent | Child detached immediately | Child remains managed until collection accessed |
| Access lazy `ManyToOne` annotated collection after detaching owner | Returned target remains managed | Returned target detached |
| Find and modify child after owner detachment, then access old owner’s collection or reference | Child remains managed; update persisted | Child unexpectedly detached; update not persisted |
| Preloaded relationships and non-cascading collection controls | Correct | Unchanged |

The lost update happens because the delayed-detach flag acts on whichever object the later lazy load returns—including an entity independently loaded and modified after the original detach.

What we need:

| Action | Reason |
| ---- | ---- |
| Rebase while preserving #2037’s collection handling | Avoid undoing the newer detach behavior and cycle protection |
| Rework the remaining lazy-reference fix | Reading a detached object must not detach independently managed entities or discard their updates |
| Add focused lifecycle regressions and retain existing detach tests | Check contains() immediately after detach, plus the independent find/modify/access/flush sequence; the PR adds no tests |
| Obtain passing existing CI | Current Jenkins status is “This commit cannot be built”; no passing CI evidence |

The remaining issue is worth addressing, but a rebase alone would not make this implementation safe to merge.

These were controlled JPA/H2 tests, including woven lazy references.

PR #300, opened 2018-12-03
SQL Server: if possible, use OFFSET and FETCH instead of JDBC methods

rdicroce:
Currently, EclipseLink uses Statement#setMaxRows() to apply a result limit when the target database is MS SQL Server. This is dangerous because mssql-jdbc implements setMaxRows() by executing SET ROWCOUNT, which applies to all queries made on the connection until SET ROWCOUNT is executed again.

Normally, that is not an issue, but it can become a problem if, for example, your query includes a call to a user-defined function that executes another query. When that happens, it creates truly baffling bugs that are very difficult to track down. mssql-jdbc acknowledges that SET ROWCOUNT is not an appropriate way to implement setMaxRows() but currently has no ability to replace it with something better; see https://github.com/Microsoft/mssql-jdbc/issues/176.

This PR alters SQLServerPlatform to add the OFFSET and FETCH NEXT clauses to the query, if the version of SQL Server in use supports them. This avoids EclipseLink calling setMaxRows(), and may improve query performance.

This PR is for 2.7. If it looks okay to whoever reviews it, I will open another PR for master.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/SQLServerPlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/SQLServerPlatform.java
@@ -45,8 +45,8 @@ import org.eclipse.persistence.queries.*;
  * @since TOPLink/Java 1.0
  */
 public class SQLServerPlatform extends org.eclipse.persistence.platform.database.DatabasePlatform {
-    /** Support for sequence objects added in SQL Server 2012 */
-    private boolean supportsSequenceObjects;
+    /** Support for sequence objects and OFFSET FETCH NEXT added in SQL Server 2012 */
+    private boolean isVersion11OrHigher;
     private boolean isConnectionDataInitialized;
 
     public SQLServerPlatform(){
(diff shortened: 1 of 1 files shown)


lukasj:
create a PR for master, please. Thanks!

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/SQLServerPlatform.java
```diff
@@ -732,4 +732,47 @@ public void writeUpdateOriginalFromTempTableSql(Writer writer, DatabaseTable tab
...
+        if (max == 0 && first == 0) {
+            super.printSQLSelectStatement(call, printer, statement);
+            return;
+        }
+        
+        // OFFSET + FETCH NEXT requires ORDER BY, so add an ordering if there are none
+        // this SQL will satisfy the query parser without actually changing the ordering of the rows
+        List<Expression> orderBy = statement.getOrderByExpressions();
+        if (orderBy.isEmpty()) {
+            orderBy.add(statement.getBuilder().literal("ROW_NUMBER() OVER (ORDER BY (SELECT null))"));
```
kaizimmerm, line 756: Hi, 

this change unfortunately causes error when the the query is using distinct
`com.microsoft.sqlserver.jdbc.SQLServerException: ORDER BY items must appear in the select list if SELECT DISTINCT is specified.`

Maybe its better not to try to trick the query parser but step out here as well, i.e.:

```java
super.printSQLSelectStatement(call, printer, statement);
return;
```

PR #284, opened 2018-11-08
[master] Bug 540929: jdbc.sql-cast property does not apply

dazey3:
for #283 
Signed-off-by: Will Dazey <dazeydev.3@gmail.com>

Issue #283, opened 2018-11-08
Bug #540929: Persistence Property 'eclipselink.jdbc.sql-cast' doesn't apply


dazey3:
Linking to [BZ #540929](https://bugs.eclipse.org/bugs/show_bug.cgi?id=540929)

PR #303, opened 2018-12-05
[master] Bug 540929: jdbc.sql-cast property does not apply - fix

dazey3:
for #283

I think the fix for #284 was incorrect. I stumbled across DatabasePlatform.copyInto() and realized that this is where the Cast value is supposed to be copied over at. 


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=540929

PR #309, opened 2018-12-14
SQL Server: if possible, use OFFSET and FETCH instead of JDBC methods

rdicroce:
Same as PR #300, except for master branch.

I'm about to go on vacation, so if there's a problem with this PR, I won't be around to look at it until January.


lukasj:
can you add sign-off to this, please? I can not merge this without it. Thanks!

rdicroce:
Whoops. Should be all set now.

lukasj:
Thanks!

PR #310, opened 2018-12-14
Bug 391279: fix the case when existing object is being updated

lukasj:
during update of unidirectional 1-m relationship only relationship constraint related fields are being inserted to the DB, so if there are additional fields then these additional fields never get stored into DB.


eclipsewebmaster:
Issue tracker reference:

PR #311, opened 2018-12-17
 Bug 391279: fix the case when existing object is being updated 

lukasj:


eclipsewebmaster:
Issue tracker reference:

PR #344, opened 2018-12-21
fix the input for jpaapi version property

lukasj:


oraboliu:
i'm ok with this change

Issue #362, opened 2019-01-25
Bug 543846: Stored Procedure named parameter with DB2/Z doesn't work


dazey3:
Linking to [BZ #543846](https://bugs.eclipse.org/bugs/show_bug.cgi?id=543846)

PR #363, opened 2019-01-25
[master] Issue 543846: Add DB2z support for named parameters with stored proce…

dazey3:
for #362

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/DB2ZPlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/DB2ZPlatform.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2015 IBM Corporation. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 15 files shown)


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/platform/database/DB2ZPlatform.java
```diff
@@ -50,4 +82,394 @@ public String getTableCreationSuffix() {
...
+                    ClassLoader cl = AccessController.doPrivileged(new PrivilegedGetContextClassLoader(Thread.currentThread()));
+                    clazz = AccessController.doPrivileged(new PrivilegedClassForName(DB2_CALLABLESTATEMENT_CLASS, true, cl));
+                    method = AccessController.doPrivileged(new PrivilegedGetMethod(clazz, methodName, methodArgs, true));
+                    Object o = statement.unwrap(clazz);
+                    AccessController.doPrivileged(new PrivilegedMethodInvoker(method, o, parameters));
+                } catch (PrivilegedActionException ex) {
+                    if (ex.getCause() instanceof ClassNotFoundException) {
+                        throw (ClassNotFoundException) ex.getCause();
+                    }
+                    throw (RuntimeException) ex.getCause();
```
jgrassel: Not sure if I'm a fan of just rethrowing a RuntimeException -- the caller might not only be expecting to handle SQLExceptions.  Yet just wrapping it within a SQLException isn't the answer either.

dazey3:
Oh, I hadn't noticed before, but the Oracle stuff depends on the Session object being passed into getConnection(). So I reverted that part of my change. It's unfortunate that if I break that stuff nothing fails in the build cause it's already compiled.

PR #367, opened 2019-02-07
[2.7] Issue 543846: Add DB2z support for named parameters with stored proce…

dazey3:
for #362

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/BindCallCustomParameter.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/BindCallCustomParameter.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -18,6 +19,7 @@ import java.io.Serializable;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Array;
+import java.sql.CallableStatement;
 import java.sql.Struct;
 import java.sql.Types;
 import java.util.Arrays;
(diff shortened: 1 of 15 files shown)


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/BindCallCustomParameter.java
```diff
@@ -71,6 +73,10 @@ public void set(DatabasePlatform platform, PreparedStatement statement, int inde
         platform.setParameterValueInDatabaseCall(obj, statement, index, session);
     }
 
+    public void set(DatabasePlatform platform, CallableStatement statement, String name, AbstractSession session) throws SQLException {
```
lukasj: would be good to add some javadoc here

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/InOutputParameterForCallableStatement.java
```diff
@@ -51,11 +52,18 @@ public InOutputParameterForCallableStatement(Object inParameter, DatabaseField o
...
     }
 
+    @Override
     public void set(DatabasePlatform platform, PreparedStatement statement, int index, AbstractSession session) throws SQLException {
         platform.setParameterValueInDatabaseCall(inParameter, statement, index, session);
         super.set(platform, statement, index, session);
     }
 
+    @Override
+    public void set(DatabasePlatform platform, CallableStatement statement, String index, AbstractSession session) throws SQLException {
```
lukasj: calling String arg here `index` looks confusing to me

dazey3:
@lukasj 
>One thing which is not clear to me - new methods do take AbstractSession as an argument and the only usage I've found is to log an exception. If this is correct, is this really needed? Cannot global AbstractSessionLog.getLog be used or some kind of recoverable exception be thrown instead?

Correct, I passed AbstractSession down the call chain so that I can log warnings. Also, there was already many existing calls in the INTERNAL api that pass AbstractSession down, so I figured it was fine to add. I didn't know a static call like `AbstractSessionLog.getLog()` was available! I updated the PR to reflect this change.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/InOutputParameterForCallableStatement.java
dazey3: I agree, ty for noticing this. Ill go back through and change the var names to be "int parameterIndex" and "String parameterName". Does that sound good?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/BindCallCustomParameter.java
dazey3: Added :) thanks!

Issue #345, opened 2018-12-27
JPQL query can not be parsed in Turkish locale


andreysubbotin:
if the query contains LIKE or IS NULL condition, it can not be executed in Turkish locale. 
Project with test: `com.company.jpa.QueryTest`
[test-jpa.zip](https://github.com/eclipse-ee4j/eclipselink/files/2712256/test-jpa.zip)

Error: 
```
Exception Description: Syntax error parsing [select u from sec$User u where u.loginLowerCase like :login]. 
[31, 59] The expression is not a valid conditional expression.
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:155)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.validate(HermesParser.java:334)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.populateQueryImp(HermesParser.java:278)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildQuery(HermesParser.java:163)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.buildEJBQLDatabaseQuery(EJBQueryImpl.java:142)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.buildEJBQLDatabaseQuery(EJBQueryImpl.java:116)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.<init>(EJBQueryImpl.java:102)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.<init>(EJBQueryImpl.java:86)
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1614)
	... 27 more
```

lukasj:
fixed in master and 2.7.5

PR #370, opened 2019-02-09
Bug 527415 #345 - Fix code when locale is tr, az or lt

ajacob:
The bug from #345 is not the same that was reported in 527415, I corrected both of them.
Changed more code that I should have but tried to be defensive to avoid other similar bugs.

This blog post gives a good explanation of the problem : https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/

PS : Is it mandatory to add ourself to Contributors at the beggining of each file ?


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=527415

eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=527415

jpa/org.eclipse.persistence.jpa.jpql.test/src/org/eclipse/persistence/jpa/tests/jpql/parser/JPQLExpressionTurkishLocaleTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.eclipse.persistence.jpa.jpql.parser.JPQLExpression;
+import org.eclipse.persistence.jpa.jpql.parser.JPQLGrammar;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JPQLExpressionTurkishLocaleTest {
+
+    @Test
+    public void testJPQLExpressionWithTurkishLocale() {
+        Locale.setDefault(new Locale("tr", "TR"));
```
lukasj: test should reset this to default when it's done so other tests, if run in the same VM are not affected

andreysubbotin:
Any updates?

PR #375, opened 2019-02-28
fix versions in readme and about  files

lukasj:


rfelcman:
Looks good.

PR #377, opened 2019-03-01
Bug_544934 - SERVICE FAILING WITH JAVA.LANG.REFLECT.INVOCATIONTARGETE…

sureshbn:
…XCEPTION


sureshbn:
Updated the Copyright

Test Case
----------
Not feasible because it replicate only in cloud integration fleet intermittently . This flow goes though various systems/products and finally reaching EL. Provided a Patch with this (above) fix in replica environment and works good as per comments/feedback

PR #378, opened 2019-03-03
Bug 544995 : Fixed batch fetching when shared cache is activated

ajacob:
Here is a fix for this multithreading issue.

Applied in production, this problem could occur from 10 to 30 times a day and is now fixed.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/ForeignReferenceMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/ForeignReferenceMapping.java
@@ -23,6 +23,8 @@
 //       - 389090: JPA 2.1 DDL Generation Support (foreign key metadata support)
 //     08/12/2015-2.6 Mythily Parthasarathy
 //       - 474752: Address NPE for Embeddable with 1-M association
+//     04/19/2020-3.0 Alexandre Jacob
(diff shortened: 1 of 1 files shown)


ajacob:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=544995

robertpatrick:
@ajacob @arjantijms the fix looks sound, but needs rebase, regression coverage and clean CI before merging.

Another thread can populate the shared cache between two checks, leaving the batch with nothing to fetch. The patch skips that unnecessary execution while preserving the cached relationships.

| Local test | Current master |  With the guard |
| ---- | ---- | ---- |
| Composite-key cache race | Invalid IN (); SQL exception | Correct relationships returned; no batch SQL |
| Scalar-key cache race | Correct, but executes unnecessary SQL | Same results without that query |
| Cold, warm and mixed cache; paging | Correct | Unchanged |
| Refresh after database changes | Returns updated data | Unchanged |

What we need:

| Action | Reason |
| ---- | ---- |
| Rebase while preserving current per-batch query cloning | Avoid undoing the newer nested-fetch fix |
| Add a deterministic race regression, including composite keys | Assert both skipped execution and correct relationships; the PR adds no tests |
| Obtain passing existing CI | Currently conflicting, 930 commits behind, with no current-head checks/statuses reported |

No production redesign is needed based on this review. I would support merging after those steps.

These were controlled EclipseLink Core/H2 tests.

PR #379, opened 2019-03-05
[nobug] move jpars related code to appropriate module

lukasj:
This change is supposed to move all JPA-RS feature related code from various locations (mainly core, jpa) to appropriate module - dbws. As a consequence of this change, rest weaving is now by default turned off and turned on only within the context of JPA-RS application


lukasj:
@rfelcman feel free to take a look as well

lukasj:
@rfelcman failing tests should be fixed..

rfelcman:
LGTM

PR #267, opened 2018-10-17
Bug 540203 - Wrong characters in SQL script dbsetup_alltests.sql

rfelcman:
Bug 540203 - Wrong characters in SQL script dbsetup_alltests.sql

This fix rewrites string concatenation expressions with "||" operator to SQL function CONCAT(…).

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=540203

Signed-off-by: Radek Felcman <radek.felcman@oracle.com>

PR #384, opened 2019-03-13
remove mysql specific part from select

lukasj:
@rfelcman this is supposed to allow running dbws tests on Derby


rfelcman:
Yes I strongly agree. Something similar is PR #267 .

PR #401, opened 2019-03-25
fix test failures in spring tests

lukasj:
this fixes failures in spring tests in maven based build on MySQL


rfelcman:
This is related only with MySQL DB. On Derby and Oracle 12c it works without any fix.
Maybe change `@Table(name=...`  in other two classes 'Address.java', 'Truck.java' too, to keep these names consistent?
Otherwise LGTM.

lukasj:
@rfelcman mysql is apparently case sensitive while other DBs are not

PR #403, opened 2019-03-25
[2.7] Bug 533724: JAVAX.PERSISTENCE.LOCK.TIMEOUT IS SET AS SECONDS IN SQL

dazey3:
for #402

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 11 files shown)


dazey3:
I haven't messed around with EclipseLink's mapping files. Is there a wiki page (couldnt find one) that explains how to use https://www.eclipse.org/eclipselink/xsds/eclipselink_persistence_map_2.3.xsd? Is this just an alternative way of setting persistence properties?

lukasj:
Actually it is old (= pre-JPA) way of doing things, aka "native eclipselink". It is supposed to be super-seeded by JPA but there are areas for which there is no JPA equivalent for given functionality (ie compare org.eclipse.persistence.annotations.NamedStoredProcedureQuery with javax.persistence. NamedStoredProcedureQuery; or look at  org.eclipse.persistence.annotations.NamedStoredFunctionQuery, which has no equivalent on the JPA side).

For this kind of functionality, there is old user guide on the wiki: https://wiki.eclipse.org/Category:EclipseLink_User%27s_Guide

lukasj:
@dazey3 as for the code to change eventually, take a look at [EclipseLinkObjectPersistenceRuntimeXMLProject](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/sessions/factories/EclipseLinkObjectPersistenceRuntimeXMLProject.java) and its supertypes and [XMLProjectReader](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/sessions/factories/XMLProjectReader.java) (and maybe in #400 there you can find something useful in a commit which removes deprecated annotation field)

dazey3:
meh, if this is an old way of doing things, do we want to continue adding support? I looked through `ObjectPersistenceRuntimeXMLProject` and I can see where "queryTimeout" is being set, but this is for DescriptorQueryManager.queryTimeout, which isnt for this PR?

lukasj:
well, I didn't want to say we should continue adding support for old stuff. What I wanted to say is that there are few features going beyond what is required by JPA (like ie [NamedStoredFunctionQuery](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/annotations/NamedStoredFunctionQuery.java) ) which may benefit from the ability to set TIMEOUT_UNIT - if it makes sense (!)

Tests for this are almost all in foundation/eclipselink.core.test module where projects are defined in an xml files (xml can override and/or extend annotations - hence my suggestion), an example is ie [EmployeeProject](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/eclipselink.core.test/resource/EmployeeProject.xml)

PR #74, opened 2018-04-19
Bug 533724 - JAVAX.PERSISTENCE.LOCK.TIMEOUT IS SET AS SECONDS IN SQL

sureshbn:
Signed-off-by: Sureshkumar Balakrishnannair <sureshkumar.balakrishnannair@oracle.com>

PR #412, opened 2019-04-08
[master] Bug 533724: Set the default to MILLISECONDS

dazey3:
for #402

I was taking a look at the merged requests and I don't think it actually fixed the issue. The JPA spec clearly says that the value you provide to `javax.persistence.lock.timeout` & `javax.persistence.query.timeout` is in milliseconds, but #74 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=456067 elected to just set the default as seconds. I had set the default to SECONDS in 456067 because I didnt want to change behavior, but I feel like this was not correct. The value should be provided in MILLISECONDS and this is what 533724 was opened to address as well.

Also, I quickly tested QUERY_TIMEOUT and it didn't seem to work (the timeout was always set as 0)! I noticed that this property didnt unprepare the query. After fixing that, QUERY_TIMEOUT started working as expected. I added a couple tests for that property.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -1523,6 +1523,8 @@ public class DatabasePlatform extends DatasourcePlatform {
     /**
      * Platforms that support the WAIT option should override this method.
      * By default the wait timeout is ignored.
+     * 
+     *  @see DatabasePlatform#supportsWaitForUpdate()
      */
     public String getSelectForUpdateWaitString(Integer waitTimeout) {
         return getSelectForUpdateString();
@@ -2333,6 +2335,16 @@ public class DatabasePlatform extends DatasourcePlatform {
      * this method.
      */
     public boolean supportsVPD() {
+        return false;
+    }
+
+    /**
+     *  INTERNAL:
+     *  Indicates whether the platform supports timeouts on For Update
+     *  
+     *  @see DatabasePlatform#getSelectForUpdateWaitString(Integer waitTimeout)
+     */
+    public boolean supportsWaitForUpdate() {
         return false;
     }
 
--- a/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
+++ b/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2017 IBM Corporation. All rights reserved.
+ * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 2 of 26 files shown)


jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
```diff
@@ -84,10 +73,44 @@ public void testJDBCQueryTimeout() throws Exception {
...
+            //Convert the timeout set (MILLISECONDS) to what is expected by the JDBC layer (SECONDS)
+            double queryTimeoutSeconds = TestQueryHints.propertyTimeout / 1000d;
+            //if there was a remainder, it should round up
+            if(queryTimeoutSeconds % 1 > 0){
+                queryTimeoutSeconds += 1;
+            }
+
+            Assert.assertEquals((int)queryTimeoutSeconds, TestQueryHints.statementTimeout);
+
+            em.getTransaction().rollback();
```
lukasj: don't you want to rollback in finally? Probably not a big deal in this particular case but I can imagine someone copy & paste the code...

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
```diff
@@ -270,9 +333,13 @@ public PreparedStatementInvocationHandler(PreparedStatement stmt) {
         }
 
         public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            //Get the query timeout being set on the statement
+            //This value should be in seconds, since that is what the Statement expects
+            System.out.println(method.getName() + "INVOKED!!!!");
             if (method.getName().equals("setQueryTimeout") && proxy instanceof PreparedStatement) {
-                if(args.length > 0){
-                    TestQueryHints.setTimeout = (Integer)args[0];
+                System.out.println("INVOKED!!!!");
```
lukasj: this line should go away (or be sent to logger)

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
dazey3: Fair point, I'll move that as suggested

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
dazey3: Dang, I forgot to remove that... thanks for catching this

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -2336,6 +2336,16 @@ public boolean supportsVPD() {
         return false;
     }
 
+    /**
+     *  INTERNAL:
+     *  Indicates whether the platform supports timeouts on For Update
+     *  
+     *  @see {@link DatabasePlatform#getSelectForUpdateWaitString(Integer waitTimeout)}
```
lukasj: link tag should not be required here

dazey3: I added it to better document the link between supportsWaitForUpdate() <-> getSelectForUpdateWaitString(Integer waitTimeout). If a platform wants supportsWaitForUpdate() to return true, then they need to override getSelectForUpdateWaitString(Integer waitTimeout) as well to make use of the wait amount.

lukasj: I'm OK with having `see` there, it is good thing. The point is that `see` should already generate the link itself (unless escaped by `:`), so one more `link` looks redundant

lukasj: ... but I can be wrong here...

dazey3: And if you mean that `@link` is not needed, meaning `@see` is sufficient , I wanted it to be a clickable link

dazey3: Maybe the `@link` is needed if you are referencing a method in another class? I can't get THAT usecase to work... huh

lukasj: [docs](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIEDI) say that  `@see package.class#member label` should work; in this case just ` @see #getSelectForUpdateWaitString(Integer)` should be enough

dazey3:
Unfortunately, I suddenly thought that maybe I should go through and alter ALL the tests that are using these 2 properties as they might be expecting the SECONDS that they were using before this change... I am also noticing that a lot of these tests only run on Oracle, which I don't believe Travis runs against, so @lukasj you might need to look at this again to make sure your ok with these changes. I dont want to break Oracle tests

PR #415, opened 2019-04-10
[master] Bug 546312 - Add missing mapping for java.time classes to DatabasePlatform

Flugtiger:
This pr fixes bug 546312 (https://bugs.eclipse.org/bugs/show_bug.cgi?id=546312)


spligo:
This is still happening for LocalDate

PR #418, opened 2019-04-18
Bug 545779 - NullPointerException when MoxyJsonProvider reads a JSONArray of primitives. for example "[1]"

rfelcman:
Bug 545779 - NullPointerException when MoxyJsonProvider reads a JSONArray of primitives. for example "[1]"

Fix for this bug. This exception handing initial implementation for JSON related errors. 


rfelcman:
> [1] is valid json, so this is supposed to be consumed correctly by MOXy and not throw an exception

Yes that is true, but not for the Entity class specified in the bug. But this is just example. This bug is about "nicer"/more precision exception message. NPE is too generic.

PR #419, opened 2019-04-18
WIP: Bug# 545940 - LIKE query must escape '[' and ']' characters

Tomas-Kraus:
Implemented for MS SQL platform.
Covers both constant and parameter expressions in JPQL query.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/Expression.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/Expression.java
@@ -57,6 +57,7 @@ import org.eclipse.persistence.mappings.DatabaseMapping;
 import org.eclipse.persistence.queries.DatabaseQuery;
 import org.eclipse.persistence.queries.ReadQuery;
 import org.eclipse.persistence.queries.ReportQuery;
+import org.eclipse.persistence.queries.transform.LikePatternTransformation;
 
 /**
  * <p>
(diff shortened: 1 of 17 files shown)


Tomas-Kraus:
Added tests are passing on MS SQL database.

robertpatrick:
@Tomas-Kraus @arjantijms The SQL Server bracket issue is real: on current master, JPQL looking for `Be[ea]dril` returns `Beedril` instead. This change fixes that simple case, but I found several regressions and incomplete paths that need attention before we can merge it.

I compared current master with the PR’s production changes adapted to current source APIs, using SQL Server, Derby, Oracle, and H2.

|---|---|---|
| SQL Server: simple JPQL bracket pattern | Returns `Beedril` | Correctly returns `Be[ea]dril` |
| SQL Server: reuse a bound query with `[ab]`, then `[cd]` | Incorrect bracket matching | First execution is correct; second returns `c`, not `[cd]` |
| SQL Server: literal `[ab]` with a bound escape character | Incorrect bracket matching | Correct with `#`; changing to `!` returns no rows because the old escaped pattern remains |
| SQL Server: literal or bound `C:\temp` pattern | Returns `C:\temp` | Incorrectly returns `C:temp` |
| SQL Server: null pattern parameter | Completes with no rows | Throws `NullPointerException` |
| SQL Server: JPQL `CONCAT`, Criteria `CONCAT`, and Criteria column patterns | Queries execute | Query preparation fails with `UnsupportedOperationException` |
| Ordinary deserialized `SQLCall`, on all four databases | Executes successfully | Throws `NullPointerException`, even without `LIKE` |
| SQL Server: Criteria bracket literal without explicit `ESCAPE` | Incorrect bracket matching | Still incorrect |
 
The repeated-execution cases expose an incomplete fix, rather than previously correct bracket handling becoming broken. The transformation definitions are removed after use, while transformed constants can remain in the call. Separately, the new transient transformation field is null after deserialization, which explains the cross-database failure.
 
What we need:
 
| Action | Reason |
|---|---|
| Rework transformation lifetime and binding | Preserve reusable definitions, isolate execution/clone state, and handle each SQL occurrence independently. |
| Preserve serialization, nulls, and existing backslashes | These are working behaviors that the patch currently breaks. |
| Support legal expression patterns and cover Criteria consistently | Function/column patterns must continue to execute; bracket handling should not depend on the API used. |
| Extend the existing tests with the cases above; clean up fixtures in `finally` | The six added tests cover useful single executions but miss these failures. |
| Rebase and obtain clean existing CI | The current `pr-merge` status is an error: “This commit cannot be built.” |
 
I’d keep the issue open, but this needs more than a rebase and green build before sign-off. A revised implementation deserves another focused review.

Issue #420, opened 2019-04-18
Bug 421056: Aggregate functions in JPQL return 0 instead of null


dazey3:
Linking to [BZ #421056](https://bugs.eclipse.org/bugs/show_bug.cgi?id=421056)

dazey3:
A review comment while delivering to Liberty was that we need a property in order to turn this off on 2.6_was and 2.7. We wont add this ability to master, since that will be a major release and should be the default without needing to turn it off.

PR #421, opened 2019-04-18
[master] Bug 421056: Aggregate functions in JPQL return 0 instead of NULL

dazey3:
for #420 

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ReportItem.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ReportItem.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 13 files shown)


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/queries/ReportItem.java
```diff
@@ -130,21 +131,16 @@ public void initialize(ReportQuery query) throws QueryException {
...
                 DatabaseMapping mapping = this.attributeExpression.getLeafMapping(query, query.getDescriptor(), query.getSession());
-                if (mapping == null && this.attributeExpression.isFunctionExpression() && this.getResultType() == null){
-                    FunctionExpression expression = ((FunctionExpression)this.attributeExpression);
-                    if (expression.getOperator().equals(ExpressionOperator.maximum()) || expression.getOperator().equals(ExpressionOperator.minimum())){
-                        mapping = expression.getBaseExpression().getLeafMapping(query, query.getDescriptor(), query.getSession());
-                    }
-                }
-                if (mapping == null){
+                if (mapping == null) {
+                    //TODO: This feels like it should be set earlier during ReportItemBuild.visit. Should investigate.
```
jgrassel: Spotted a TODO, did you intend to keep that in here, or investigate the issue?

dazey3: I intended to leave that one there. I was running through this code section so much in my testing that I started to think that this other mapping check might be a poorly coded section that should have been fixed somewhere else (similar to the code right before that I altered). But it wasnt related to what I was currently fixing so I wanted to comment on it and not lose track of a possible issue there.

PR #423, opened 2019-04-24
Bug 521424: Add MOXY to logger categories

ZwS:
This would fix `IllegalArgumentException: Unknown logging category name.` on startup when `org.eclipse.persistence.logging.slf4j.SLF4JLogger` is selected as a logger.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/logging/LogCategory.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/logging/LogCategory.java
@@ -37,6 +37,7 @@ import org.eclipse.persistence.config.PersistenceUnitProperties;
  * <tr><td>&nbsp;</td><td>METADATA</td>       <td>&nbsp;</td><td>= "metadata"</td></tr>
  * <tr><td>&nbsp;</td><td>METAMODEL</td>      <td>&nbsp;</td><td>= "metamodel"</td></tr>
  * <tr><td>&nbsp;</td><td>MONITORING</td>     <td>&nbsp;</td><td>= "monitoring"</td></tr>
+ * <tr><td>&nbsp;</td><td>MOXY</td>           <td>&nbsp;</td><td>= "moxy"</td></tr>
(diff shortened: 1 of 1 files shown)


rfelcman:
LGTM

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/logging/LogCategory.java
```diff
@@ -61,15 +62,16 @@
...
-    PROCESSOR(  (byte)0x0D, SessionLog.PROCESSOR),
-    PROPAGATION((byte)0x0E, SessionLog.PROPAGATION),
-    PROPERTIES( (byte)0x0F, SessionLog.PROPERTIES),
-    QUERY(      (byte)0x10, SessionLog.QUERY),
-    SEQUENCING( (byte)0x11, SessionLog.SEQUENCING),
-    SERVER(     (byte)0x12, SessionLog.SERVER),
-    SQL(        (byte)0x13, SessionLog.SQL),
-    TRANSACTION((byte)0x14, SessionLog.TRANSACTION),
-    WEAVER(     (byte)0x15, SessionLog.WEAVER);
+    MOXY(       (byte)0x0D, SessionLog.MOXY),
```
lukasj: moxy as a _new_ category should be using some _new_ id (0x16) and not redefine existing one (0x0D) and shifting others. This is in a public package and can eventually break existing apps...

ZwS: Hello @lukasj . I've Added MOXY category in this way because I was guided by the following commit https://github.com/eclipse-ee4j/eclipselink/commit/e73f1d759dff45e032733599fdc89adf0848d6b2#diff-1bd332bafe844727dd0be92daa2fd1d2L59 . Is it incorrect as well?

lukasj: oh well, let's keep it your way then...

PR #424, opened 2019-04-24
Bug 501058 - Unmarshalling JSON with @XmlValue + @XmlAttribute(s) leaves value blank if value property is before attributes

rfelcman:
Bug 501058 - Unmarshalling JSON with @XmlValue + @XmlAttribute(s) leaves value blank if value property is before attributes

Bug Fix + unit test.


lukasj:
@rfelcman rebase this so it can be merged, please. Thanks!

Issue #426, opened 2019-04-29
Bug 546831: fix stored procedure memory leak


rdicroce:
EclipseLink has a severe memory leak when using stored procedures with long-lived EntityManagers.

StoredProcedureQueryImpl contains this comment:

// Add this query to the entity manager open queries list.
// The query will be closed in the following cases:
// Within a transaction:
//  - on commit
//  - on rollback
// Outside of a transaction:
//  - em close
// Other safeguards, we will close the query if/when
//  - we hit the end of the results.
//  - this query is garbage collected (finalize method)

With the current code, this comment is only about half correct.

- On commit/rollback: correct for a local transaction, but not for a JTA transaction.
- On EM close: always correct.
- Hit end of results: correct if the SP is executed using executeUpdate() or getSingleResult(), but not for execute() or getResultList().
- Query is GC'd: correct, but the query will never be GC'd because it is added to a WeakHashMap as both key AND value. Per the WeakHashMap Javadoc, keys are weakly referenced but values are strongly referenced.

Additionally, closing the query at end of results seems wrong, because closing the query prevents further calls to getOutputParameterValue(). And for getSingleResult(), the Javadoc says "Retrieve a single result from the next result set", so it sounds like it should be possible to invoke getSingleResult() multiple times on the same query.

Therefore, the proposed PR:
- Adds close on commit/rollback for JTA transactions.
- Removes close on end of results.
- Fixes use of WeakHashMap so the query can be GC'd.

I've tested the changes against my application and did not encounter any problems.

PR #427, opened 2019-04-29
[2.7] Bug 546831 - Fix memory leak caused by stored procedure queries

rdicroce:
Fixes stored procedure memory leak. See #426. If this looks okay to whoever reviews it, I will create another PR for master branch.


arjantijms:
@rdicroce it's been a very long time. What shall we do with this PR?

rdicroce:
The project where I needed this is now dormant and I've long since moved on to other work. I don't have time to update this PR. I don't even know if this is still a problem in newer versions.

arjantijms:
Thanks a lot for the swift reply!

I'll close the PR then.

PR #441, opened 2019-05-14
Fix erronous Javadoc update from bug 547023

jgrassel:
Signed-off-by: Joe Grassel <fyrewyld@gmail.com>

The bugfix for 547023 used '>' in javadoc in the new Oracle8 platform, which causes problems in javadoc generation. Pull request to fix this.


jgrassel:
It's got to do with the parsing to HTML, I suppose.  Hence the HTML escape codes for the '>' symbol.

PR #447, opened 2019-05-23
[Non bug] JPA JSE test fix

rfelcman:
This is fix for JPA JSE tests. Some tests can throw _NPE_ during `em.close(); em.getTransaction().rollback();` calls.

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
+++ b/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
@@ -16,236 +16,290 @@ package org.eclipse.persistence.jpa.test.query;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+
 import java.util.List;
 import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.TypedQuery;
+
 import org.eclipse.persistence.jpa.test.framework.DDLGen;
 import org.eclipse.persistence.jpa.test.framework.Emf;
 import org.eclipse.persistence.jpa.test.framework.EmfRunner;
 import org.eclipse.persistence.jpa.test.framework.Property;
 import org.eclipse.persistence.jpa.test.query.model.Dto01;
 import org.eclipse.persistence.jpa.test.query.model.EntityTbl01;
+import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
 @RunWith(EmfRunner.class)
 public class TestQueryCase {
-    @Emf(createTables = DDLGen.DROP_CREATE, classes = { EntityTbl01.class }, 
-            properties = { @Property(name="eclipselink.logging.level", value="FINE") })
+    @Emf(createTables = DDLGen.DROP_CREATE, classes = {EntityTbl01.class},
+            properties = {@Property(name = "eclipselink.logging.level", value = "FINE")})
     private EntityManagerFactory emf;
 
     private static boolean POPULATED = false;
 
     @Test
     public void testQueryCase1() {
-        if (emf == null)
-            return;
+        EntityManager em = null;
+        try {
+            if (emf == null)
+                return;
 
-        if(!POPULATED) 
-            populate();
+            if (!POPULATED)
+                populate();
 
-        EntityManager em = emf.createEntityManager();
+            em = emf.createEntityManager();
 
-        em.getTransaction().begin();
+            em.getTransaction().begin();
 
-        TypedQuery<EntityTbl01> query = em.createQuery(""
-                + "SELECT t FROM EntityTbl01 t "
+            TypedQuery<EntityTbl01> query = em.createQuery(""
+                    + "SELECT t FROM EntityTbl01 t "
                     + "WHERE t.itemString1 = ( "
-                        + "CASE t.itemInteger1 "
-                            + "WHEN 1000 THEN '047010' "
-                            + "WHEN 100 THEN '023010' "
-                            + "ELSE '033020' "
-                        + "END )", EntityTbl01.class);
+                    + "CASE t.itemInteger1 "
+                    + "WHEN 1000 THEN '047010' "
+                    + "WHEN 100 THEN '023010' "
+                    + "ELSE '033020' "
+                    + "END )", EntityTbl01.class);
 
-        List<EntityTbl01> dto01 = query.getResultList();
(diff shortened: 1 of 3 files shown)


jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
```diff
@@ -16,236 +17,290 @@
...
+            assertNotNull(dto01);
+            assertEquals(1, dto01.size());
+
+            assertEquals("A", dto01.get(0).getItemString1());
+            assertEquals("B", dto01.get(0).getItemString2());
+            assertEquals("C", dto01.get(0).getItemString3());
+            assertEquals("D", dto01.get(0).getItemString4());
+            assertEquals(new Integer(1), dto01.get(0).getItemInteger1());
+        } catch (Exception e) {
+            fail(e.getLocalizedMessage());
```
lukasj: I would probably avoid adding this explicit catch and just add finally - in case of an exception, it will bubble up. With  proposed change, full stacktrace is lost and one has to re-run updated the test to get it

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryHints.java
```diff
@@ -60,7 +60,7 @@
 
     private final static int propertyTimeout = 3099;
 
-    @Emf(name = "defaultEMF", classes = { Employee.class }, createTables = DDLGen.DROP_CREATE)
+    @Emf(name = "defaultEMF", classes = { Employee.class })
```
lukasj: I believe this was intentional to allow running single test... What about adding `createTables = DDLGen.DROP_CREATE` with calls to `populate` to tests, which you are trying to fix?

dazey3: Yes, I noticed this issue in JSE tests as well and I was slowly introducing the following structure:

``` java
try {
    ...
} finally {
    if (em.getTransaction().isActive()) {
        em.getTransaction().rollback();
    }
    if(em.isOpen()) {
        em.close();
     }
}
```

My thought was, if `em` == null, then an exception will have been thrown from `createEntityManager()` causing that to fail. That exception should be thrown before the `try` and fail the test. If it makes it into the finally block, then `em` != null since it passed the `createEntityManager()` invocation. Thoughts?

jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
dazey3: I agree. I've been trying to phase out the `catch` blocks from the JSE bucket since @jgrassel brought up the same issue. Also, I'm not sure I understand why an `em != null` check is needed at the bottom. If you use the structure I referenced in my other review comment, `em` cannot be null within the finally block.

lukasj: @dazey3 we've been discussing this with @rfelcman today and issues he is seeing happen if and only if TestQueryCase runs before others from the same package. When it runs last, no failure/issue is being observed - so the problem is related to the test execution order.

btw: when going through the test class itself, I also suggested to simplify it - it's JUnit 4 after all, so one can use Before/After to avoid copy/pasting stuff in tests (ie calls to populate, em creation etc)

rfelcman:
Just for info. It never happens during Ant build. I discovered this trouble when I locally transformed build into Maven (moved sources by `git mv` -> excecution order change). I received this error on staging Jenkins too.
Behind `java.lang.reflect.UndeclaredThrowableException` is trouble with `DROP TABLE oepjtbmEmployee` command. See `org.eclipse.persistence.jpa.test.basic.model.Employee`.
I simulated it against Derby and MySQL.

Tests execution order:
```[INFO] Running org.eclipse.persistence.jpa.test.concurrency.TestConcurrencyPersistence	OK
[INFO] Running org.eclipse.persistence.jpa.test.mapping.TestUnidirectionalOneToMany		OK
[INFO] Running org.eclipse.persistence.jpa.test.oraclefeatures.TestOracleLOBLocatorFeature	OK
[INFO] Running org.eclipse.persistence.jpa.test.query.TestQueryCase
[INFO] Running org.eclipse.persistence.jpa.test.query.TestQueryHints
```

This is error stack trace which I sometimes received:
```.....
[ERROR] Tests run: 4, Failures: 3, Errors: 1, Skipped: 0, Time elapsed: 0.111 s <<< FAILURE! - in org.eclipse.persistence.jpa.test.query.TestQueryHints
[ERROR] testJDBCQueryTimeout(org.eclipse.persistence.jpa.test.query.TestQueryHints)  Time elapsed: 0.074 s  <<< ERROR!
java.lang.NullPointerException
	at org.eclipse.persistence.jpa.test.query.TestQueryHints.testJDBCQueryTimeout(TestQueryHints.java:92)

[ERROR] testMultipleNamedQueryWithScrollableCursor(org.eclipse.persistence.jpa.test.query.TestQueryHints)  Time elapsed: 0 s  <<< FAILURE!
java.lang.AssertionError: java.lang.reflect.UndeclaredThrowableException
	at org.eclipse.persistence.jpa.test.query.TestQueryHints.testMultipleNamedQueryWithScrollableCursor(TestQueryHints.java:273)

[ERROR] testQueryTimeoutUnitMinutes(org.eclipse.persistence.jpa.test.query.TestQueryHints)  Time elapsed: 0.022 s  <<< FAILURE!
java.lang.AssertionError: java.lang.reflect.UndeclaredThrowableException
	at org.eclipse.persistence.jpa.test.query.TestQueryHints.testQueryTimeoutUnitMinutes(TestQueryHints.java:202)

[ERROR] testQueryTimeoutUnitSeconds(org.eclipse.persistence.jpa.test.query.TestQueryHints)  Time elapsed: 0.002 s  <<< FAILURE!
java.lang.AssertionError: java.lang.reflect.UndeclaredThrowableException
	at org.eclipse.persistence.jpa.test.query.TestQueryHints.testQueryTimeoutUnitSeconds(TestQueryHints.java:148)```

dazey3:
Interesting! I wouldnt have thought execution order would have impacted this. Perhaps it has something to do with the fact that TestQueryHints uses `org.eclipse.persistence.jpa.test.basic.model.Employee` (which is a common class used by other tests) and also uses `drop-and-create-tables`.

For whatever reason, the test seems to be failing on `emf.createEntityManager();`, which leaves `em`=null! Removing the catch (so that we can see the reason createEntityManager() is failing) and moving `emf.createEntityManager();` outside the `try {` block should clean up the error

PR #449, opened 2019-05-24
[2.7] Bug 421056: Add ability to disable MAX/MIN null result

dazey3:
#420 


dazey3:
@lukasj I actually hadnt planned on adding a test since #438 had included tests to make sure the fix works and this is just adding code to make sure I can turn off that fix... but that is a good suggestion and I added a test for this. Thanks!

PR #468, opened 2019-06-13
[master] Bug 548125: SDO Types generation is very slow for big schemas

### LINKED ISSUE #467: SDO Types generation is very slow for big schemas
[Link to BZ #548125](https://bugs.eclipse.org/bugs/show_bug.cgi?id=548125)

Happens in case of generation lots of SDO types, like a hundred thousand.

The generation is slow and the major bottleneck is the SDOType.setupInheritance method.
The profiling showed that the 2 methods consuming most of the time are:
1. NamespacesResolver#getNamespaces()
2. NamespacesResolver#put(String, String)

Call tree trace file is attached. - [Trace.zip](https://github.com/eclipse-ee4j/eclipselink/files/3282474/Trace.zip)

asanitckii:
Fixes #467 

Performance fixes:
1. Avoid creating a Vector object for each NamespaceResolver#getNamespaces() call
2. Use a simpler List structure instead LinkedHashMap for the interim buffer in the NamespaceResolver#put(String, String) method.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/oxm/NamespaceResolver.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/oxm/NamespaceResolver.java
@@ -14,17 +14,21 @@
 //     Oracle - initial API and implementation from Oracle TopLink
 package org.eclipse.persistence.internal.oxm;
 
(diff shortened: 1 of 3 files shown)


asanitckii:
signed ECA, but still cannot pass the check.

Will try to open another pull request.

PR #471, opened 2019-06-18
[master] Bug 391279: Fix usecase with existing FK mapping

dazey3:
for #233 

The original 391279 did not account for the usecase of an existing foreign key mapping on the owning side and instead expected the FK would be added by the @JoinColumn. This results in an exception similar to this:
```
Execute query InsertObjectQuery(model.CommentC@378deffb)
INSERT INTO COMMENTC (COMPLEXID_C_A, CONTENT, COMPLEXID_C_A, TYPE) VALUES (?, ?, ?, ?)
Internal Exception: java.sql.SQLSyntaxErrorException: Column name 'COMPLEXID_C_A' appears more than once times in the column list of an INSERT statement. 
```

The issue is due to the use of `ObjectBuilder.buildRow(AbstractRecord databaseRow, Object object, AbstractSession session, WriteType writeType)` in `OneToManyMapping`. This method calls `DatabaseMapping.writeFromObjectIntoRow(Object object, AbstractRecord row, AbstractSession session, WriteType writeType)`, which adds to the given row, causing duplicates. 


dazey3:
@lukasj thanks for the reviews!

PR #485, opened 2019-07-15
[2.6_WAS] Bug 547173 - EntityManager.unwrap(Connection.class) returns null

jgrassel:
Bug 547173 - EntityManager.unwrap(Connection.class) returns null

Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=547173 for 2.6_WAS branch

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerImpl.java
@@ -35,6 +35,8 @@
  *       - 393867: Named queries do not work when using EM level Table Per Tenant Multitenancy.
  *     02/27/2017-2.6_WAS Jody Grassel
  *       - 512255: Eclipselink JPA/Auditing capablity in EE Environment fails with JNDI name parameter type
+ *     05/10/2019-2.6_WAS Jody Grassel
+ *       - 547173: EntityManager.unwrap(Connection.class) returns null
  ******************************************************************************/
 package org.eclipse.persistence.internal.jpa;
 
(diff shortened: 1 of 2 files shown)


jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerImpl.java
```diff
@@ -2779,13 +2781,33 @@ public LockModeType getLockMode(Object entity) {
...
                 return (T) this.getSessionBroker();
             } else if (cls.equals(java.sql.Connection.class)) {
-                UnitOfWorkImpl unitOfWork = (UnitOfWorkImpl) this.getUnitOfWork();
-                if(unitOfWork.isInTransaction() || unitOfWork.getParent().isExclusiveIsolatedClientSession()) {
-                    return (T) unitOfWork.getAccessor().getConnection();
+                final UnitOfWorkImpl unitOfWork = (UnitOfWorkImpl) this.getUnitOfWork();
+                final Accessor accessor = unitOfWork.getAccessor();
+                if (unitOfWork.getParent().isExclusiveIsolatedClientSession()) {
+                    // If the ExclusiveIsolatedClientSession hasn't serviced a query prior to the unwrap, 
+                    // there will be no available Connection.
```
dazey3: Didn't we establish a connection during DatabaseSessionImpl.loginAndDetectDatasource()? What happens with that connection?

jgrassel: The idea with an Exclusive Isolated Client Session is to have a unique connection that is not shared between EntityManagers during use -- it is often used for enhanced security per connection (something you probably don't want to share with a different EntityManager that you intend to have a different security context with).  So the connection from DatabaseSessionImpl.loginAndDetectDatasource() is not really relevant.

PR #486, opened 2019-07-15
[Master] Bug 547173 - EntityManager.unwrap(Connection.class) returns null

jgrassel:
Bug 547173 - EntityManager.unwrap(Connection.class) returns null

Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=547173 for Master branch


jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/basic/TestEntityManagerInterface.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2019 IBM Corporation and/or its affiliates. All rights reserved.
```
lukasj: no need to have '1998' here, '2019' is enough

jgrassel: Noted, and fixed.  Thanks!

Issue #482, opened 2019-06-28
CacheId#compareTo violates equals() contract, resulting in wrong number of entites deleted when @Id is of type byte[] 


jpt-avaleo:
We have two entities, both that have UUID marked as `@Id` and with `BINARY(16)` as a MySQL representation of that. 

```
@Entity
@Table(name = "blog_entry")
@Convert(attributeName = "id", converter = UuidBinaryConverter.class)
public class BlogEntry {

	@Id
	@Column(name = "id", columnDefinition = "binary(16)")
	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "uuid_generator")
	private UUID id;

	@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.EAGER)
	@JoinColumn(name = "blog_entry_id", referencedColumnName = "id", nullable = false)
	private Collection<Comment> comments = new ArrayList<>();

}
```
and
```
@Entity
@Table(name = "comment")
@Convert(attributeName = "id", converter = UuidBinaryConverter.class)
public class Comment {

	@Id
	@Column(name = "id", columnDefinition = "binary(16)")
	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "uuid_generator")
	private UUID id;


}
```
Please note `orphanRemoval = true` on comments, but this issue is not related only to unidirectional relationship, it happens with bi-directional as well. It's not limited to orphanRemoval either, if we did Spring Data JPA repository for Comment entity with a method `deleteByBlogEntryId(UUID blogEntryId)`  
for bi-directional relationship it will yield same results.

The issue here is that when I clear comments collection that originally had 10 values with `comments.clear()` and commit the transaction then only one of the elements to be deleted is actually being deleted from the database. 

This is due to the fact that `org.eclipse.persistence.internal.sessions.CommitManager#sort` and especially `org.eclipse.persistence.internal.identitymaps.CacheId#compareTo` implementations being broken. 

`org.eclipse.persistence.internal.identitymaps.CacheId#compareTo` for `byte[]` type catches `java.lang.ClassCastException` when casting `byte[]` to `java.lang.Comparable` and returns `0` value for all the elements, regardless of its actual value. That in turn means that `org.eclipse.persistence.internal.sessions.CommitManager#sort` returns only one value,  even if you pass a list with multiple objects. 

This results in only one of the elements being actually deleted, regardles of how many we pass. 

[...]

PR #496, opened 2019-07-25
#482 - fix for CacheId comparision with BINARY arrays

mst-avaleo:
Hi

I would like to post a fix for bug #482. The fix is basically an implementation of `CacheId.compareTo()` which supports char, byte or Object arrays in primary keys. The implementation was moved to `CacheIdComparator` for better encapsulation. 

Besides array support, I've changed current implementation for objects which don't implement `Comparable`. Current approach returns `0` when an exception is thrown during comparison. This create inconsistency with `equals()` method. For example: `new Object().compareTo(new Object())` returns `0`,
but new `Object().equals(new Object())` returns `false`. I've switch to compareTo based on hashCodes
which should give equals() consistency and stable results.

Signed-off-by: Michal Stochmialek <uek@kmd.dk>

### DIFF
--- a/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/junit/identitymaps/CacheIdTest.java
+++ b/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/junit/identitymaps/CacheIdTest.java
@@ -1,0 +1,208 @@
+package org.eclipse.persistence.testing.tests.junit.identitymaps;
+
+import static org.junit.Assert.*;
+
+import org.eclipse.persistence.internal.identitymaps.CacheId;
+import org.junit.Test;
+
+public class CacheIdTest {
+
+	@Test
+	public void compareToForByteArrays() {
+		equal("many primaryKeys - equal",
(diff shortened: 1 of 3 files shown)


mst-avaleo:
@dazey3 Any idea how to restart CI build on this PR?

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/junit/identitymaps/CacheIdTest.java
```diff
@@ -0,0 +1,208 @@
+package org.eclipse.persistence.testing.tests.junit.identitymaps;
```
lukasj: this file needs license header..

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/identitymaps/CacheIdComparator.java
```diff
@@ -0,0 +1,101 @@
+package org.eclipse.persistence.internal.identitymaps;
```
lukasj: license

PR #500, opened 2019-08-01
[2.7] Bug#531910 Add support for JGroups ver. 4.x in session coordination

mshaposhnik:
Add support for JGroups ver. 4.x in session coordination mechanism.
That will allow to use newest `KUBE_PING` protocols and resolve some issues with Openshift 4 based replication (see https://github.com/eclipse/che/issues/13607 for example)

Original issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=531910

### DIFF
--- a/buildsystem/compdeps/pom.xml
+++ b/buildsystem/compdeps/pom.xml
@@ -91,7 +91,7 @@
         <org.slf4j.version>1.7.25</org.slf4j.version>
         <wsdl4j.version>1.6.2</wsdl4j.version>
         <!-- CQ #7143 (Workswith) JGroups 3.2.8+ -->
-        <org.jgroups.version>3.6.17.Final</org.jgroups.version>
+        <org.jgroups.version>[3.6.17.Final,4.1.0.Final]</org.jgroups.version>
     </properties>
 
     <build>
--- a/buildsystem/compdeps/pom.xml
+++ b/buildsystem/compdeps/pom.xml
@@ -91,7 +91,7 @@
         <org.slf4j.version>1.7.25</org.slf4j.version>
         <wsdl4j.version>1.6.2</wsdl4j.version>
         <!-- CQ #7143 (Workswith) JGroups 3.2.8+ -->
-        <org.jgroups.version>[4.1.0.Final]</org.jgroups.version>
+        <org.jgroups.version>4.1.0.Final</org.jgroups.version>
     </properties>
 
     <build>
--- a/buildsystem/compdeps/pom.xml
+++ b/buildsystem/compdeps/pom.xml
@@ -91,7 +91,7 @@
         <org.slf4j.version>1.7.25</org.slf4j.version>
         <wsdl4j.version>1.6.2</wsdl4j.version>
         <!-- CQ #7143 (Workswith) JGroups 3.2.8+ -->
-        <org.jgroups.version>[3.6.17.Final,4.1.0.Final]</org.jgroups.version>
+        <org.jgroups.version>[4.1.0.Final]</org.jgroups.version>
     </properties>
 
     <build>
--- a/antbuild.properties
+++ b/antbuild.properties
@@ -14,7 +14,7 @@
 
 # The directory that holds the oracle-specific jar files.
 # You can either put the jars in this directory, or specify your own directory.
-junit.lib= /home/Makis/.m2/repository/junit/junit/4.12/junit-4.12.jar
+junit.lib=../extension.lib.external/junit.jar
 tools.lib=${env.JAVA_HOME}/lib/tools.jar
 
 jacocoant.lib=../extension.lib.external/jacocoant.jar
--- a/antbuild.properties
+++ b/antbuild.properties
@@ -14,7 +14,7 @@
 
 # The directory that holds the oracle-specific jar files.
 # You can either put the jars in this directory, or specify your own directory.
-junit.lib=../extension.lib.external/junit.jar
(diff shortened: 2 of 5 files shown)


slemeur:
@rfelcman Is there someone who can review the PR?

rfelcman:
Usually @lukasj but he is really busy in these days.
But could You please rebase this PR/branch against latest commit in 2.7. I fixed there Travis-CI configuration file. Travis-CI check now is functional.

dazey3:
@slemeur Make sure you update your 2.7 branch to include https://github.com/eclipse-ee4j/eclipselink/pull/505, which was merged only a couple days ago. The travis build will not be able to run tests without these changes. Thanks!

mshaposhnik:
@dazey3  @rfelcman  build is passing now, thanks for pointing out

dazey3:
@mshaposhnik I appreciate you looking into this and I can appreciate that this is a blocker for the Eclipse Che release. 

Have you tested with this change to validate it fixes the issue in Che? I'm personally not familiar with jgroups, nor EclipseLink's testing in this space. I see that EclipseLink has some tests in `foundation\eclipselink.core.test\src\org\eclipse\persistence\testing\tests\distributedservers\rcm\jgroups`. Though, I doubt we are running these tests currently -_-'

Also, should `foundation\org.eclipse.persistence.extension\pom.xml` be updated to use the latest version?

mshaposhnik:
@dazey3 yes, we have tested that it resolves our issue.  And yes, seems that pom.xml must be updated as well, thanks for pointing.

buildsystem/compdeps/pom.xml
```diff
@@ -91,7 +91,7 @@
         <org.slf4j.version>1.7.25</org.slf4j.version>
         <wsdl4j.version>1.6.2</wsdl4j.version>
         <!-- CQ #7143 (Workswith) JGroups 3.2.8+ -->
-        <org.jgroups.version>3.6.17.Final</org.jgroups.version>
+        <org.jgroups.version>[3.6.17.Final,4.1.0.Final]</org.jgroups.version>
```
dazey3: I think just go with `<org.jgroups.version>4.1.0.Final</org.jgroups.version>` instead of keeping around the old version.

lukasj:
dependency from `foundation\org.eclipse.persistence.extension\pom.xml` should be removed completely (it's actually a bug that it is there) and `uploadToNexus.xml` should be updated to add the dependency to the appropriate pom during deployment (.... I hope we move away from these unintuitive steps soon....)

buildsystem/compdeps/pom.xml
lukasj: keep '4.1.0.Final' only, remove 3.6.17.Final

mshaposhnik: Ok

lukasj:
btw: it is value of `extension.dependencies` property in `uploadToNexus.xml` which should be updated to define optional dependency on this jgroups artifact

mshaposhnik:
Sorry guys it seems i will require further assintace in removing the jgroups dependency from pom and moving into  `uploadToNexus.xml`.
I understoot that i have to rewrite something like this into it:
`<property name="extension.dependencies"   value="${deps.head}${dep.core}${dep. jgroups}${deps.foot}"/>`
and then define the `dep.jgroups` itself, writing something like:
`<property name="dep.jgroups"           value="${dep.grp}org.eclipse.persistence${dep.art}org.jgroups${dep.ver}${maven.version}${dep.type}jar${dep.scope}compile${dep.opt}false${dep.foot}`

Am i correct?

lukasj:
Yes, something like that should be correct. Possible way to verify the change is described at https://github.com/eclipse-ee4j/eclipselink/wiki/Building-EclipseLink#antbuild

mshaposhnik:
@dazey3 @lukasj please review if i descriped dependency in a right way. On my local, command
```
ant -f uploadToNexus.xml -DsnapshotId=local -DsnapshotURL=file:~/m2/repository  -Drelease.version=2.7.5 -Dbuild.type=SNAPSHOT -Dgit.hash=local  -Dmavenant.dir=/home/Makis -Dversion.string=2.7.5.qualifier
```  
worked successfully.

dazey3:
@mshaposhnik LGTM, but @lukasj is more knowledgeable in these build files than me

mshaposhnik:
@lukasj can you review pls? Is there anything more we should do ? Thanks.

lukasj:
@mshaposhnik did you try to create some dummy project with dependency on the artifact installed to the local repo to verify the pom is correct and artifact properly resolvable by maven? Also note that dependency on jgroups should be 'optional=true'

mshaposhnik:
@lukasj Ok let me check

mshaposhnik:
@lukasj unfortunately seems test if failed = maven complains about corrupted `maven-metadata-local.xml`. This file doen't contain an  `<snapshotVersions>` enumeration. Maybe i'm missing something in the installation command? (see [above](https://github.com/eclipse-ee4j/eclipselink/pull/500#issuecomment-528235394) )?

lukasj:
@mshaposhnik I usually blame Tycho for this... Can you confirm that the pom in the local repo is valid xml at least?

mshaposhnik:
@lukasj Yes pom looks good and jgroups dependency is correctly gerenated ( butin 1 row :) )

lukasj:
@mshaposhnik can you port this also to master, please?

PR #502, opened 2019-08-02
Update travis config

lukasj:
@dazey3 this change looks promising...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ branches:
 addons:
   apt:
     packages:
+      - oracle-java8-installer
       - libmysql-java
       - mysql-server
       - mysql-client
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,7 @@ before_script:
   - echo 'db.user=root' >> $HOME/build.properties
   - echo 'db.pwd=root' >> $HOME/build.properties
   - echo 'db.platform=org.eclipse.persistence.platform.database.MySQLPlatform' >> $HOME/build.properties
-  - if [[ $JAVA_HOME == *openjdk* ]]; then openssl x509 -in <(openssl s_client -connect oss.sonatype.org:443 -prexit 2>/dev/null) -out ~/example.crt; sudo keytool -importcert -file ~/example.crt -alias example -keystore $JAVA_HOME/lib/security/cacerts -storepass changeit -noprompt; fi
+  - if [[ $JAVA_HOME == *openjdk10 ]]; then openssl x509 -in <(openssl s_client -connect oss.sonatype.org:443 -prexit 2>/dev/null) -out ~/example.crt; sudo keytool -importcert -file ~/example.crt -alias example -keystore $JAVA_HOME/lib/security/cacerts -storepass changeit -noprompt; fi
 
 script:
   - cat $HOME/build.properties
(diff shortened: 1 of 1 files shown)


dazey3:
Maybe changing the distribution to "xenial" changes the support clause: https://forums.mysql.com/read.php?11,677237,677268#msg-677268

.travis.yml
```diff
@@ -10,17 +10,19 @@
 # SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 #
 
+dist: xenial
```
dazey3: Since "trusty" seems to have gone out of support (https://forums.mysql.com/read.php?11,677237,677268#msg-677268), is there an issue just updating to "bionic" (18.04) instead of "xenial" (16.04)?

.travis.yml
```diff
@@ -10,17 +10,19 @@
...
 language: java
 sudo: required
 
+services:
+    - mysql
```
dazey3: I think I read that only "services" is needed if "addons" isnt being used. Not sure about "sources"... but maybe "addons" can be removed completely if we are using "services" for mysql now.

.travis.yml
lukasj: xenial is current default, we used to be on trusty on which mysql became unsupported...

.travis.yml
lukasj: if it is not listed in services then the db server won't be started by default (from xenial), that's how I understand the doc.

I don't think libmysql-java is there by default, hence we likely need addons/apt/packages at least

.travis.yml
lukasj: ...in the other words - I don't have time to play with it right now, will be away on Mon and then from Fri. This is mostly quick way to unblock people while I'm gone....

lukasj:
hmmm, we need to figure out, why tests crashed on openjdk8 :-/

dazey3:
@lukasj What exception are the tests failing with?

lukasj:
```
3940    [junit] Running org.eclipse.persistence.testing.tests.junit.failover.ConnectionPoolFailoverTest
3941    [junit] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0 sec
3942    [junit] Test org.eclipse.persistence.testing.tests.junit.failover.ConnectionPoolFailoverTest FAILED (crashed)
```

dazey3:
@lukasj "crashed" seems really vague... can you access the machine and grab the real exception?

lukasj:
no way to do that. The only way is to install openjdk8 and run tests locally. These are quick to run as they are pure junit ones but before doing so, let me try it one more time....

.travis.yml
dazey3: I see. "trusty" wasnt declared anywhere as the distribution, right? Why wasnt "xenial" used then as the new default (seems to have been set as the default from May, 2019.

.travis.yml
dazey3: Ok, then maybe "mysql-server" in "services" is just redundant

dazey3: I was reading this doc to see what needs to be changed for xenial: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

dazey3:
I downloaded "OpenJDK8U-jdk_x64_windows_openj9_8u212b03_openj9-0.14.0", but I cant seem to get any failures. Not sure what is going on with OpenJDK8

rfelcman:
Hello, I created similar PR #504 with some fix of Ant build files.

lukasj:
not relevant anymore

PR #514, opened 2019-09-03
JPA WDF Tests fix

rfelcman:
This prevents collision with other tests executed against same DB table.
There was collision at the DB row level.
Entities with same id "...new Department(1..." was used in multiple tests above same DB table.
This error happened at new ci-staging.eclipse.org build server.
For example there: https://ci-staging.eclipse.org/eclipselink/view/New%20Jobs/job/eclipselink-nightly-master/4/console

### DIFF
--- a/jpa/eclipselink.jpa.wdf.test/src/org/eclipse/persistence/testing/tests/wdf/jpa1/entitymanager/TestClear.java
+++ b/jpa/eclipselink.jpa.wdf.test/src/org/eclipse/persistence/testing/tests/wdf/jpa1/entitymanager/TestClear.java
@@ -30,8 +30,8 @@ public class TestClear extends JPA1Base {
         JPAEnvironment env = getEnvironment();
         EntityManager em = env.getEntityManager();
         try {
-            Department dep1 = new Department(1, "one");
-            Department dep2 = new Department(2, "two");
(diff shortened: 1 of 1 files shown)


dazey3:
Is this a case where the tests ran in a different order than previously? Should these tests clean up what they inserted instead?

rfelcman:
Yes this is related with different execution order. As I know, it never happened before.
In the new Jenkins job https://ci-staging.eclipse.org/eclipselink/view/New%20Jobs/job/eclipselink-nightly-master/4/console
following tests was executed almost at the same time.
org.eclipse.persistence.testing.tests.wdf.jpa1.entitymanager.TestPersist
org.eclipse.persistence.testing.tests.wdf.jpa1.entitymanager.TestClear

About drop/create strategy: There is `<property name="eclipselink.ddl-generation" value="drop-and-create-tables"/>` switch in _persistence.xml_ but tests are executed in parallel way. Switch into sequential execution strategy  will increase execution time from ~ 2 minutes to ~ 15 minutes.

This is from my point quickest solution.

Conflict with `TestClear.testClearNoTransaction()`: I don't think so. There is different `@Id` (primary key) 1 x 11 or 1111 x 11 after update.

Issue #518, opened 2019-09-06
Bug 518774: SAP HANA DB 2.x is detected as DB2Platform instead of HANAPlatform


dazey3:
Linking to [BZ #518774](https://bugs.eclipse.org/bugs/show_bug.cgi?id=518774)

marcopar:
Hi, may i ask what exactly is branch 2.6_WAS and why this is not merged to 2.6?
Looks like 2.6.x releases don't have this fix while 2.7.x have it.

PR #520, opened 2019-09-06
[master] Bug 518774: HANAPlatform is not detected; improving platform detection

dazey3:
for #518

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/ddl/TestDBPlatformHelper.java
+++ b/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/ddl/TestDBPlatformHelper.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2015 IBM Corporation. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/ddl/TestDBPlatformHelper.java
```diff
@@ -51,19 +51,47 @@ public void test() {
...
+    }
+
+    @Test
+    public void testOracle() {
+        //Returned from jcc driver (DRDA)
+        Assert.assertEquals("org.eclipse.persistence.platform.database.oracle.Oracle9Platform", getPlatformClass("Oracle", "9", "Oracle Database 9c"));
+        Assert.assertEquals("org.eclipse.persistence.platform.database.oracle.Oracle10Platform", getPlatformClass("Oracle", "10", "Oracle Database 10c"));
+        Assert.assertEquals("org.eclipse.persistence.platform.database.oracle.Oracle11Platform", getPlatformClass("Oracle", "11", "Oracle Database 11c"));
+        Assert.assertEquals("org.eclipse.persistence.platform.database.oracle.Oracle12Platform", getPlatformClass("Oracle", "12", "Oracle Database 12c"));
+        Assert.assertEquals("org.eclipse.persistence.platform.database.oracle.Oracle18Platform", getPlatformClass("Oracle", "18", "Oracle Database 18c"));
```
lukasj: 'org.eclipse.persistence.platform.database.oracle.OraclePlatform' is valid too, maybe wanna add it?

dazey3: Sure! What are the conditions to have OraclePlatform be set as the database platform? The only test case that comes to mind for me is when the version  < 9?

lukasj: the case used to be the fallback when either the exact version is not recognized or if 'oracle' jar is missing

dazey3: I'm not sure how to test for the latter, so I just added a test for the former. Does this work for you?

lukasj: Yes, sth like this is good enough.

BTW: NOT doing force-pushes makes it easier for others to find the exact diff between code which was already reviewed and new revision. Imagine the diff has 1k lines, one line needs to be changed and one has to verify that only that one particular line was changed...; one can use 'squash and merge' in github UI to make one commit out of the whole PR...

dazey3: I suppose that would make review easier, thank you for the suggestion.

Issue #521, opened 2019-09-11
Bug 550951: Parameter binding for java.time types results in SQL syntax exception


dazey3:
Linking to [BZ #550951](https://bugs.eclipse.org/bugs/show_bug.cgi?id=550951)

PR #522, opened 2019-09-11
[2.7] Bug 550951: Parameter binding for java.time types results in SQL syntax exception

dazey3:
for #521


dazey3:
I opened #540 to account for @rfelcman review comment

Issue #512, opened 2019-08-30
Bug 550542: Add support for Oracle named parameters


dazey3:
Linking to [BZ #550542](https://bugs.eclipse.org/bugs/show_bug.cgi?id=550542)

PR #530, opened 2019-09-13
[master] Bug 550542: Add support for Oracle named parameters

dazey3:
for #512


rfelcman:
Hi Will,

I checked this fix against Oracle12c but I think, that is not fixing this bug correctly.
I didn't run complete LRG tests just Core LRG tests with following results.
Before this fix I received in `ScrollableCursorForwardOnlyResultSetTest` similar bug as You described in 550542:
```
...
Caused by: Error : 6550, Position : 53, Sql = BEGIN Read_All_Employees(RESULT_CURSOR=>RESULT_CURSOR=>:0); END;, OriginalSql = BEGIN Read_All_Employees(RESULT_CURSOR=>?); END;, Error Msg = ORA-06550: line 1, column 54:
PLS-00103: Encountered the symbol ">" when expecting one of the following:
...

```
but after fix I am getting:
```
...
Internal Exception: java.sql.SQLException: operation not allowed: Ordinal binding and Named binding cannot be combined! Error Code: 17090 Call: BEGIN Read_All_Employees(?); END; bind => [1 parameter bound] Query: DirectReadQuery()

Local Exception Stack:
Exception [EclipseLink-4002] (Eclipse Persistence Services - 3.0.0.qualifier): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLException: operation not allowed: Ordinal binding and Named binding cannot be combined!
Error Code: 17090
Call: BEGIN Read_All_Employees(?); END;
bind => [1 parameter bound]
...

```
I think, that fix leads in some cases into SQL (anonymous PL/SQL block) with some binding variables passed positionally (by index) and some passed there by name in the same SQL query. But It's not allowed in the Oracle DB (jdbc).
See attached test report. Not all errors there are related with this PR
[test-report.zip](https://github.com/eclipse-ee4j/eclipselink/files/3615700/test-report.zip)

dazey3:
@rfelcman Wow, thanks for running these tests!

Taking a look at the failure you hit, it seems there may be in issue with leaving "org.eclipse.persistence.queries.StoredProcedureCall.cursorOrdinalPositions" hanging around... 
That implementation seems to have been added to work around the named parameter support, but I should just yank it out to find all the places named -> ordinal conversion is still occurring.

dazey3:
Ok! I removed the `.getObject(index)` invocations I could find in favor of a class dependent implementation. 

`DatabasePlatform.setParameterValueInDatabaseCall(..., String name, ...)` is called from StoredProcedureCalls when there are named arguments. Now, the same named arguments should also be used when getting the OUT objects from the java.sql.Statement (instead of just using an index). 

I ran this against TestRunModel (on Oracle) and I saw the previous failures pass.

dazey3:
@rfelcman mind taking another look at this? I have been running the `test-jpa22` target with Oracle and while I do still see some failures, I think they are unrelated to this PR. 

Also, seeing these failures in here for Oracle, we should look into why these are failing and how to fix. Its difficult to run on my end as the test-jpa22 target takes about 3 hours to run (remote Oracle connection)

rfelcman:
I thin, that there are three bugs related with Oracle DB
1. This one related with stored procedure and named parameters
2. SQL MAX, MIN function test. There is problem with different type BigDecimal vs expected Integer. [in progress].
3. ....ORA-00904: "Smith": invalid identifier Error Code: 904 Call... [solved by #548 , #550 ]

I'll check code from there and I'll let You know.

rfelcman:
I checked latest version of this PR. See attached file.
There are three test results:
jpa_jse - OK
core_tests (OraclePlatform, Oracle12Platform) - check core_tests_Oracle12cPlatform.html it is better. There are still some Oracle named parameter errors. Search for `ORA-06550` text.
As I said. I'am working on MAX, MIN bugs.

[test-report.zip](https://github.com/eclipse-ee4j/eclipselink/files/3652987/test-report.zip)

dazey3:
@rfelcman Hmm, interesting. Based on your test results, it looks like `StoredProcedureTest_Inout_Out_In` is failing, but only on core_tests_Oracle12cPlatform? I dont see those test failures in core_tests_OraclePlatform. 

The failures I see in core_tests_OraclePlatform seem to be MAX/MIN failure and "Unsupported feature: createArrayOf". These don't sound like they are related to this PR. NOTE: There used to be a NPE failure as well, was that fixed locally for you?

The failures I see in core_tests_Oracle12Platform seem to be the same (MAX/MIN failures) as core_tests_OraclePlatform, but with 4 `StoredProcedureTest_Inout_Out_In` failures as well. 

Am I understanding this correctly? It seems odd to me that changing the Platform had this effect of creating additional failures. I am looking through `StoredProcedureTest_Inout_Out_In` test now to see if I can figure out why this is failing

lukasj:
Important is for tests to pass with the concrete platform from the ‘oracle’ specific module and common OraclePlatform from the core module. I’m not worried that much about 12cPlatform defined in the core

rfelcman:
@dazey3 MIN/MAX was fixed see #557 #558 .
In test results above I used as `db.platform` `org.eclipse.persistence.platform.database.OraclePlatform` and `org.eclipse.persistence.platform.database.oracle.Oracle12Platform`.
`org.eclipse.persistence.platform.database.oracle.Oracle12Platform` is used in our internal Jenkins.
I'am searching for a platform and test differences too and "createArrayOf" error.
I think, that I have to compare OraclePlatform classes.

rfelcman:
Just for info there are core and JPA tests against 'org.eclipse.persistence.platform.database.oracle.Oracle12Platform' . I manually applied there code from PRs #563 #560.

[test-core-jpa-reportsOracle12Platform.zip](https://github.com/eclipse-ee4j/eclipselink/files/3663510/test-core-jpa-reportsOracle12Platform.zip)

dazey3:
ooo! I was working on this! I think there is still a test failure

lukasj:
oops :-( should have put a note here... never mind, what Radek fixed was merged, this one as well, so from current master you both should have the same bits. The issue you see can be fixed in the new PR and it may be easier for others to see it, if there is a need to..

rfelcman:
@dazey3 I tested today (30-SEP-2019) core and JPA (master branch) and there are latest reports (for booth platforms `org.eclipse.persistence.platform.database.oracle.OraclePlatform`, `org.eclipse.persistence.platform.database.oracle.Oracle12Platform`) 
In JPA tests with `org.eclipse.persistence.platform.database.oracle.OraclePlatform` just ignore TestCase OracleJPATestSuite. Are You still working on StoredProcedureNullInOutTest test?
[test-core-jpa-reports.zip](https://github.com/eclipse-ee4j/eclipselink/files/3671938/test-core-jpa-reports.zip)

dazey3:
@rfelcman 

## core-tests-Oracle12Platform (5 failures)
TestRunModel:
1. **StoredProcedureTest_Inout_Out_In proc storedProc named don'tBind**
        This is due to a bug where if binding is disabled for the Call, then the Oracle driver assumes everything is ordinal.  Therefore setting named parameters is incorrect. Im looking into this situation.
2. **StoredProcedureNullInOutTest bind = false**
      Same as 1
3. **StoredProcedureTest_Inout_Out_In proc storedProc named don'tBind**
      Same as 1
4. **StoredProcedureNullInOutTest bind = false**
      Same as 1
5. **ProjectXMLStoredFunctionCallTest**
      `ClassCastException: java.lang.Long cannot be cast to java.lang.Integer`
      I thought there was a PR that fixed this. Seems like test bug

## core-tests-OraclePlatform (3 failures)
  TestRunModel:
1. **CustomSQLTestModel**
      `java.sql.SQLFeatureNotSupportedException: Unsupported feature: createArrayOf`
      I have no clue about this failure because I havent looked into it at all. Can we track down when this started failing?
2. **InsuranceObjectRelationalTestModel**
      Same as 1
3. **MappingWMIntegrationStoredProcedureTestModel**
      Same as 1

## jpa-tests-Oracle12Platform (4 failures)
  AllCMP3TestRunModel:
1. **testCoalesceJPQLQueryWithNullParameterValue**
  AllCMP3TestRunModel2
1. **testPESSIMISTIC_LockWithDefaultTimeOutUnit**
      This failure looks related to #412, but I don't think 412 is wrong. The timeout is set to 2000 and the default unit is MILLISECONDS which should be comparable to the original test of 2 and default unit of SECONDS. A LockTimeoutException means that the lock was set for longer than expected. Does this mean that something is wrong with the implementation for #412? That it thinks the default timeunit is still SECONDS, making it a timeout of 2000 SECONDS??? I'm not sure about this since my testing never hit that issue. Then again ... `testPESSIMISTIC_LockWithSecondsTimeOutUnit` does seem to be passing, so setting the default to SECONDS manually seems to work... hmm, I'm not sure whats going on here
2. **testEmpRecordInOut**
      This seems to be another issue with my NamedParameter support for Oracle -_-' . I need to look into this one
3. **testEmpRecordInOut**
      Same as 2

## jpa-tests-OraclePlatform (4 failures?)
  Same as jpa-tests-Oracle12Platform.html

rfelcman:
@dazey3 
Thank You for a nice summary.
### core-tests-Oracle12Platform
5. **ProjectXMLStoredFunctionCallTest**
    It's fixed by #567 , #569
### core-tests-OraclePlatform
1.-3. `Unsupported feature: createArrayOf`
    I'am working on this. This message comes from Oracle JDBC driver
    `oracle.jdbc.driver.PhysicalConnection#createArrayOf`
    I think, that I'll remove these tests from OraclePlatform or maybe just selected methods.

dazey3:
@rfelcman 
So, I took a bit of a look into the `Unsupported feature: createArrayOf` issue. Something I found curious is that lack of support for `java.sql.Connection.createArrayOf(...)` with Oracle is a known thing and EclipseLink works around the issue with the overriding behavior in `org.eclipse.persistence.platform.database.oracle.Oracle8Platform.createArray(...)`.
The exception stack has this:
```
Caused by: java.sql.SQLFeatureNotSupportedException: Unsupported feature: createArrayOf
at oracle.jdbc.driver.PhysicalConnection.createArrayOf(PhysicalConnection.java:6641)
at org.eclipse.persistence.internal.databaseaccess.DatabasePlatform.createArray(DatabasePlatform.java:3392)
at org.eclipse.persistence.internal.databaseaccess.DatabasePlatform.createArray(DatabasePlatform.java:3373)
```
but shouldn't the stack contain `org.eclipse.persistence.platform.database.oracle.Oracle8Platform` since it overrides `createArray`??

PR #532, opened 2019-09-14
Skip duplicate processing for @ManyToMany mappings

### LINKED ISSUE #531: @ManyToMany: Primary key depends on alphabetical order of entity class names
I use Eclipselink 2.7.4 and try to create a simple bidirectional many-to-many mapping:
```
@Entity
@Table(name = "Users")
public class User {

    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Integer id;

    private String username;

    @ManyToMany
    private Set<Role> roles = new HashSet<>();
}

@Entity
public class Role {

    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Integer id;

    private String rolename;

    @ManyToMany(mappedBy = "roles")
    private Set<User> users = new HashSet<>();
}
```
With theese entities Eclipselink generates this relation table:

    CREATE TABLE Users_ROLE (roles_ID INTEGER NOT NULL, users_ID INTEGER NOT NULL,
        PRIMARY KEY (roles_ID, users_ID));

Note columns order in the primary key (and it seems to me this order is incorrect).

If I change the class name `User` to `AppUser` (I change only the class name, its table name remain the same - "Users"), then Eclipselink generates:

    CREATE TABLE Users_ROLE (users_ID INTEGER NOT NULL, roles_ID INTEGER NOT NULL,
        PRIMARY KEY (users_ID, roles_ID));

What I see is the primary key columns order depends on the alphabetical order entity class names instead of the mappedBy attribute:
- in first case `Role`, `User`;
- in second case `AppUser`, `Role`.

And also a primary key doesn't change if I move `mappedBy` attribute from `Role` to `User`.

oshatrk:
Fixes #531 . This PR is a breaking change as with it the order of columns in primary keys becomes dependent on a relation owning and presense of the `mappedBy` attribute.

Signed-off-by: Oleg Chtchoukine [olegpsu@mail.ru](mailto:olegpsu@mail.ru)

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/tools/schemaframework/DefaultTableGenerator.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/tools/schemaframework/DefaultTableGenerator.java
@@ -413,6 +413,14 @@ public class DefaultTableGenerator {
      * Build relation table definitions for all many-to-many relationships in a EclipseLink descriptor.
      */
     protected void buildRelationTableDefinition(ForeignReferenceMapping mapping, RelationTableMechanism relationTableMechanism, DatabaseField listOrderField, ContainerPolicy cp) {
+        if(mapping.isReadOnly()) {
+            // Skip non-owning mappings to prevent duplicate processing.
+            // See how mappings with a "mappedBy" attribute are marked
+            // as read-only in ManyToManyAccessor.process():
+            // [Eclipselink 2.7.4]: https://github.com/eclipse-ee4j/eclipselink/blob/ad5b7c6b2ada222561e59a25cd3f09ee6eef7dfb/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/accessors/mappings/ManyToManyAccessor.java#L162
+            return;
+        }
+
         //first create relation table
         TableDefinition table = getTableDefFromDBTable(relationTableMechanism.getRelationTable());
 


robertpatrick:
@oshatrk @arjantijms This PR needs changes and to be rebased.  It is not safe to merge as written.

The underlying issue is real: generated join-table primary-key column order depends on which relationship mapping is processed first. The patch makes the owning side determine that order, but skips too much.

| Locally tested case | Current master | With PR change |
| ---- | ---- | ---- |
| Inverse mapping processed first | Target column first | Owner column first—intended improvement |
| Owning mapping processed first | Owner column first | Unchanged |
| Independently read-only, unidirectional relationship | Join table generated; relationship loads | Join table omitted; loading fails with “table not found” |

The cause is the new if (mapping.isReadOnly()) return;. Read-only does not necessarily mean inverse or redundant. A relationship can be independently read-only and still need its table generated.  [Changed code, line 415](https://github.com/eclipse-ee4j/eclipselink/blob/342760fdb2b6379eb2bda2c860c3d9937a79efdc/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/tools/schemaframework/DefaultTableGenerator.java#L415).

What we need is:

| Action | Why |
| ---- | ---- |
| Narrow the duplicate-processing check | Preserve tables and metadata required by independently read-only relationships |
| Add focused schema-generation tests | Cover both processing orders and the read-only case; the PR currently adds no tests |
| Document the intentional PK-order change | This changes generated DDL; maintainers should explicitly accept that compatibility change |
| Rebase and obtain passing existing CI | Conflicts with master, 1,020 commits behind, and no current-head checks or statuses reported |

Conditional verdict: worthwhile after correction, targeted tests and clean CI. The ordering issue itself is not a demonstrated Persistence-spec violation.

The regression was reproduced through JPA/H2 using the PR’s functional patch on current master.

Issue #437, opened 2019-05-10
Bug 547173 - EntityManager.unwrap(Connection.class) returns null


jgrassel:
Eclipse bugtracker https://bugs.eclipse.org/bugs/show_bug.cgi?id=547173

An EntityManager, using an EntityManagerFactory with the property "eclipselink.jdbc.exclusive-connection.mode" set to "Always" returns a null value if the EntityManager hasn't been exercised in any operations that interact with the database (a simple em.find() will work around this issue.)  With a EntityManager that is configured to retain an exclusive connection it seems plausible that this should return a Connection.

arunvc:
Connection is still null in 2.7.7

Connection is returned after 
EntityManager.getTransaction().begin() or a dummy EntityManager.find()

PR #535, opened 2019-09-16
[master] Bug 547173: em.unwrap returns NULL in EE application

dazey3:
#437


dazey3:
When testing the previous change in an EE application server, we discovered that `em.unwrap(Connection.class)` was returning a null value. This was due to the Accessor being stale. The original change obtained the accessor AFTER beginning an early transaction. When we put this logic back to the original, it fixed the issue.

PR #536, opened 2019-09-17
[master] Bug 551815: When an entity with an aggregate with all fields set to null is changed to null for the aggregate field then Eclipselink executes a bogus no-op SQL update statement

eperez:


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateMapping.java
```diff
@@ -354,6 +354,39 @@ public ChangeRecord compareForChange(Object clone, Object backup, ObjectChangeSe
...
+
+        return fields;
+    }
+
+    public static boolean checkNull(final Object o) {
+        if (o == null) {
+            return true;
+        }
+        final Set<java.lang.reflect.Field> fields = getAllFields(new HashSet<java.lang.reflect.Field>(), o.getClass());
+        for (final java.lang.reflect.Field f : fields) {
```
dazey3: If `Object o` is not null and contains no fields, then return `true`?

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/models/aggregate/SimpleAggregate.java
```diff
@@ -0,0 +1,13 @@
+package org.eclipse.persistence.testing.models.aggregate;
```
dazey3: No headers on any files

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/models/aggregate/SimpleAggregateProject.java
```diff
@@ -0,0 +1,84 @@
+package org.eclipse.persistence.testing.models.aggregate;
```
dazey3: No header

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/mappings/AggregateMapping.java
eperez: Yes, that is intended as there should not be any difference. Do you see any problem in that behavior?

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/models/aggregate/SimpleAggregate.java
eperez: I added headers.

foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/models/aggregate/SimpleAggregateProject.java
eperez: I added headers.

lukasj:
How does this deal with fields marked '@Transient'? What if app uses annotations on methods and not on fields? ie eclipselink can trigger getter which in turn returns something different from what is stored in the field..

lukasj (changes requested):
instead of direct reflection, appropriate accessors and possibly objectbuilder should be used

eperez:
> instead of direct reflection, appropriate accessors and possibly objectbuilder should be used

@lukasj I have changed the code to use ObjectBuilder instead. Is it OK now?

rfelcman:
Could You please fix year in the copyright header:
1. In case of new files (just creation year) e.g.:
`Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.`
2. In case of modified files (creation year, modification year) e.g.:
`Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.`

same case for #606

robertpatrick:
@eperez @arjantijms The unnecessary-update case is real, but this version also skips required writes when an embeddable uses a converter. I wouldn’t merge it as written.

I tested a converter that stores Java null as the text <NULL>. An empty embeddable therefore has different database content from a null embeddable, even though all its Java attributes are null.

| ---- | ---- | ---- |
| Plain all-null embeddable ↔ null | Redundant UPDATE | UPDATE correctly avoided |
| Converter-backed empty embeddable → null | Clears stored value to SQL NULL | Incorrectly retains <NULL> |
| Null → converter-backed empty embeddable | Writes <NULL> | Incorrectly leaves SQL NULL |
| Populated-value changes and other owner-field updates | Correct | Unchanged |

These results reproduced with deferred change tracking on both H2 and Derby. A separate woven attribute-tracking comparison was unchanged—the optimization does not cover that path.

What we need:

| Action | Reason |
| ---- | ---- |
| Restrict the optimization to mappings with equivalent persisted values | All-null Java attributes do not necessarily mean SQL NULL; preserve conversion and null-validation behavior |
| Add named regression cases to the existing aggregate LRG suite | The new test model is not selected by normal CI. Cover both directions, with and without the converter, explicit starting-state assertions and actual database/SQL checks |
|  Rebase and obtain passing existing CI | The branch is 981 commits behind. |

A narrower version could be useful, but preserving required writes needs to come first.

These local tests were controlled JPA/H2 and Derby tests.

PR #537, opened 2019-09-17
[2.6_WAS] Update Eclipselink Bundled ASM from 7.0 to 7.1

jgrassel:
To support JDK 13, we need to update the level of ASM from 7.0 to 7.1.


jgrassel:
Piggybacking on existing CQ for ASM 7.1: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20831

jgrassel:
Waiting for the IP review first.  Then going to need to figure out how to resolve the compdeps -- actually cannot build as it complains about compdeps.

lukasj:
the piggyback should have been done against CQ 19168 (if you took sources from maven - preferred) and not CQ 19240 (means you took from eclipse orbit, ie eclipse update site which is not the case here)

jgrassel:
> the piggyback should have been done against CQ 19168 (if you took sources from maven - preferred) and not CQ 19240 (means you took from eclipse orbit, ie eclipse update site which is not the case here)

I cloned the source from ASM's gitlab repo https://gitlab.ow2.org/asm/asm

lukasj:
CQ 19168 is better then

jgrassel:
> CQ 19168 is better then

And sounds like perhaps I should grab the source from that resource, rather than the gitlab clone I sourced instead.

lukasj:
or you could clone from the tag mentioned in CQ 19521

jgrassel:
Still, I would like to actually be able to build locally, so how do I resolve the compdeps headache?

jgrassel:
Looking at the jar in CQ 19521, it's missing a large swath of packages  (asm.commons, asm.tree, asm.util).  I don't think we can use that one, unless it has companions that include those other packages.

lukasj:
and not cause headache to me? :-D well...

assuming you are not behind proxy (which makes it more complicated), something like this should work:

```ant -f antbuild.xml clean build -Declipse.install.dir=/path/to/eclipse/ide```

in case of troubles, try to look for some hints at https://wiki.eclipse.org/EclipseLink/Building#Updating_dependencies and https://wiki.eclipse.org/EclipseLink/Building/UpdatingDependencies

**Comment by jgrassel** [comment]
That helps, I've been going off the instructions in the README in `/plugins/org.eclipse.persistence.asm` . Thanks!

**Comment by jgrassel** [comment]
In the meantime, I put a comment on my CQ request to change the piggyback to CQ 19168 as that contains the full set of asm packages.

**Comment by jgrassel** [comment]
I've also downloaded the source jars from CQ 19168 and will rebuild this update using them instead of the gitlab clone.

**Comment by dazey3** [comment]
@lukasj Thanks for the assistance on this! Will we want to port this change to 2.6 when we are done? or is 2.6 (JPA 2.1) stabilized?

**Comment by jgrassel** [comment]
Forced update on the branch, clean room update using ASM 7.1 sources from CQ 19168

**Comment by jgrassel** [comment]
Looks like the IP Review went through
```
------- Comment #1 From Joe Grassel CLA 2019-09-17 16:57:52 [reply] -------
I'd like to change this to piggy back off of CQ 19168
------- Comment #2 From Genie Eclipse CLA 2019-09-17 17:00:04 [reply] -------
Thank you for your request.  Please wait for IP Team assistance.
------- Comment #3 From Shawn Kilpatrick CLA 2019-09-19 13:40:15 [reply] -------
Approve reuse of license certified content.
```
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20831

**Comment by jgrassel** [comment]
I think it's ready, pretty much waiting for the review, then merge it into the 2.6_WAS branch and rebuild compdeps.

**Comment by jgrassel** [comment]
@lukasj How does it look, good to merge and build compdeps?  Once 2.6_WAS is clean (making sure I've got the process right before addressing the other branches), I'll get to updating 2.7 and master.

**Comment by jgrassel** [comment]
@lukasj  So I've already built the ASM binary and added it to this pull request, is that okay, or there a reason for the two-commit approach?

**Comment by lukasj** [comment]
I was hoping git hash of the commit from which the binary was built gets it somehow to the manifest but apparently that's not the case here...
So only make sure that the binary was built using 1.7 for 2.6 (if you'll be porting this there) and using 1.8 for 2.7/master

**Comment by jgrassel** [comment]
Ah, I believe I was using JDK 8 to build, so if it needs to be JDK 7 then I will have to do a rebuild of the ECL-ASM bundles.

**Comment by lukasj** [comment]
2.6 has the min required jdk 7. Don’t think we want to change this

**Comment by jgrassel** [comment]
Checking the major version of the compiled class in the generated asm jar, it comes up with 51, which corresponds to Java 7, so I think we're good there.

```
javap -verbose -cp org.eclipse.persistence.asm_7.1.0.v201909181055.jar org.eclipse.persistence.internal.libraries.asm.AnnotationVisitor 
Classfile jar:file:/Users/jgrassel/Dev/EE4J/eclipselink/plugins/org.eclipse.persistence.asm_7.1.0.v201909181055.jar!/org/eclipse/persistence/internal/libraries/asm/AnnotationVisitor.class
  Last modified Sep 18, 2019; size 2124 bytes
  MD5 checksum 3611ee93b54ea509bd82f4895ee8c1c7
  Compiled from "AnnotationVisitor.java"
public abstract class org.eclipse.persistence.internal.libraries.asm.AnnotationVisitor
  minor version: 0
  major version: 51
  flags: ACC_PUBLIC, ACC_SUPER, ACC_ABSTRACT
```

lukasj:
Good. Make sure to use 8 in 2.7 as java 13 drops support for 7 I think

jgrassel:
Thankfully, that doesn't seem to be the case.  Page 73 on the Java Virtual Machine Spec for JDK 13 (http://cr.openjdk.java.net/~iris/se/13/latestSpec/java-se-13-jvms-diffs.pdf) states the following:

Table 4.1-A. class file format major versions

| Java SE | Corresponding major version | Supported major versions |
| ------- | --------------------------- | ------------------------ |
| 1.0.2 | 45 | 45 |
| 1.1 | 45 | 45 |
| 1.2 | 46 | 45 .. 46 |
| 1.3 | 47 | 45 .. 47 |
| 1.4 | 48 | 45 .. 48 |
| 5.0 | 49 | 45 .. 49 |
| 6 | 50 | 45 .. 50 |
| 7 | 51 | 45 .. 51 |
| 8 | 52 | 45 .. 52 |
| 9 | 53 | 45 .. 53 |
| 10 | 54 | 45 .. 54 |
| 11 | 55 | 45 .. 55 |
| 12 | 56 | 45 .. 56 |
| 13 | 57 | 45 .. 57 |

It looks like JDK 13 will support java class files compiled at JDK v1.1 or greater. Whew!

jgrassel:
Merged, kicked off compdep build https://ci.eclipse.org/eclipselink/job/eclipselink-update-compdeps/52/

PR #580, opened 2019-10-09
[master] JPA test fix for SELECT...FOR UPDATE WAIT xx

rfelcman:
JPA test fix for SELECT...FOR UPDATE WAIT xx

This is fix for JPA with `SELECT...FOR UPDATE WAIT xx` queries.
It's executed on OraclePlatform only.
Some description for SQL SELECT FOR UPDATE WAIT xx option (from Oracle DB doc):

The NOWAIT and WAIT clauses let you tell the database how to proceed if
the SELECT statement attempts to lock a row that is locked by another user.

Specify WAIT to instruct the database to wait integer seconds for the row to become available and then return control to you.

Before this fix `QueryHints.PESSIMISTIC_LOCK_TIMEOUT` was applied to first query, but
to test this feature correctly it should be applied to the second query in test.
This is not instruction how long DB keeps lock, but how long query will
wait for a lock release if there are some locked rows in the selection.
To simulate real environment more more accurately some timeout and lock release
instructions for a first query are executed in parallel (in thread) with second query.

`testPESSIMISTIC_LockWithDefaultTimeOutUnit` expects, that selected rows are
available before timeout (FOR UPDATE WAIT xx)

`testPESSIMISTIC_LockWithSecondsTimeOutUnit` expects `javax.persistence.LockTimeoutException`
Selected rows will not be available before timeout (FOR UPDATE WAIT xx)

https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6

### DIFF
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/jpql/AdvancedQueryTestSuite.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/jpql/AdvancedQueryTestSuite.java
@@ -1507,6 +1507,10 @@ public class AdvancedQueryTestSuite extends JUnitTestCase {
         // Lock timeout only supported on Oracle.
         if (! isOnServer() && session.getPlatform().supportsWaitForUpdate()) {
             EntityManager em = createEntityManager();
+            // sleep for 2 seconds (2000 milliseconds)
+            // this timeout value must be smaller, than PESSIMISTIC_LOCK_TIMEOUT property for the second query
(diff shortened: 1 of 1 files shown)


dazey3:
@rfelcman ah, I see my confusion and it seems so obvious now that the `WAIT` clause is how long the QUERY will wait. I think what was confusing me was the `WAIT` clause on the first query as well, which seems unnecessary. haha, oops. Thanks for clearing this up!

dazey3:
@rfelcman We should port this to 2.7 as well. Thanks!

PR #582, opened 2019-10-09
[2.6_WAS] Bug 550542: Addressing additional test failures

dazey3:
for #512

I included a port of `Bug 469182` in this PR as it was needed for the Coalesce fix and was never ported to 2.6_WAS originally


dazey3:
@rfelcman Thanks for the review!

PR #597, opened 2019-10-23
Fix #595: annotations with ElementType.TYPE_USE are not handled properly

### LINKED ISSUE #595: Using Jersey 2.29.1 for validation causes EclipseLink to generate erroneous code for @NotNull on byte arrays
With code

```java
import javax.persistence.Column;
import javax.persistence.Entity;
import javax.persistence.Id;
import javax.persistence.Lob;
import javax.validation.constraints.NotNull;

@Entity
public class MyEntity {
  @Id
  private Long id;

  @NotNull
  @Lob
  @Column(nullable = false, length = 5120, columnDefinition = "BLOB")
  private byte[] bytes;
}
```
This causes the following generate code
```java
import (@javax.validation.constraints.NotNull :: byte);
import javax.annotation.Generated;
import javax.persistence.metamodel.SingularAttribute;
import javax.persistence.metamodel.StaticMetamodel;

@Generated(value="EclipseLink-2.7.5.v20191016-rNA", date="2019-10-22T09:50:03")
@StaticMetamodel(MyEntity.class)
public class MyEntity_ { 

    public static volatile SingularAttribute<MyEntity, NotNull :: byte)[]> bytes;
    public static volatile SingularAttribute<MyEntity, Long> id;

}
```
Which results in compiler errors.
I've isolated the issue in a Maven project here: 
https://github.com/verymahler/sampljejpa

lukasj:
Fixes #595

### DIFF
--- a/jpa/org.eclipse.persistence.jpa.modelgen/src/main/java/org/eclipse/persistence/internal/jpa/modelgen/visitors/AnnotationValueVisitor.java
+++ b/jpa/org.eclipse.persistence.jpa.modelgen/src/main/java/org/eclipse/persistence/internal/jpa/modelgen/visitors/AnnotationValueVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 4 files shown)


rfelcman:
It seems, that this fix leads into NPE during _org.eclipse.persistence.jpa.test_ module build.
```
...
[INFO] --- maven-processor-plugin:3.3.3:process-test (generate-canonical-meta-model-classes) @ org.eclipse.persistence.jpa.test ---
[INFO] diagnostic: Note: Creating static metadata factory ...
[ERROR] diagnostic: error: java.lang.NullPointerException
[ERROR] error on execute: use -X to have details 
...
```
Maybe something wrong at:
`at org.eclipse.persistence.internal.jpa.modelgen.visitors.ElementVisitor.visitVariable (ElementVisitor.java:297)`


Stack trace (for command `mvn install -pl :org.eclipse.persistence.jpa.test -X` with more details is attached.
[pr597_ErrorStackTrace.txt](https://github.com/eclipse-ee4j/eclipselink/files/3766324/pr597_ErrorStackTrace.txt)

lukasj:
yep, I noticed that already :-(

Problem is that annotations with ElementType.TYPE_USE are not recognized by javax.lang.model API. That also mean that even cases like `void x(@NotNull String param)` are not handled properly.

lukasj:
@rfelcman this should be better, I hope

PR #605, opened 2019-10-30
[master] Dependencies and Maven plugins upgrade

rfelcman:
[master] Dependencies and Maven plugins upgrade

In some cases this upgrade leads into some pom.xml or Java sources change:

1. mongodb driver from 3.2.0 to 3.11.1
   There is different error message in case of wrong authentication
2. maven-bundle-plugin from 4.1.0 to 4.2.1
   From >=4.2.0 includes into "Import-Package" packages from "Class.forName(...)" as mandatory regardless if maven dependency is specified or not.
3. SLF4J update from 1.7.25 to 2.0.0-alpha1
   Without this update tests in org.eclipse.persistence.extension will fail


lukasj:
dependency tree shows that hamcrest 1.3 is still in use (transitive dependency of junit 4.12) - is that intentional? For this particular change I'd prefer it comes with junit update to 4.13-something in a separate PR

rfelcman:
> dependency tree shows that hamcrest 1.3 is still in use (transitive dependency of junit 4.12) - is that intentional? For this particular change I'd prefer it comes with junit update to 4.13-something in a separate PR

OK I returned hamcrest version from 2.2 to 1.3.

Issue #610, opened 2019-11-08
ASM 7.2 Support


jgrassel:
ASM 7.2 was released recently, providing support for JDK 14.  Even though we just recently updated to ASM 7.1, updating to this sooner than later may be desirable.

IP Review for using ASM 7.2 in Eclipselink has been approved with CQ 21030 ( https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21030)

PR #611, opened 2019-11-08
[2.7] Update to ASM 7.2 (#610)

jgrassel:
for #610 

Signed-off-by: Joe Grassel <fyrewyld@gmail.com>

Update ASM to 7.2 for JDK 14 support.


jgrassel:
Needed to change:
https://github.com/eclipse-ee4j/eclipselink/blob/a225926f972b06125249db5fea4136a3e8401f94/plugins/org.eclipse.persistence.asm/src/org/eclipse/persistence/internal/libraries/asm/tree/Util.java#L43-L46

to:

```
  static <T> List<T> add(final List<T> list, final T element) {
    List<T> newList = list == null ? new ArrayList<T>(1) : list;
    newList.add(element);
    return newList;
  }
```

As I got a compiler failure, but only when trying to produce the release asm build.

Issue #618, opened 2019-11-22
Bug: Oracle positional parameters with ParameterMode.REF_CURSOR out parameter


rfelcman:
EclipseLink throw exception during a call of Oracle DB stored procedure with signature like this:
`PROCEDURE GetEmpIdFNameLNameFromRS(in_param IN INTEGER, p_recordset OUT SYS_REFCURSOR) AS`
It happens only if input parameter is passed via position e.g.
`spq.setParameter(1, 1);`
If input parameter is passed via name like:
`spq.setParameter("in_param", 1);`
it works OK.
I think, that this issue is related with #512 .
There is attached retest application to simulate this error. DB object creation is disabled there, but there is attached _init.sql_ script to create required DB objects.
Main test code is in the following methods
`com.oracle.jpa.bugtest.TestBug#bugTestNamedParam` - OK

`com.oracle.jpa.bugtest.TestBug#bugTestPositionParam` - throws Exception

[jpa-bug-EJB_FailedToUseStoredprocedureQuery.tar.gz](https://github.com/eclipse-ee4j/eclipselink/files/3880273/jpa-bug-EJB_FailedToUseStoredprocedureQuery.tar.gz)

As I checked this kind of call (positional parameters + `ParameterMode.REF_CURSOR`) declared via entity annotation `@NamedStoredProcedureQuery` is not tested in EclipseLink now.

Error stack:
```
javax.persistence.PersistenceException: java.lang.IllegalArgumentException: Invalid output parameter name : 2. operation not allowed: Ordinal binding and Named binding cannot be combined!

	at org.eclipse.persistence.internal.jpa.StoredProcedureQueryImpl.getResultList(StoredProcedureQueryImpl.java:589)
	at com.oracle.jpa.bugtest.TestBug.bugTestPositionParam(TestBug.java:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:515)
[...]

PR #619, opened 2019-11-27
[master] Bug #618: Oracle positional parameters with ParameterMode.REF_CURSOR out parameter

rfelcman:
[master] Bug #618: Oracle positional parameters with ParameterMode.REF_CURSOR out parameter

This is fix for bug #618 with additional unit tests. There is small refactoring
(renaming) in test method names and NamedStoredProcedureQuery annotation names.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
@@ -785,8 +785,7 @@ public abstract class DatabaseCall extends DatasourceCall {
                 // outParameter contains all the info for registerOutputParameter call.
                 OutputParameterForCallableStatement outParameter = new OutputParameterForCallableStatement(outField, session, isCursor);
                 this.parameters.set(i, outParameter);
-                // nothing to do during translate method
-                this.parameterTypes.set(i, LITERAL);
+                this.parameterTypes.set(i, parameterType);
             }
         }
         if (this.returnsResultSet == null) {
@@ -1181,9 +1180,6 @@ public abstract class DatabaseCall extends DatasourceCall {
                         parametersValues.add(value);
                     }
                 } else if (parameterType == LITERAL) {
-                    if ((parameter != null) && (parameter instanceof OutputParameterForCallableStatement)) {
-                        ((OutputParameterForCallableStatement) parameter).getOutputField().setIndex(index);
-                    }
                     parametersValues.add(parameter);
                 } else if (parameterType == IN) {
                     Object value = getValueForInParameter(parameter, translationRow, modifyRow, session, true);
@@ -1194,6 +1190,11 @@ public abstract class DatabaseCall extends DatasourceCall {
                 } else if (parameterType == INOUT) {
                     Object value = getValueForInOutParameter(parameter, translationRow, modifyRow, session);
                     parametersValues.add(value);
+                } else if (parameterType == OUT || parameterType == OUT_CURSOR) {
+                    if (parameter != null) {
+                        ((OutputParameterForCallableStatement) parameter).getOutputField().setIndex(index);
+                    }
+                    parametersValues.add(parameter);
                 }
             }
             setParameters(parametersValues);
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
@@ -1181,6 +1181,9 @@ public abstract class DatabaseCall extends DatasourceCall {
                         parametersValues.add(value);
                     }
                 } else if (parameterType == LITERAL) {
+                    if ((parameter != null) && (parameter instanceof OutputParameterForCallableStatement)) {
+                        ((OutputParameterForCallableStatement) parameter).getOutputField().setIndex(index);
+                    }
                     parametersValues.add(parameter);
                 } else if (parameterType == IN) {
                     Object value = getValueForInParameter(parameter, translationRow, modifyRow, session, true);
--- a/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa21/advanced/StoredProcedureQueryTestSuite.java
+++ b/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa21/advanced/StoredProcedureQueryTestSuite.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -83,7 +83,8 @@ public class StoredProcedureQueryTestSuite extends JUnitTestCase {
         suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureParameterAPI"));
         suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQuerySysCursor_Named"));
         suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQuerySysCursor_Positional"));
-        suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQuerySysCursor2"));
+        suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQuerySysCursor2_Named"));
+        suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQuerySysCursor2_Positional"));
         suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQueryExceptionWrapping1"));
         suite.addTest(new StoredProcedureQueryTestSuite("testStoredProcedureQueryExceptionWrapping2"));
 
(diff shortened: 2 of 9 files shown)


jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java
```diff
@@ -527,6 +527,15 @@ public Object getOutputParameterValue(String parameterName) {
         return null;
     }
```
dazey3: /*
* Returns false if any query parameter is a named parameter, otherwise returns true.
*/

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
```diff
@@ -1181,6 +1181,9 @@ public void translate(AbstractRecord translationRow, AbstractRecord modifyRow, A
                         parametersValues.add(value);
                     }
                 } else if (parameterType == LITERAL) {
```
dazey3: Interesting. So the `parameterType` is DatasourceCall.LITERAL and not DatasourceCall.OUT_CURSOR?

jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa22/advanced/StoredProcedureQueryTestSuite.java
```diff
@@ -1056,6 +1057,33 @@ public void testStoredProcedureQuerySysCursor2() {
         }
     }
 
+    /**
+     * Tests a StoredProcedureQuery using a system cursor. Also tests
+     * getParameters call AFTER query execution. Parameters are passed via position.
+     */
+    public void testStoredProcedureQuerySysCursor2_Positional() {
+        if (supportsStoredProcedures() && getPlatform().isOracle() ) {
```
dazey3: is `supportsStoredProcedures()` not enough? What about running this on `getPlatform().isMySQL()`?

dazey3:
Did `org.eclipse.persistence.testing.tests.jpa22.advanced.StoredProcedureQueryTestSuite.testStoredProcedureQuerySysCursor_Named()` & `org.eclipse.persistence.testing.tests.jpa22.advanced.StoredProcedureQueryTestSuite.testStoredProcedureQuerySysCursor_Positional` fail before this change?

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java
dazey3: hmm. I'm not sure I like this implementation. It feels sorta hackish, right? Checking all of the parameters to make a decision on one? Idk.

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java
```diff
@@ -557,7 +566,12 @@ public List getResultList() {
                 if (hasMoreResults()) {
                     if (isOutputCursorResultSet) {
                         // Return result set list for the current outputCursorIndex.
-                        List results = (List) getOutputParameterValue(getCall().getOutputCursors().get(outputCursorIndex++).getName());
```
dazey3: Interesting. This appears to have always worked (or it was a bug before my fix) because I didn't make this change.

dazey3: This same fix may be needed in getSingleResult() as well.

jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa21/advanced/StoredProcedureQueryTestSuite.java
```diff
@@ -1063,6 +1064,33 @@ public void testStoredProcedureQuerySysCursor2() {
         }
     }
 
+    /**
+     * Tests a StoredProcedureQuery using a system cursor. Also tests
+     * getParameters call AFTER query execution. Parameters are passed via position.
+     */
+    public void testStoredProcedureQuerySysCursor2_Positional() {
```
dazey3: This test doesn't test "getParameters", it tests "getResultList()" as opposed to the other tests (`testStoredProcedureQuerySysCursor_Named`/`testStoredProcedureQuerySysCursor_Positional`) that test `getOutputParameterValue(String name)`/`getOutputParameterValue(int index)`

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
dazey3: Interesting. So the parameterType is DatasourceCall.LITERAL and not DatasourceCall.OUT_CURSOR? That seems wrong, right?

jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa21/advanced/StoredProcedureQueryTestSuite.java
dazey3: This test doesn't test "getParameters", it tests "getResultList()" as opposed to the other tests (testStoredProcedureQuerySysCursor_Named/testStoredProcedureQuerySysCursor_Positional) that test getOutputParameterValue(String name)/getOutputParameterValue(int index)

testStoredProcedureQuerySysCursor2_Named above this isnt testing getResultList(), it's testing getOutputParameterValue() as well. Maybe we need to make that distinction in the test names so we can cover everything.

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java
dazey3: This same fix may be needed in StoredProcedureQueryImpl.getSingleResult() as well.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java
dazey3: Eww, `org.eclipse.persistence.internal.databaseaccess.DatabaseCall.prepareInternalParameters(AbstractSession session)` looks like it sets the type as `DatasourceCall.LITERAL`:
https://github.com/eclipse-ee4j/eclipselink/blob/a842a6e2d9a64c1d26210bdf2179e17942035fd7/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseCall.java#L788-L789

I don't particularly like the logic of "well, LITERAL doesn't do anything, so lets just set it as a LITERAL type". That thinking becomes void if we start adding logic into LITERAL! Now, we are setting the index during translate if the type is OUT_CURSOR... Maybe we can fix this up?

dazey3: Gross... This might extend to other places too... 
https://github.com/eclipse-ee4j/eclipselink/blob/a842a6e2d9a64c1d26210bdf2179e17942035fd7/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java#L438-L447
It might not be entirely safe to undo this gross logic... thoughts?

rfelcman:
I have to solve some JPA errors (MySQL only).

jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa22/advanced/StoredProcedureQueryTestSuite.java
```diff
@@ -1011,7 +1012,7 @@ public void testStoredProcedureQuerySysCursor_Positional() {
      * Tests a StoredProcedureQuery using a system cursor. Also tests
      * getParameters call AFTER query execution.
      */
-    public void testStoredProcedureQuerySysCursor2() {
+    public void testStoredProcedureQuerySysCursor_ResultList_Named() {
```
dazey3: Everything looks good in the tests so far except I think this test needs updated to use `getResultList()` like `testStoredProcedureQuerySysCursor_ResultList_Positional` does. This version still uses the old `getOutputParameterValue("")` operation that is tested by `testStoredProcedureQuerySysCursor_Named`

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/StoredProcedureQueryImpl.java
```diff
@@ -437,13 +437,18 @@ protected StoredProcedureCall getCall() {
...
+                    field = ((OutputParameterForCallableStatement) parameter).getOutputField();
+                } else if (parameterType == getCall().OUT_CURSOR) {
                     if (parameter instanceof OutputParameterForCallableStatement) {
-                        // Case: Oracle OUT_CURSOR after execution.
                         field = ((OutputParameterForCallableStatement) parameter).getOutputField();
+                    } else {
+                        field = (DatabaseField) parameter;
                     }
+                } else if (parameterType == getCall().LITERAL) {
+                    field = (DatabaseField) parameter;
```
dazey3: Is LITERAL supposed to do anything? I mean, the old code flow was: "If LITERAL, only do anything if it's also an `OutputParameterForCallableStatement`". Presumable, the `instanceof` is covered under OUT and OUT_CURSOR now, right? So LITERAL doesnt need to do anything??? Maybe just keep LITERAL as the old implementation and add in OUT/OUT_CURSOR?

dazey3:
Also, can we get this backported to 2.7 branch as well?

Issue #625, opened 2019-12-05
NullPointerException QueryBasedValueHolder getValue(UnitOfWorkImpl uow) not thread safe


Thorsten356:
For a detailed description and example code see attachment. Tested with eclipselink 2.7.5.

[testEclipselink20QueryBasedValueHolderGetValue.zip](https://github.com/eclipse-ee4j/eclipselink/files/3927858/testEclipselink20QueryBasedValueHolderGetValue.zip)
[testEclipselink20QueryBasedValueHolderGetValue.pdf](https://github.com/eclipse-ee4j/eclipselink/files/3927860/testEclipselink20QueryBasedValueHolderGetValue.pdf)

PR #654, opened 2020-01-02
[master] Bug #625 - NullPointerException QueryBasedValueHolder getValue(UnitOfWorkImpl uow) not thread safe

rfelcman:
This is fix for bug #625 - NullPointerException QueryBasedValueHolder getValue(UnitOfWorkImpl uow) not thread safe

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/indirection/UnitOfWorkValueHolder.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/indirection/UnitOfWorkValueHolder.java
@@ -163,10 +163,10 @@ public abstract class UnitOfWorkValueHolder extends DatabaseValueHolder implemen
                         return wrapped.instantiateForUnitOfWorkValueHolder(this);
                     }
                 }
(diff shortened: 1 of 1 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/indirection/UnitOfWorkValueHolder.java
```diff
@@ -162,17 +163,18 @@ protected Object instantiateImpl() {
...
                     }
                 }
-            }
-            if (!this.wrappedValueHolder.isInstantiated()){
-                //if not instantiated then try and load the UOW versions to prevent the whole loading from the cache and cloning
-                //process
-                Object result = ((DatabaseValueHolder)this.wrappedValueHolder).getValue((UnitOfWorkImpl) this.session);
-                if (result != null){
-                    return result;
+                if (!((DatabaseValueHolder)this.wrappedValueHolder).isInstantiated()){
```
lukasj: is this explicit cast required here? If so, would it make sense to store/cache the result of the cast in some local field and reuse it a line or two later?

PR #469, opened 2019-06-13
Bug 548125** - SDO Types generation is very slow for big schemas

asanitckii:
Fixes #467

Performance fixes:

1. Avoid creating a Vector object for each NamespaceResolver#getNamespaces() call
2. Use a simpler List structure instead LinkedHashMap for the interim buffer in the NamespaceResolver#put(String, String) method.

Signed-off-by: asanitskii <asanitsky@informatica.com>

PR #678, opened 2020-01-18
Bug 548125: SDO Types generation is very slow for big schemas

asanitckii:
Author: Aleksandr Sanitckii <asanitsky@informatica.com>

Performance fixes:

Avoid creating a Vector object for each NamespaceResolver#getNamespaces() call
Use a simpler List structure instead LinkedHashMap for the interim buffer in the NamespaceResolver#put(String, String) method.
Signed-off-by: Aleksandr Sanitckii <asanitsky@informatica.com>

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/NamespaceResolver.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/NamespaceResolver.java
@@ -14,17 +14,21 @@
 //     Oracle - initial API and implementation from Oracle TopLink
 package org.eclipse.persistence.internal.oxm;
 
+import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Vector;
 
+import org.eclipse.persistence.internal.oxm.util.NamespaceResolverStorage;
 import org.eclipse.persistence.platform.xml.XMLNamespaceResolver;
 import org.eclipse.persistence.platform.xml.XMLPlatformFactory;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
+
+import static org.eclipse.persistence.internal.oxm.util.VectorUtils.emptyVector;
 
 /**
  * <p >It is common for an XML document to include one or more namespaces.
@@ -58,14 +62,15 @@ import org.w3c.dom.Node;
  */
 public class NamespaceResolver implements XMLNamespaceResolver {
     private static final String BASE_PREFIX = "ns";
+    private static final Vector EMPTY_VECTOR = emptyVector();
 
     private String defaultNamespaceURI;
-    private Map<String, String> prefixesToNamespaces;
-    int prefixCounter;
+    private NamespaceResolverStorage prefixesToNamespaces;
+    private int prefixCounter;
     private Node dom;
 
     /**
-    * Default constructor, creates a new NamespaceResolver.
+     * Default constructor, creates a new NamespaceResolver.
      */
     public NamespaceResolver() {
         super();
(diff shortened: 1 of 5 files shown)


asanitckii:
Th original issue [469](https://github.com/eclipse-ee4j/eclipselink/pull/469) was closed in order to migrate the changes on maven build.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/util/NamespaceResolverStorage.java
```diff
@@ -0,0 +1,159 @@
...
+    public NamespaceResolverStorage(int initialCapacity) {
+        super(initialCapacity);
+    }
+
+    private void readObject(java.io.ObjectInputStream in)
+        throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        namespaces = emptyVector();
+        modified = true;
+    }
```
rfelcman: Is this method needed?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/util/NamespaceResolverStorage.java
```diff
@@ -0,0 +1,159 @@
+package org.eclipse.persistence.internal.oxm.util;
```
rfelcman: Please move to NamespaceResolverStorage to the org.eclipse.persistence.internal.oxm package.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/util/VectorUtils.java
```diff
@@ -0,0 +1,243 @@
+package org.eclipse.persistence.internal.oxm.util;
```
rfelcman: Please move to VectorUtils to the org.eclipse.persistence.internal.oxm package.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/NamespaceResolver.java
```diff
@@ -14,18 +14,22 @@
...
 import java.util.Map.Entry;
 import java.util.Vector;
 
+import org.eclipse.persistence.internal.oxm.util.NamespaceResolverStorage;
 import org.eclipse.persistence.platform.xml.XMLNamespaceResolver;
 import org.eclipse.persistence.platform.xml.XMLPlatformFactory;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
 
+import static org.eclipse.persistence.internal.oxm.util.VectorUtils.emptyVector;
```
rfelcman: It could be removed after NamespaceResolverStorage and VectorUtils move. VectorUtils.emptyVector is used only once, I don't think, that static import is required there.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/util/NamespaceResolverStorage.java
asanitckii: This is a part of the Serializable interface implementation. Seems in some cases it was used for deserialization. Unfortunately the testing was done long ago, so I'm not sure now.

rfelcman:
Please move copyright to the top (to the first line) of every new/modified class. In the `NamespaceResolver` class just change year in the origin header and remove copyright header after imports.
Some example class see: https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/indirection/UnitOfWorkValueHolder.java
I'd like remove `private void readObject(java.io.ObjectInputStream in)` method from `NamespaceResolverStorage` class to see test results after change in Travis-CI. After every new commit Travis-CI will execute some tests.
Another reason from my point of view is security during deserialization.

asanitckii:
All noticed issues were addressed, please, let me know if anything else is required.

Issue #628, opened 2019-12-07
Bug 392762: Add support for Oracle IN Clause parameter limit


dazey3:
Linking to [BZ #392762](https://bugs.eclipse.org/bugs/show_bug.cgi?id=392762)

PR #685, opened 2020-01-23
[master] Bug 392762: Add Oracle support for IN Clause parameter limit

dazey3:
for #628

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/OraclePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/OraclePlatform.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2019 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -1235,4 +1235,8 @@ public class OraclePlatform extends org.eclipse.persistence.platform.database.Da
         }
     }
 
+    @Override
+    public int getINClauseLimit() {
+        return 1000;
+    }
 }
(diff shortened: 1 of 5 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/OraclePlatform.java
```diff
@@ -1235,4 +1235,8 @@ public boolean checkTableExists(final DatabaseSessionImpl session,
         }
```
rfelcman: Copyright change?

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/criteria/TestCriteriaBuilder.java
```diff
@@ -107,4 +111,70 @@ public void testCriteriaCompoundSelectionModel() throws Exception {
...
+                em.getTransaction().rollback();
+            }
+            if(em.isOpen()) {
+                em.close();
+            }
+        }
+    }
+
+    @Test
+    public void testCriteriaBuilder_NOTIN_ClauseLimit() throws Exception {
```
rfelcman: Oracle looks OK, but against MySQL this test fails as You can see from Travis-CI.

dazey3: Interesting! Ill need to revisit this and figure out whats going wrong.

dazey3:
@rfelcman Ok, so digging into the failure, I figured out whats going wrong. The issue is that my tests are using Arrays as the parameter type and I'm hitting this bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=349728
I could just update the test to use Collection types, but Im in the code right now, so II am going to look into fixing that bug first and then return to this after.

lukasj:
@dazey3 is this ready? It is hard to say since you did force-push instead of simple push which would keep track of your work.

btw: I'd strongly suggest to not do force-push unless you are immediately merging the stuff to make reviews easier. You could also get to the same result by using 'squash and merge' after the PR is approved.

dazey3:
@lukasj yes, this is ready for review! Thanks for taking a look! I ended up changing the test to use List instead of Array. We seem pretty hardcoded to expect Collections and adding support for Array was more than I wanted to tackle atm.

> force-push instead of simple push which would keep track of your work
I'm not sure I understand why force push would mess anything up here. I `--amend` my changes to the latest commit, so there's only ever one commit and thus not multiple to form a history of changes. I use '--amend' mostly to eliminate the case where "Squash and merge" isnt selected, which would create a pretty gross commit history.

lukasj:
> I'm not sure I understand why force push would mess anything up here.

With simple push one could see and review only the part which was changed and became `hidden` by `--amend`. To rearrange/squash commits later, I'm doing `git rebase -i HEAD~N` where N is number of commits I want to alter.

Issue #709, opened 2020-02-14
EclipseLink doesn't update field marked with @ReturnInsert annotation of composite Id marked with @Embeddable


rfelcman:
Eclipselink throws IllegalArgumentException if there are database mappings with multiple mappings for a single database table column.   
There is a table with virtual column participating in the primary key e.g.
```
CREATE TABLE employee  (
    id NUMBER (2) AS ( TO_NUMBER(TO_CHAR("hiredate",'DD')) ) VIRTUAL NOT NULL ,
    hiredate DATE NOT NULL ,
    salary   NUMBER (15) NOT NULL
  );
ALTER TABLE employee ADD CONSTRAINT pk_employee PRIMARY KEY ( id, hiredate, salary );
```
and secondary table with foreign key like this
```
ALTER TABLE jobs ADD CONSTRAINT fk_jobs FOREIGN KEY ( id, hiredate, salary ) REFERENCES employee ( id, hiredate, salary ) NOT DEFERRABLE ;
```
primary key class contains
```
@Embeddable
public class JobPK implements Serializable {

	private static final long serialVersionUID = 123456L;

	@ReturnInsert(returnOnly = true)
	@Column(name = "id")
	private long id;

	@Temporal(TemporalType.TIMESTAMP)
	@Column(name = "hiredate", insertable = false, updatable = false)
	private java.util.Date hireDate;
```

this is a valid configuration however, the EclipseLink code in use with Coherence integration did not support those mappings.
This bug happens only when Oracle Coherence cache integration is used.

Error stack contains:
```
Exception Description: An internal error occurred accessing the primary key
object [.......].
Internal Exception: java.lang.IllegalArgumentException: Can not set long
field ....... to
java.sql.Timestamp
Descriptor:
RelationalDescriptor(....... -->
[DatabaseTable(.......)])
at
org.eclipse.persistence.exceptions.DescriptorException.errorUsingPrimaryKey(De
scriptorException.java:1937)
at
org.eclipse.persistence.internal.jpa.CMP3Policy$FieldAccessor.setValue(CMP3Pol
icy.java:689)
at
org.eclipse.persistence.descriptors.CMPPolicy.setFieldValue(CMPPolicy.java:323
)
at 
org.eclipse.persistence.descriptors.CMPPolicy.createPrimaryKeyInstanceFromPrim
aryKeyValues(CMPPolicy.java:417) 
```

PR #710, opened 2020-02-14
[master] JPA Eclipselink Application Throws IllegalArgumentException With Coherence Enabled

rfelcman:
This is fix for #709 .

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
@@ -2064,18 +2064,34 @@ public class ClassDescriptor extends CoreDescriptor<AttributeGroup, DescriptorEv
         }
     }
 
+    /**
+     * INTERNAL:
+     * Return fields used to build insert statement.
+     */
     public Vector<DatabaseField> getReturnFieldsToGenerateInsert() {
         return this.returnFieldsToGenerateInsert;
     }
 
+    /**
+     * INTERNAL:
+     * Return fields used to build update statement.
+     */
     public Vector<DatabaseField> getReturnFieldsToGenerateUpdate() {
         return this.returnFieldsToGenerateUpdate;
     }
 
+    /**
+     * INTERNAL:
+     * Return fields used in to map into entity for insert.
+     */
     public Vector<DatabaseField> getReturnFieldsToMergeInsert() {
         return this.returnFieldsToMergeInsert;
     }
 
+    /**
+     * INTERNAL:
+     * Return fields used in to map into entity for update.
+     */
     public Vector<DatabaseField> getReturnFieldsToMergeUpdate() {
         return this.returnFieldsToMergeUpdate;
     }
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
@@ -171,8 +171,8 @@ public class ClassDescriptor extends CoreDescriptor<AttributeGroup, DescriptorEv
     protected transient List<DatabaseField> allSelectionFields;
     protected transient Vector<DatabaseField> returnFieldsToGenerateInsert;
     protected transient Vector<DatabaseField> returnFieldsToGenerateUpdate;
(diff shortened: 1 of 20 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
```diff
@@ -3993,11 +4031,57 @@ public void postInitialize(AbstractSession session) throws DescriptorException {
...
     }
 
+    private void postInitializeReturningPolicies() {
+        //Initialize ReturningPolicies
+        List<ReturningPolicy> returningPolicies = new ArrayList<>();
+        if (this.hasReturningPolicy()) {
+            returningPolicies.add(this.getReturningPolicy());
+        }
+        for (DatabaseMapping databaseMapping :this.getMappings()) {
+            ClassDescriptor referenceDescriptor = databaseMapping.getReferenceDescriptor();
```
GordonYorke: We only want the returning policies from the referenceDescriptor when the type is an AggregateObjectMapping otherwise we will get returning policies for related tables and not just the tables of this entity.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatasourceCallQueryMechanism.java
```diff
@@ -379,8 +381,8 @@ public void insertObject() throws DatabaseException {
             shouldAcquireValueAfterInsert = descriptor.getSequence().shouldAcquireValueAfterInsert();
         }
         Collection returnFields = null;
-        if (descriptor.hasReturningPolicy()) {
-            returnFields = descriptor.getReturningPolicy().getFieldsToMergeInsert();
+        if (descriptor.hasReturningPolicy() && descriptor.getReturnFieldsToMergeInsert() != null) {
```
GordonYorke: Don't really need to check for returningPolicy anymore as we did that during descriptor init

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
rfelcman: Fixed, see line 4071

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatasourceCallQueryMechanism.java
rfelcman: Fixed

rfelcman:
> Do you have test cases that you will be checking in?

Test cases are available at jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/returninsert/TestReturnInsert.java
There are various locations of `@ReturnInsert` annotation. In the entity, embedded entity, embedded Id and parent class in entity inheritance. `org.eclipse.persistence.jpa.returninsert.model.ReturnInsertDetailEmbedded` contains @ReturnUpdate annotation too.
Currently ReturnInsert implementation and tests is limited to Oracle DB (platform).

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
```diff
@@ -3993,11 +4054,63 @@ public void postInitialize(AbstractSession session) throws DescriptorException {
...
 
+    private void postInitializeReturningPolicies() {
+        //Initialize ReturningPolicies
+        List<ReturningPolicy> returningPolicies = new ArrayList<>();
+        if (this.hasReturningPolicy()) {
+            returningPolicies.add(this.getReturningPolicy());
+        }
+        for (DatabaseMapping databaseMapping :this.getMappings()) {
+            if (databaseMapping instanceof AggregateObjectMapping) {
+                ClassDescriptor referenceDescriptor = databaseMapping.getReferenceDescriptor();
```
GordonYorke: EclipseLink does support nested aggregates so you will need to check the mappings of the reference descriptor too.  If the AggregateObject's ClassDescriptor has already been initialized you may be able to just get it's list of return policies.

rfelcman: It's changed into recursive method to analyze multi-level embedded entities/objects.

PR #734, opened 2020-03-23
Lookup BeanManager for Tomcat

ewsni:
Lookup for java:comp/env/BeanManager because it is not found on tomcat with java:comp/BeanManager


lukasj:
red cross next to eclipsefdn/eca check means that we cannot accept this contribution because the commit is not properly signed (git commit --amend -s) and it does not pass the ECA check. Have you already signed [ECA](https://accounts.eclipse.org/user/eca)?

for the other redcross - can you retry by rebasing this on top of current master, please?

lukasj:
replaced by #807

PR #743, opened 2020-04-03
BUG#561757 - MOXY UNMARSHALLING JSON: NULL VALUE IGNORED IN THE PAYLOAD

rtummuru:
Signed-off-by: Ravi Babu Tummuru <ravi.babu.tummuru@oracle.com>


rfelcman:
There is new PR #819 with fix and requested unit test.

PR #747, opened 2020-04-08
[master] eclipseLink.backup folder removal.

rfelcman:
This folder was created as a backup folder during Mavenization process and it's not needed now.


lukasj:
does it make sense to also remove 'mavenize' folder?

Issue #757, opened 2020-04-16
Bug 392762: JoinColumn with same name and referencedColumnName throws exception


dazey3:
Linking to [BZ #561664](https://bugs.eclipse.org/bugs/show_bug.cgi?id=561664)

PR #760, opened 2020-04-16
[master] Bug 561664: JoinColumn with same name and referencedColumnName throws exception

dazey3:
for #757 


dazey3:
@lukasj @rfelcman Do you know why there are so many tests failing? The tests seem to be passing on 2.7

lukasj:
as of now, there should be only few failing tests which are caused by a bug in one of the dependencies and that should be fixed within #771 The rest has been fixed already and was caused mainly by missing javax->jakarta conversion

dazey3:
@lukasj thanks for the review!

PR #765, opened 2020-04-18
[master] EclipseLink release pipeline and script.

rfelcman:
This PR contains EclipseLink release pipeline and script.
Comments:
- In _release.sh_ RELEASE_VERSION is applied by _versions-maven-plugin_. But there some modules (Oracle) where path is dynamic (based on Maven property) e.g. `${oracle.modules.subdirectory}/org.eclipse.persistence.oracle`. In this case _versions-maven-plugin_ will not update `project.parent.version`. It ignores `-Doracle.modules.subdirectory` command parameter too. _sed_ text editor is used there to change required values in POMs.
- Due a limitations of nexus-staging-maven-plugin there is some change in project POMs. Modules which are not deployed are organized in Maven profile with `<id>no-deploy</id>`. Deploy command is executed with `-Ddeploy` parameter to deactivate this profile.
-git release branch (version change and tagging) in not implemented yet

### DIFF
--- a/bundles/pom.xml
+++ b/bundles/pom.xml
@@ -32,9 +32,6 @@
 
     <modules>
         <module>eclipselink</module>
-        <module>others</module>
-        <module>distribution</module>
-        <module>p2site</module>
     </modules>
 
     <profiles>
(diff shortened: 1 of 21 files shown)


bundles/pom.xml
```diff
@@ -44,6 +41,19 @@
...
+        <profile>
+            <id>no-deploy</id>
+            <activation>
+                <property>
+                    <name>!deploy</name>
+                </property>
+            </activation>
+            <modules>
+                <module>others</module>
+                <module>distribution</module>
```
lukasj, line 53: and what about main distribution zip - this one?

rfelcman: But distribution zip is not deployed to Maven repo just only to Eclipse.org download pages. _release.sh_ creates it by `mvn --no-transfer-progress -DskipTests clean install`.

lukasj: I forgot about that... then it does not make sense to deploy rename neither, at least not yet...

PR #766, opened 2020-04-20
Jersey version jakartified

senivam:
Jersey 3.0.0-M1 is visible in cetral now, so it's possible to update its dependency to jakartified version.

Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>


lukasj:
hmm, and who will re-enable disabled tests and make sure they run/pass?

PR #786, opened 2020-05-01
Add package-to-scan feature

ajacob:
# Context

Some applications can have thousands of classes.
When JPA is configured to scan classes (`exclude-unlisted-classes` is `false` in `persistence.xml`), the time used to scan all classes can be quite long.
With reasonable application size, class scanning can take only ten's of milliseconds but for heavy application it can take a few seconds.

In order to help reducing startup time, we can restrict class scanning to a pre-configured list of known packages to avoid loading unnecessary classes metadata.

# What was done

- Introduced a new property `PersistenceUnitProperties.PACKAGES_TO_SCAN`: `eclipselink.packages-to-scan`
- Modified `PersistenceUnitProcessor.getClassNamesFromURL` to only return eligible classes

This new property takes as a value a list of packages as a string: `com.foo.bar, some.other.packages`

# Performance result

I've tested this feature with 2 applications, one with ~1000 classes and another with more than 30k classes.
`MetadataProcessor.initPersistenceUnitClasses` was faster by ~25% for the small application and ~45% for the big one.

**NOTE:** This is a port from a `2.7.4` patched version running in production

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/PersistenceUnitProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/PersistenceUnitProperties.java
@@ -3907,6 +3907,15 @@ public class PersistenceUnitProperties {
     public static final String NAMING_INTO_INDEXED = "eclipselink.jpa.naming_into_indexed";
 
     /**
+     * The "<code>eclipselink.packages-to-scan</code>" property
+     * allows you to specify a comma separated list of packages you allow to be scanned.
+     * This could help reducing startup time.
+     * <p>
+     * By default, no restriction will be applied and classes from every packages will be considered.
+     */
+    public static final String PACKAGES_TO_SCAN = "eclipselink.packages-to-scan";
+
+    /**
      * INTERNAL: The following properties will not be displayed through logging
      * but instead have an alternate value shown in the log.
      */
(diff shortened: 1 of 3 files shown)


t-beckmann (approved):

robertpatrick:
@ajacob @arjantijms I think the feature is worth keeping: the opt-in filtering works, and I didn’t find a regression in the existing default behavior. I’d ask for a small revision and rebase before merging.

I compared current master with the proposed changes using archive scans and actual persistence-unit bootstrap.

| ---- | ---- | ---- | 
| Property absent or blank | Normal discovery | Unchanged |
| Package/subpackage matching in directory, JAR and WAR archives | Unrestricted discovery | Correct filtering, including exclusion of similarly named sibling packages |
| Configuration in persistence.xml and bootstrap-map overrides | No package filtering | Filtering and override precedence work |
| Explicitly listed or ORM-mapped classes outside the selected packages | Registered | Still registered |
| Classes in separately declared jar-file entries | Scanned and registered | Still scanned and registered, regardless of the filter |
| Existing PersistenceUnitProcessor unit tests | 12 passed | 12 passed |

The PR’s two new helper tests also passed after adapting them to the current JUnit 5 test location.

The main question is the intended scope. The persistence-unit root scan receives the configuration, but the separate jar-file scan receives null properties. Consequently, a JAR used as the persistence-unit root is filtered, while a separately declared JAR is not.  That could be intentional—I wouldn’t call it a regression—but the property documentation should make it clear.

What we need:

| Action | Reason |
| ---- | ---- |
| Confirm and document whether filtering is root-only or includes declared JARs | If broader filtering is intended, pass the configuration consistently; otherwise retain and test the root-only behavior |
| Move the helper tests into the current JUnit 5 PersistenceUnitProcessorTest and add focused bootstrap coverage | Verify configuration propagation, JAR scope, and preservation of explicit/ORM registrations through the existing test infrastructure |
| Rebase and obtain passing checks | The PR currently conflicts with master and has no recorded current-head checks establishing a passing build |

With that clarification, coverage and clean CI, this looks likely low-risk. The 2023 approval is encouraging, but it doesn’t replace current validation.

PR #798, opened 2020-05-11
[master] Jenkins build - server test deactivation

rfelcman:
There is temporary server tests deactivation until JEE server with "Jakarta" packages will be available.


lukasj:
don't we want to disable them also on travis?

rfelcman:
> don't we want to disable them also on travis?

Yes it makes sense. Disabled too.

PR #801, opened 2020-05-11
[2.7] Bug 538296: Wrong month is returned if OffsetDateTime is used in JPA …

jgrassel:
for #893

Fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=538296

Corrected a number of issues in the java.time conversion logic, and added unit tests to verify conversion  operations.

Prior to this fix, the ConversionManager defaulted to the JVM's system default time zone instead of UTC.  This could lead to unexpected time zone changes, especially for an application system that  could be distributed over multiple timezones.  This fix changes the default time zone used in certain conversions* to UTC.

Applications which want to return to the prior behavior, or select a specific timezone may take advantage of one of two JVM system properties introduced by this fix:

- `org.eclipse.persistence.conversion.useTimeZone` - Set this to a value that is an acceptable input for `java.time.ZoneId.of()`.   This time zone will then be used for select conversions* with ConversionManager.
- `org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime` - Set to true to use `ZoneId.systemDefault()` (the behavior before this fix) for select conversions with ConversionManager, or false to use `ZoneOffset.UTC`.

There is an order of precedence with these properties.  If `useTimeZone` is set, and has a valid value, then `useDefaultTimeZoneForJavaTime` will be ignored.  However, if `useTimeZone` is either not set or is set to an invalid value, then, if it is set, `org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime` will be processed.  If this system property is set true, then the previous behavior of using the system's default time zone will be used; If set to false (or is unset), then ConversionManager will use UTC.

`* select conversions include conversions to LocalDateTime, OffsetDateTime, and OffsetTime)`

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
@@ -29,6 +29,7 @@ import java.sql.*;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 
+import org.eclipse.persistence.config.SystemProperties;
 import org.eclipse.persistence.exceptions.*;
 import org.eclipse.persistence.internal.core.helper.CoreConversionManager;
 import org.eclipse.persistence.internal.security.PrivilegedAccessHelper;
@@ -53,7 +54,7 @@ public class ConversionManager extends CoreConversionManager implements Serializ
     static {
         ZoneId tzoneid = null;
         try {
-            String tzone = PrivilegedAccessHelper.getSystemProperty("org.eclipse.persistence.conversion.useTimeZone");
+            String tzone = PrivilegedAccessHelper.getSystemProperty(SystemProperties.CONVERSION_USE_TIMEZONE);
             if (tzone != null) {
                 
                 try {
@@ -69,7 +70,7 @@ public class ConversionManager extends CoreConversionManager implements Serializ
         
         try {
             if (tzoneid == null) {
-                String propVal = PrivilegedAccessHelper.getSystemProperty("org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime", "false");
+                String propVal = PrivilegedAccessHelper.getSystemProperty(SystemProperties.CONVERSION_USE_DEFAULT_TIMEZONE, "false");
                 if (Boolean.parseBoolean(propVal)) {
                     tzoneid = java.time.ZoneId.systemDefault();
                 } else {
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2018 IBM Corporation and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 IBM Corporation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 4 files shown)


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -44,6 +48,31 @@
...
+            String propVal = PrivilegedAccessHelper.getSystemProperty("org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime", "false");
+            if (Boolean.parseBoolean(propVal)) {
+                tzoneid = java.time.ZoneId.systemDefault();
+            } else {
+                tzoneid = ZoneOffset.UTC;
+            }
+        }
+        
+        defaultZoneOffset = tzoneid;
+    }
```
dazey3, line 75: Pehaps it may be better design to use use the lazy initialization pattern similar to how `org.eclipse.persistence.internal.helper.ConversionManager.defaultManager` is initialized?

jgrassel: The value wouldn't be initialized until the ConversionManager type is first loaded.  This is a quick operation that only occurs once -- when the class is loaded.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -44,6 +48,31 @@
  */
 public class ConversionManager extends CoreConversionManager implements Serializable, Cloneable {
     protected Map defaultNullValues;
+    private static final ZoneId defaultZoneOffset;
+    
+    static {
+        ZoneId tzoneid = null;
+        String tzone = PrivilegedAccessHelper.getSystemProperty("org.eclipse.persistence.conversion.useTimeZone");
```
dazey3, line 55: Can we register these properties in `org.eclipse.persistence.config.SystemProperties`?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
dazey3: I am suggesting that we should not use static blocks. I think a cleaner implementation would be using lazy initialization of a singleton.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
jgrassel: Sure, I don't see a reason why not.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -44,6 +49,42 @@
...
+    static {
+        ZoneId tzoneid = null;
+        try {
+            String tzone = PrivilegedAccessHelper.getSystemProperty(SystemProperties.CONVERSION_USE_TIMEZONE);
+            if (tzone != null) {
+                
+                try {
+                    tzoneid = java.time.ZoneId.of(tzone);
+                } catch (Throwable t) {
+                    // If an invalid time zone id is supplied, then fall back to checking for checking for using
```
dazey3, line 63: I feel like there should be some sort of warning here to inform the user that the property value they supplied is invalid.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
lukasj: I agree with @dazey3 here

jgrassel: I accept the suggestion to change it.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
jgrassel: Addressed, please re-review.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/helper/ConversionManager.java
jgrassel: Addressed, please re-review.

dazey3 (approved):
LGTM! Thanks for making the changes!

PR #823, opened 2020-06-01
[2.6] Javax validation upgrade 1.1.0 -> 2.0.1

rfelcman:
This is javax.validation upgrade from 1.1.0 to 2.0.1.
There is change in Travis-CI config file too to update hibernate-validator into 6.0.20.Final and Eclipse SDK is added to generate and use local compilation dependencies.
Minor changes in MOXy JAXB tests.


moxy/eclipselink.moxy.test/src/org/eclipse/persistence/testing/jaxb/beanvalidation/BeanValidationSpecialtiesTestCase.java
```diff
@@ -292,7 +294,15 @@ public ParameterNameProvider getParameterNameProvider() {
 
         @Override
         public void close() {
+        }
 
+        @Override
```
lukasj, line 299: do not add the annotation here to allow running tests against 1.0.1

PR #675, opened 2020-01-17
BUG#559307 - ECLIPSELINK ON ALL VERSIONS CAN DEAD-LOCK FOREVER

rtummuru:
Signed-off-by: Ravi Babu Tummuru <ravi.babu.tummuru@oracle.com>

I do not have any reproducer.

PR #837, opened 2020-06-18
[2.6] Bug 559307 - EclipseLink can Dead-lock forever

rfelcman:
This is improved version of PR #675 .
Thread wait time should be passed as 

- System property e.g. `-Declipselink.concurrency.manager.sleeptime=60000`
- property from _persistence.xml_ file e.g. `<property name="eclipselink.concurrency.manager.sleeptime" value="40000"/>`

If value is not specified of equals zero origin code in `org.eclipse.persistence.internal.helper.ConcurrencyManager` with `wait()` is called.

In this implementation is `eclipselink.concurrency.manager.sleeptime` property applied to all `org.eclipse.persistence.internal.helper.ConcurrencyManager` usages like
`deployLock` in `org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl`,
`transactionMutex` in `org.eclipse.persistence.internal.sessions.AbstractSession`
and thread management in L2 caching.


rfelcman:
Don't merge it now

PR #841, opened 2020-06-23
Property: only add marshal methods

diorcety:


rfelcman:
Sorry, but I am missing there a few things:

- some description for this PR (reason why this fix was created or bug number)

- some attached unit test. It helps to understand why fix was created.
   For example some JAXB compiler tests are in 
   https://github.com/eclipse-ee4j/eclipselink/tree/master/moxy/org.eclipse.persistence.moxy/src/test/java/org/eclipse/persistence/testing/jaxb/annotations
   https://github.com/eclipse-ee4j/eclipselink/blob/master/moxy/org.eclipse.persistence.moxy/src/test/java/org/eclipse/persistence/testing/jaxb/annotations/PropertyTypeTestCases.java

- last one is more or less formal. All committers and contributors must sign CLA. Commit message must contain sign-off footer e.g.     
    `Signed-off-by: Radek Felcman <radek.felcman@oracle.com>`
 For more details please see https://github.com/eclipse-ee4j/eclipselink/wiki/Creating-a-Pull-Request 
 Commit message example https://github.com/eclipse-ee4j/eclipselink/pull/819/commits/f97e2f765074f923cf53ab35ac93116f5859eec3

diorcety:
The sorry is mine. I have quickly done the PR during a hot fix on our system.
I will try to create a test but the issue was triggered by a different JVM (OpenJ9) compare to my development machine.
The issue is that adapterCls.getMethods() returns all the classes' methods. The current code add all the methods to "marshalMethods", so marshal methods but also getClass, equals and so on.
On openjdk the marshal method is the first one to pop, which is not the case on OpenJ9, the first one is getClass
So in the second loop:
`JavaClass paramType = method.getParameterTypes()[0];` an ArrayIndexOutOfBoundsException is thrown.
The modification will only add to "marshalMethods" methods named marshal with one argument. Which seems to me the wanted behaviour from the start.

diorcety:
@rfelcman You can see the issue in the travis of PR #849 
https://travis-ci.org/github/eclipse-ee4j/eclipselink/jobs/701609851

PR #849, opened 2020-06-24
Add adoptopenjdk11-openj9 jdk to travis

diorcety:


diorcety:
Only to expose the issue of #841

rfelcman:
OK thank You for a test for  #841 .

PR #851, opened 2020-06-24
[2.7] Property: only add marshal methods - backport from master

rfelcman:
The issue is that on OpenJ9. Method adapterCls.getMethods() returns all the classes' methods. The current code add all the methods to "marshalMethods", so marshal methods but also getClass, equals and so on.
On openjdk the marshal method is the first one to pop, which is not the case on OpenJ9, the first one is getClass
So in the second loop:
JavaClass paramType = method.getParameterTypes()[0]; an ArrayIndexOutOfBoundsException is thrown.
The modification will only add to "marshalMethods" methods named marshal with one argument. Which seems to me the wanted behaviour from the start.


lukasj (approved):
LGTM

Issue #822, opened 2020-05-29
Native Query: "?" in SQL comments is treated as bindable query parameter


jnehlmeier:
We are executing SQL files using 
```
EntityManager.createNativeQuery(sqlFileContent).executeUpdate()
```
and discovered that "?" in SQL comments, e.g. `-- This is legit, isn't it?`, are detected as bindable parameters by Eclipselink. This results in an exception being thrown:

```
Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.6.v20200131-b7c997804f): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: org.postgresql.util.PSQLException: The column index 1 is out of range. Number of columns: 0.
Error Code: 0
Call: 
<SQL FILE CONTENTS HERE>
bind => [null]
```

Note the `bind => [null]` at the bottom, which tries to bind a parameter to `null`. However the SQL does not have any parameter and our code never called `setParameter()`, thus PostgreSQL complains.

rfelcman:
Sorry but from my point of view there is wrong SQL query. This is not SQL script interpret. It expects just pure SQL command like `SELECT * FROM emp`.

jnehlmeier:
Hm but `createNativeQuery` generally supports multiple SQL commands separated by `;` and also supports comments. As a consequence I would argue that comments should be handled correctly.

That we read the SQL commands from a file does not really matter. A developer could have written shorter query sequences by hand and might want to comment some of them.

If JPA spec says `createNativeQuery` is meant to be used with a single query only, then Eclipselink should probably fail executing multiple SQL commands instead of accepting it?

PR #852, opened 2020-06-25
[master] Bugfix for Native Query: "?" in SQL comments is treated as bindable query parameter

rfelcman:
Bugfix + unit test for a bug #822 .

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatasourceCall.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatasourceCall.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2019 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatasourceCall.java
```diff
@@ -484,6 +485,25 @@ public void translateCustomQuery() {
...
     }
 
+    /**
+     * INTERNAL:
+     * Remove line comments from query string.
+     * @param queryString
+     * @return Query string without line comments.
+     */
+    private String removeLineComments(String queryString) {
+        StringTokenizer multiLineTokenizer = new StringTokenizer(queryString, "\n");
```
lukasj, line 495: will this work for windows line endings (`\r\n`) as well?

rfelcman: I think, that it is tested at NamedNativeQueryJUnitTest.java  line 175

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatasourceCall.java
```diff
@@ -484,6 +485,25 @@ public void translateCustomQuery() {
...
+        StringBuffer sb = new StringBuffer();
+
+        while (multiLineTokenizer.hasMoreTokens())  {
+            String s = multiLineTokenizer.nextToken();
+            if (!"--".equals(s.trim().substring(0,2)))
```
jnehlmeier, line 500: What about comments at the end of the line?

```sql
insert into table_name (col1, col2)
  values
     (1,2), -- a comment is also possible here, right?
     (3,4);
```

jnehlmeier: Also `--` can be part of a string, e.g.

```sql
-- valid comment ? (1)
insert into table_name (textCol)
  values -- valid comment ? (2)
     ('test test -- no comment ?
-- no comment ? test test
test test');
```
Above code would remove comment (1), not detect comment (2) and accidentally remove the second "no comment" line inside a string.

robertpatrick:
@rfelcman @arjantijms This PR is not safe to merge as written.

It addresses a real defect: question marks inside SQL comments can be mistaken for bind parameters. But its preprocessing removes line breaks and can change SQL semantics.

| Locally tested case | Current master | With PR change|
| ---- | ---- | ---- |
| Leading comment containing `?` | Binding error | Fixed |
| `UPDATE` with an inline comment before the `WHERE` line | Updates the intended row | Updates all three fixture rows |
| SQL tokens separated by a newline | Works | Tokens concatenate; syntax error |
| Whitespace-only or one-character line | Works | `StringIndexOutOfBoundsException` |
| Multiline string containing a line starting with `--` | Preserves content | Deletes string content |

The most serious example:

```sql
  UPDATE REVIEW_ROWS SET LABEL = ?1 -- ordinary comment
  WHERE ID = 42
```
becomes:

``` sql
  UPDATE REVIEW_ROWS SET LABEL = ? -- ordinary commentWHERE ID = 42
```

The `WHERE` clause becomes part of the comment.

What we need from the author:

| Action | Why |
| ---- | ---- | 
| Replace the comment-removal approach with quote-aware parameter scanning that preserves SQL and line boundaries | Adding a length check alone would leave the dangerous corruption |
| Add focused regressions for these cases, including affected-row verification | Existing added tests cover leading comments but miss these failures |
| Rebase, resolve conflicts, and obtain passing expected CI | The branch is 897 commits behind; no current-head checks or statuses were returned |

The reproduction used the PR’s functional patch on current master through JPA/H2.

Issue #847, opened 2020-06-24
JPA Criteria API: Empty IN Statement leads to SQL-Error


laeubi:
Transfer of https://bugs.eclipse.org/bugs/show_bug.cgi?id=415771

dfex55:
The same SQL error appears using jpql and setting an empty list as a parameter.

Example:

jpql query:
"select e from Example e where e.id in (:myIds)"

Java Code:
<typedQuery>.setParameter("myIds", new ArrayList<>()).getResultList();

Is this intended? Could this be fixed in a similar way (maybe as part of this issue)? Or should there be a separate issue?

rfelcman:
> The same SQL error appears using jpql and setting an empty list as a parameter.
> 
> Example:
> 
> jpql query:
> "select e from Example e where e.id in (:myIds)"
> 
> Java Code:
> .setParameter("myIds", new ArrayList<>()).getResultList();
> 
> Is this intended? Could this be fixed in a similar way (maybe as part of this issue)? Or should there be a separate issue?

Sorry but JPQL queries execution is different. It more "straight" to database. These queries are not build from methods like `CriteriaBuilder` way. In this case it's up to user to pass there correct parameter value.

laeubi:
> Sorry but JPQL queries execution is different.

Can you please reopen? the original bugreport was acout JPA Criteria Queriesm JPQL was brought in by @dfex55 in the discussion...

laeubi:
Okay I missed that there are some action already so @rfelcman is this actually fixed (for JPA Criteria?)

PR #856, opened 2020-06-30
[master] JPA Criteria API: Empty IN Statement leads to SQL-Error

rfelcman:
Bug fix + unit test for #847 .

Before this fix EclipseLink for (assume the `inList` is empty or `null`)
`Predicate in = from.get("projektID").in(inList);`
produced invalid SQL query like
`SELECT ... FROM ... WHERE PROJEKTID IN ()`
After this fix generated SQL query for empty or null collection used in `.in(...)` method/expression is e.g.
`SELECT ... FROM ... WHERE PROJEKTID IN (NULL)`
Which is valid SQL query.


lukasj (approved):
LGTM

PR #874, opened 2020-08-07
Indentation disapears with custom escape handler

kristofdho:
If a custom escape handler is set in combination with the `JAXB_FORMATTED_OUTPUT` option, the tabs are not printed as the `tab()` call expects the escaped value in the `os` stream.

I haven't done any significant testing with this fix.


kristofdho:
closed in favor of the signed version

PR #875, opened 2020-08-07
Indentation disappears with custom escape handler

kkriske:
If a custom escape handler is set in combination with the JAXB_FORMATTED_OUTPUT option, the tabs are not printed as the tab() call expects the escaped value in the os stream.

I haven't done any significant testing with this fix.

Duplicate of #874 for signing reasons


robertpatrick:
@kkriske @arjantijms The fix looks sound.   I found no production-code problem with it.

The custom escape handler currently writes indentation into the document instead of the temporary indentation buffer. The PR’s one-line change uses the correct destination.

| Local validation | Result |
| ---- | ---- |
| Nested and repeated indentation with spaces/tabs | Fixed |
| Large text crossing buffer boundaries | Fixed; values preserved |
| Unformatted output and UTF-16 controls | Unchanged |
| Exact-output comparisons | Improved from 8/12 to 12/12 |
| XML text and attribute checks | All 38 samples passed in both versions |

I would ask for one exact-output regression test in the existing OXM test class. Its current test removes whitespace, so it cannot detect missing indentation.

PR #880, opened 2020-08-12
BUG545909 - INCORRECT TYPE GENERATED IN ECLIPSELINK-DBWS-OR.XML FOR N…

rtummuru:
…ESTED OBJECTS

Signed-off-by: Ravi Babu Tummuru <ravi.babu.tummuru@oracle.com>

### DIFF
--- a/utils/org.eclipse.persistence.dbws.builder/src/main/java/org/eclipse/persistence/tools/dbws/oracle/OracleHelper.java
+++ b/utils/org.eclipse.persistence.dbws.builder/src/main/java/org/eclipse/persistence/tools/dbws/oracle/OracleHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


rfelcman:
I tested this PR against Oracle DB by `mvn verify -pl :org.eclipse.persistence.dbws.builder.oracle.test -P oracle` command, but unfortunately changes in this PR leads into some current test errors. This PR broke some current business logic in EclipseLink DBWS Builder. See attached Maven test results and STDOUT.
[DBWS.Builder.test-results.tar.gz](https://github.com/eclipse-ee4j/eclipselink/files/5074252/DBWS.Builder.test-results.tar.gz)

robertpatrick:
@rtummuru @arjantijms I confirmed that the nested-table type problem still exists on the current master branch, so this is worth fixing. However, I wouldn’t merge the current patch: it changes more than the SQL type name and breaks behavior that currently works.

I compared master with this change using Oracle, loading the generated ORM/OXM before invoking the DBWS operations.

| ---- | ---- | ---- |
| Read a single detail object | Passes | Passes |
| Read a header containing two nested-table elements | Passes, preserving both values | Fails with a descriptor exception |
| Round-trip that header through an Oracle function | Fails because the array mapping uses the element’s SQL type | Still fails; XML unmarshalling also loses the element values |
| Generated collection-wrapper checks | Classes implement Collection and match the call metadata | Changed classes lose Collection behavior or no longer match the call metadata |

The distinction is that the database type must identify the SQL collection, while the Java reference must still identify each element. The patch changes both. Also, `_CollectionWrapper` isn’t just a naming convention—it selects the generated class’s Collection-implementing superclass.

The encouraging part: a separate local experiment changing only the nested mapping’s SQL type name passed all three Oracle cases above, while preserving the wrapper and element-reference checks.

What we need:

| Action | Reason |
| ---- | ---- |
| Narrow the fix to the SQL collection type, retaining Java element references and wrapper names | Fixes the reported problem without the demonstrated regressions |
| Add a focused Oracle test that reloads generated ORM/OXM and round-trips multiple nested elements | Verifies the actual values survive, not just that the generated type name looks right |
| Run that test alongside the existing object-table, VARRAY, PL/SQL collection and descriptor tests | Protects the surrounding collection behavior |
| Rebase and obtain passing checks | The current PR conflicts with master and has no recorded checks establishing a passing build |

A narrowed version looks promising and likely low-risk, but this version isn’t ready.

Issue #900, opened 2020-09-15
Bug 567013: NullPointerException from weaved entity class


dazey3:
Linking to [BZ #567013](https://bugs.eclipse.org/bugs/show_bug.cgi?id=567013)

dazey3:
I believe this issue is actually fixed with https://github.com/eclipse-ee4j/eclipselink/issues/625

PR #903, opened 2020-09-15
[master] Bug 567013: NullPointer check in weaved code

dazey3:
for #900

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


lukasj:
Isn't it callers responsibility to not call the setter with null value?

dazey3:
@lukasj Do you have some documentation that states "null" values cannot be passed to this valueholder method? If it does get passed in it seems bad that we fail with a NullPointerException. Would it not be better to simply null check the value before attempting to use it? 

```java 
public void _persistence_set_foo_vh(WeavedAttributeValueHolderInterface paramWeavedAttributeValueHolderInterface) {
    this._persistence_foo_vh = paramWeavedAttributeValueHolderInterface;
    if (paramWeavedAttributeValueHolderInterface != null && paramWeavedAttributeValueHolderInterface.isInstantiated()) {
        Foo foo1 = _persistence_get_foo();
        Object object = paramWeavedAttributeValueHolderInterface.getValue();
        if (foo1 != object)
            _persistence_set_foo((Foo)object); 
        } else {
            this.user = null;
        } 
    }
}
```

lukasj:
well, I could ask the same question the other way around: "Do you have documentation that states "null" values can be passed to this valueholder method?" Do you now what exactly one has to do to pass "null" there? Is it user who can do that? Is it some eclipselink code passing "null" there? A test would really help here...

lukasj:
hmm, so when `DatabaseMapping.getAttributeValueFromObject()` is returning `{UnitOfWorkQueryValueHolder: not instantiated}` shouldn't `BasicIndirectionPolicy.getOriginalValueHolder()` return  `{UnitOfWorkQueryValueHolder: not instantiated}` too instead? Just asking...

lukasj (approved):
I would rather see the change on a different place but I can live with your solution

dazey3:
@lukasj I really appreciate you looking into this issue. I also agreed with your assessment that without a test, this change was not satisfactory. After reinvestigating, I think the fix for [514486](https://github.com/eclipse-ee4j/eclipselink/issues/625) fixes the issue I was seeing in 2.6_WAS. Thank you.

PR #906, opened 2020-09-17
[master] MOXy - JAXBContext and JAXBContextFactory promotion as Java service in META-INF/services.

rfelcman:
This commit promotes `org.eclipse.persistence.jaxb.JAXBContext` and `org.eclipse.persistence.jaxb.JAXBContextFactoryService` classes as a Java services into `jakarta.xml.bind.JAXBContext` `jakarta.xml.bind.JAXBContextFactory` in `META-INF/services` in _org.eclipse.persistence.moxy-*.jar_ and _eclipselink.jar_ artifacts.
 Access to EclipseLink JAXB implementation via `META-INF/services` is for example used in JAXB-TCK tests.

### DIFF
--- a/moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/JAXBContextFactoryService.java
+++ b/moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/JAXBContextFactoryService.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


arjantijms:
@rfelcman In GF this may be the cause of many issues we're seeing. See https://github.com/eclipse-ee4j/eclipselink/issues/1380

lukasj:
@arjantijms how is the API supposed to find the implementation through service loader if there is no services file/provides definition?

arjantijms:
@lukasj how can be distinguish between two competing implementations of the service in GF now?

// cc @dmatej @dmak

dmak:
> @lukasj how can be distinguish between two competing implementations of the service in GF now?

Thanks for including me. If I have the "classpath hell" issue, I would probably first try to solve it using [maven-shade-plugin](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html).

PR #917, opened 2020-09-25
RelationshipsTestSuite offset now accounts for daylight savings and any other offset

### LINKED ISSUE #916: Build tests fail when DST is active
Doing `mvn clean install` fails build tests at dbws.testing.relationships.RelationshipsTestSuite.getAllEmployees(RelationshipsTestSuite.java:864) due to the returning XML being unexpected, the data in the `<end-time>` tags being wrong, it is returned in the form `<end-time>00:00:00+01:00</end-time>` when the expected value is `<end-time>00:00:00Z</end-time>`. This occurs for me running in the UK with daylight savings (GMT+1)

Cousjava:
This fixes #916 .
Tested by setting the computer time zone to various ones around the world, and rerunning the test.


lukasj (approved):
LGTM

PR #919, opened 2020-10-02
[master] Maven build - performance test fix

rfelcman:
There is dependency fix for _jakarta.persistence_ and JMH + _exec-maven-plugin_ upgrade.
Performance test should be executed by `mvn test -pl :org.eclipse.persistence.performance.test -Ptest-performance` comand.


lukasj (approved):
LGTM

Issue #922, opened 2020-10-15
Bug 567891: Boolean literals are replaced with Integer values in CASE queries


dazey3:
Linking to [BZ #567891](https://bugs.eclipse.org/bugs/show_bug.cgi?id=567891)

PR #925, opened 2020-10-15
[master] Bug 567891: Boolean literals are replaced with Integer values in CASE queries

dazey3:
for #922


lukasj (approved):
LGTM

PR #927, opened 2020-10-16
[master] Bug/New Feature 559307: EclipseLink on all versions can dead-lock forever

rfelcman:
This new feature adds diagnostic logic to detect and log thread dead-lock, or potential dead-lock issues related with access to L2 entity cache.
Main purpose of this patch/new feature is identify threads/user code which leads into dead-lock above L2 entity cache.
There are multiple new system properties which control dead-lock detection mechanism and amount of messages produced in the log output. Dead-lock detection is by default disabled.
Most functional changes are in `ConcurrencyManager` and `WriteLockManager` classes. Major part of dead-lock detection logic is in the new `ConcurrencyUtil` class.
How this dead-lock issue should be simulated? In general it should happen when one JPA `EntityManagerFactory` ,`EntityManager` and entities are shared across multiple threads, which is not right, but could happen.
Attached Unit test `CacheDeadLockDetectionTest` create 100 main threads + additional threads created by CDI/Weld container (async events). I couldn't simulate this issue without CDI/Weld. It depends on No of CPU cores and other system resources. No of threads in Unit test should be changed by `CacheDeadLockDetectionTest.NO_OF_THREADS` constant. 


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyManager.java
```diff
@@ -46,26 +45,58 @@
...
+
+    /**
+     * In the places where we use this constant normally if a thread is stuck it is because it is doing object building.
+     * Blowing the threads ups is not that dangerous. It can be very dangerous for production if the dead lock ends up
+     * not being resolved because the productive business transactions will become cancelled if the application has a
+     * limited number of retries to for example process an MDB. However, the code spots where we use this constant are
+     * not as sensible as when the write lock manager is starving to run commit.
+     *
+     */
+    private static final Boolean ALLOW_INTERRUPTED_EXCEPTION_TO_BE_FIRED = true;
```
lukasj, line 90: this cannot be set to a different value - is it expected?

lukasj (commented):
* Is there a reason to prefer `new Date()` on various places over `System.currentTimeMillis()`? The former just wraps the latter; by using the latter, one can work directly with `long` instead of `Object`. 

* Would it make sense to allow setting relevant properties through `persistence.xml`? System properties are going to affect all apps in given JVM...

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyManager.java
```diff
@@ -582,4 +757,130 @@ private static String getPropertyRecordStackOnLock() {
...
+    protected void removeReadLockFromReadLockManager() {
+        Thread currentThread = Thread.currentThread();
+        ReadLockManager readLockManager = getReadLockManager(currentThread);
+        if (readLockManager != null) {
+            ConcurrencyManager concurrencyManagerCacheKey = this;
+            readLockManager.removeReadLock(concurrencyManagerCacheKey);
+            removeReadLockManagerIfEmpty(currentThread);
+        } else {
+            // We have a problem we do not want ever see a decrement on the number of readers if we
+            // are not tracing one or more precessro add read lock keys.
```
lukasj, line 843: typo

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyManager.java
```diff
@@ -582,4 +757,130 @@ private static String getPropertyRecordStackOnLock() {
...
+     * Same as {@link #getReadLockManager(Thread)} but in this case a not null result is ensured
+     *
+     * @param thread
+     *            the thread wanting its read lock manager
+     * @return the read lock manager for the current thread.
+     */
+    protected static ReadLockManager getReadLockManagerEnsureResultIsNotNull(Thread thread) {
+        Map<Thread, ReadLockManager> readLockManagers = getReadLockManagers();
+        if (!readLockManagers.containsKey(thread)) {
+            readLockManagers.putIfAbsent(thread, new ReadLockManager());
```
lukasj, line 863: would it make sense to directly return newly created `ReadLockManager` from here instead of doing additional `get` later?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
```diff
@@ -0,0 +1,1490 @@
...
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+import java.security.AccessController;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class ConcurrencyUtil {
+
+    public static final ConcurrencyUtil SINGLETON = new ConcurrencyUtil();
```
lukasj, line 39: Is this supposed to be real singleton or is it the variable name causing confusion on my end?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
```diff
@@ -0,0 +1,1490 @@
...
+            writer.write(TraceLocalization.buildMessage("concurrency_util_summary_read_locks_on_thread_step002_7", new Object[] {releaseReadLockProblemNr, currentReleaseReadLockProblem}));
+        }
+        writer.write("\n\n");
+        return writer.toString();
+    }
+
+    private static String getPropertyConcurrencyManagerAcquireWaitTime() {
+        return (PrivilegedAccessHelper.shouldUsePrivilegedAccess()) ?
+                AccessController.doPrivileged(new PrivilegedGetSystemProperty(SystemProperties.CONCURRENCY_MANAGER_ACQUIRE_WAIT_TIME))
+                : System.getProperty(SystemProperties.CONCURRENCY_MANAGER_ACQUIRE_WAIT_TIME, "0");
```
lukasj, line 755: Privileged call does not use default value, non-privileged does use one - intentional? Also note that [`Long.getLong()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#getLong-java.lang.String-java.lang.Long-) could probably be used here as I see no handling of invalid values on this place.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
```diff
@@ -0,0 +1,1490 @@
...
+                writer.write("\n\n");
+            }
+        }
+
+        // (d) We have some more information to pump out namely errors we have traced each time the number of readers was decremented
+        writer.write("\n\n");
+        writer.write(TraceLocalization.buildMessage("concurrency_util_summary_read_locks_on_thread_step002_6", new Object[] {threadName, readLockManager.getRemoveReadLockProblemsDetected().size()}));
+        long releaseReadLockProblemNr = 0;
+        for (String currentReleaseReadLockProblem : readLockManager.getRemoveReadLockProblemsDetected()) {
+            releaseReadLockProblemNr++;
```
lukasj, line 745: why not using "old" `for` syntax (`for (int x =...; x < ...; x++)`) if you need to know the index?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
```diff
@@ -0,0 +1,1490 @@
...
+        final long currentThreadId = currentThread.getId();
+
+        try {
+            // (a) search for the stack trace of the current
+            final ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+            final ThreadInfo[] threadInfos = threadMXBean.getThreadInfo(new long[] { currentThreadId }, 700);
+            for (ThreadInfo threadInfo : threadInfos) {
+                final StringWriter writer = new StringWriter();
+                enrichGenerateThreadDumpForThreadInfo(writer, threadInfo);
+                return writer.toString();
```
lukasj, line 781: This looks strange - asking for an array of threads  while passing in just one id and looping through the result expecting to handle first item only. Isn't different [`ThreadMXBean.getThreadInfo()`](https://docs.oracle.com/javase/8/docs/api/java/lang/management/ThreadMXBean.html#getThreadInfo-long-int-) what you want?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
```diff
@@ -0,0 +1,1490 @@
...
+        return new ConcurrencyManagerState(
+                readLockManagerMapClone, deferredLockManagerMapClone, unifiedMapOfThreadsStuckTryingToAcquireWriteLock,
+                mapThreadToWaitOnAcquireReadLockClone, setThreadWaitingToReleaseDeferredLocksClone,
+                mapOfCacheKeyToDtosExplainingThreadExpectationsOnCacheKey,
+                mapThreadToObjectIdWithWriteLockManagerChangesClone);
+    }
+
+
+    //TODO remove after EclipseLink logging
+    public String createHeader(String msg) {
```
lukasj, line 602: method forgotten to be removed?

rfelcman: It's removed now.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyManager.java
rfelcman: It's transformed into JPA persistence property.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java
rfelcman: Yes it's real singleton. Private constructor was added.

rfelcman:
>     * Is there a reason to prefer `new Date()` on various places over `System.currentTimeMillis()`? The former just wraps the latter; by using the latter, one can work directly with `long` instead of `Object`.

In most places `new Date()` was replaced by recomended `System.currentTimeMillis()`. It's more meaningful to measure time intervals with this.
> 
>     * Would it make sense to allow setting relevant properties through `persistence.xml`? System properties are going to affect all apps in given JVM...
Done. All new system properties were converted into `persistence.xml` properties.

lukasj (commented):
My suggestion was to *keep* system props and just *add* ability to configure same options per PU (through persistence.xml)

rfelcman:
> My suggestion was to _keep_ system props and just _add_ ability to configure same options per PU (through persistence.xml)

OK system properties are back now.

lukasj (approved):
LGTM

mafaul:
We took this solution for a spin and found these logs. Hope it helps in getting the root cause fixed.
Our config was:
```
<property name="eclipselink.concurrency.manager.waittime" value="40000"/>
<property name="eclipselink.concurrency.manager.maxsleeptime" value="40000"/>
<property name="eclipselink.concurrency.manager.maxfrequencytodumptinymessage" value="120000"/>
<property name="eclipselink.concurrency.manager.maxfrequencytodumpmassivemessage" value="300000"/>
<property name="eclipselink.concurrency.manager.allow.readlockstacktrace" value="false"/>
<property name="eclipselink.concurrency.manager.allow.concurrencyexception" value="true"/>
<property name="eclipselink.concurrency.manager.allow.interruptedexception" value="true"/>
```
**Thread Dump:**
[threadDump.txt](https://github.com/eclipse-ee4j/eclipselink/files/6076205/threadDump.txt)

**Log File:**
[logback.2021-03-03.1.log](https://github.com/eclipse-ee4j/eclipselink/files/6076235/logback.2021-03-03.1.log)

Awesome work so far, thank you.

rfelcman:
In this case provided log output looks OK. It seems, that there is some code call (query) which keeps lock in JPA L2 cache longer than provided 40000 `<property name="eclipselink.concurrency.manager.waittime" value="40000"/>`.
Locked object/entity in the JPA L2 means e.g. new/updated entity complete initialization time. It means long running query to initialize all entity properties include e.g. `@OneToMany fetch=EAGER`.
From provided log output it should be something behind:
`        at za.co.pickxx.entity.pickxxEntity.getLatest(pickxxEntity.java:262)
        at za.co.pickxx.entity.pickxxEntity.getLatestStatusType(pickxxEntity.java:317)
`
or
`at za.co.pickxx.ejb.platform.SystemUserFacade.getByEmail(SystemUserFacade.java:236)`

PR #928, opened 2020-10-16
[Master] Revert "Change UoW reference to WeakReference for GC"

jgrassel:
This reverts commit 1b94ba6ae84d5636656b25b1a94942d3e17f84a7.

Open Liberty JPA Relationship tests using Eclipselink 3.0 on Open Liberty began failing randomly after moving to a commit after M-1, with a NPE:

```
java.lang.NullPointerException
	at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.updateChangeRecord(AbstractDirectMapping.java:1163)
	at org.eclipse.persistence.internal.descriptors.changetracking.AttributeChangeListener.internalPropertyChange(AttributeChangeListener.java:162)
	at org.eclipse.persistence.internal.descriptors.changetracking.AttributeChangeListener.propertyChange(AttributeChangeListener.java:115)
	at com.ibm.ws.jpa.fvt.relationships.manyXmany.entities.bi.annotation.MMBiEntA._persistence_propertyChange(MMBiEntA.java)
```

The NPE was a result of `objectChangeSet` being null:

```
        UnitOfWorkImpl uow = getUnitOfWork();
        if(uow != null) {
            if (uow.getUnitOfWorkChangeSet() == null) {
                uow.setUnitOfWorkChangeSet(new UnitOfWorkChangeSet(uow));
            }
            if (objectChangeSet == null) {//only null if new or if in a new UOW
                //add to tracker list to prevent GC of clone if using weak references
                //put it in here so that it only occurs on the 1st change for a particular UOW
                uow.addToChangeTrackedHardList(owner);
                objectChangeSet = getDescriptor().getObjectBuilder().createObjectChangeSet(owner, (UnitOfWorkChangeSet) uow.getUnitOfWorkChangeSet(), false, uow);
            }
        }
```

Because the change to be reverted made the UOW reference a weak reference, it could be GC'd (which contributed to the random appearance of the failure, tests failed 25% of the time), it creates a situation where objectChangeSet would not be created if it comes in as null.  Which sets up the later NPE.

This PR reverts 1b94ba6ae84d5636656b25b1a94942d3e17f84a7; after reverting it, I have not been able to observe the failure after 30 retries (usually 3-4 attempts were necessary to recreate).  So I am 100% confident this change introduced the problem.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/changetracking/AttributeChangeListener.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/changetracking/AttributeChangeListener.java
@@ -15,7 +15,6 @@
 package org.eclipse.persistence.internal.descriptors.changetracking;
 
 import java.beans.*;
-import java.lang.ref.WeakReference;
 
 import org.eclipse.persistence.descriptors.ClassDescriptor;
 import org.eclipse.persistence.descriptors.changetracking.*;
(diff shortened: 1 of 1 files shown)


lukasj (approved):
LGTM

PR #930, opened 2020-10-19
clean up the build for eclipselink.jar

lukasj:
* clean up dependencies a bit
* remove unused properties
* changed the way jars are built to not use assembly plugin and using unpack/jar instead


rfelcman (commented):
`jakarta.persistence` and `jakarta.xml.bind` packages are excluded from _eclipselink-3.0.0-SNAPSHOT-javadoc.jar_ . Is it correct?

lukasj:
> `jakarta.persistence` and `jakarta.xml.bind` packages are excluded from _eclipselink-3.0.0-SNAPSHOT-javadoc.jar_ . Is it correct?

correct, same for sdo api

rfelcman (approved):
LGTM

Issue #893, opened 2020-09-05
Bug 538296: Wrong month is returned if OffsetDateTime is used


dazey3:
Linking to [BZ #538296](https://bugs.eclipse.org/bugs/show_bug.cgi?id=538296)

jgrassel:
Fix delivered in 2.7 and master.

PR #933, opened 2020-10-20
[Master] Bug 538296: Wrong month is returned if OffsetDateTime is used

jgrassel:
for #893

Fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=538296

Corrected a number of issues in the java.time conversion logic, and added unit tests to verify conversion  operations.

Prior to this fix, the ConversionManager defaulted to the JVM's system default time zone instead of UTC.  This could lead to unexpected time zone changes, especially for an application system that  could be distributed over multiple timezones.  This fix changes the default time zone used in certain conversions* to UTC.

Applications which want to return to the prior behavior, or select a specific timezone may take advantage of one of two JVM system properties introduced by this fix:

- `org.eclipse.persistence.conversion.useTimeZone` - Set this to a value that is an acceptable input for `java.time.ZoneId.of()`.   This time zone will then be used for select conversions* with ConversionManager.
- `org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime` - Set to true to use `ZoneId.systemDefault()` (the behavior before this fix) for select conversions with ConversionManager, or false to use `ZoneOffset.UTC`.

There is an order of precedence with these properties.  If `useTimeZone` is set, and has a valid value, then `useDefaultTimeZoneForJavaTime` will be ignored.  However, if `useTimeZone` is either not set or is set to an invalid value, then, if it is set, `org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime` will be processed.  If this system property is set true, then the previous behavior of using the system's default time zone will be used; If set to false (or is unset), then ConversionManager will use UTC.

`* select conversions include conversions to LocalDateTime, OffsetDateTime, and OffsetTime)`

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
@@ -12,6 +12,8 @@
 
 // Contributors:
 //     tware - added to allow pluggage archive factory
+//     05/11/2020-2.7.0 Jody Grassel
+//       - 538296: Wrong month is returned if OffsetDateTime is used in JPA 2.2 code
 package org.eclipse.persistence.config;
 
 /**
@@ -75,6 +77,20 @@ public class SystemProperties {
      */
     public static final String ENFORCE_TARGET_SERVER = "eclipselink.target-server.enforce";
 
+    /**
+     * This system property can be set the specific time zone used by ConversionManager to convert
+     * LocalDateTime, OffsetDateTime, and OffsetTime types.
+     */
+    public static final String CONVERSION_USE_TIMEZONE = "org.eclipse.persistence.conversion.useTimeZone";
+    
+    /**
+     * This system property can be set to restore ConversionManager behavior with converting 
+     * LocalDateTime, OffsetDateTime, and OffsetTime types back to using the JVM's default time zone instead
+     * of UTC.  This restores behavior prior to fixing Bug 538296.  This property is ignored if the
+     * System Property CONVERSION_USE_TIMEZONE has been set.
+     */
+    public static final String CONVERSION_USE_DEFAULT_TIMEZONE = "org.eclipse.persistence.conversion.useDefaultTimeZoneForJavaTime";
+    
     /**
      * This property can be set to <code>false</code> to enable UPDATE call to set
      * foreign key value in the target row in unidirectional 1-Many mapping
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2018 IBM Corporation and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 IBM Corporation and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,6 +15,8 @@
 //     Oracle - initial API and implementation from Oracle TopLink
 //     09/30/2014-2.6.0 Dalia Abo Sheasha
 //       - 445546: NullPointerException thrown when an Array of Bytes contains null values
+//     05/11/2020-2.7.0 Jody Grassel
+//       - 538296: Wrong month is returned if OffsetDateTime is used in JPA 2.2 code
 package org.eclipse.persistence.internal.helper;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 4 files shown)


dazey3 (approved):
LGTM

lukasj:
this introduced a failure on default inmemory derby DB:

```
[EL Fine]: connection: 2020-10-21 23:58:47.044--Thread(Thread[main,5,main])--Detected database platform: org.eclipse.persistence.platform.database.JavaDBPlatform
[EL Fine]: connection: 2020-10-21 23:58:47.044--ServerSession(846014380)--Connection(1872515144)--Thread(Thread[main,5,main])--connecting(DatabaseLogin(
	platform=>JavaDBPlatform
	user name=> "root"
	datasource URL=> "jdbc:derby://localhost/memory:ecltests;create=true"
))
```

when I run `mvn -U -C clean install -Pstaging -pl :org.eclipse.persistence.jpa.jse.test` on my box, I get:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.282 s - in org.eclipse.persistence.jpa.compositeid.TestCompositeId
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TestJavaTimeTypeConverter.timeConvertUtilDateToLocalDate:86 expected:<JANUARY> but was:<DECEMBER>
[INFO] 
[ERROR] Tests run: 184, Failures: 1, Errors: 0, Skipped: 8
[INFO] 
```

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
```diff
@@ -75,6 +77,20 @@
      */
     public static final String ENFORCE_TARGET_SERVER = "eclipselink.target-server.enforce";
 
+    /**
+     * This system property can be set the specific time zone used by ConversionManager to convert
+     * LocalDateTime, OffsetDateTime, and OffsetTime types.
+     */
+    public static final String CONVERSION_USE_TIMEZONE = "org.eclipse.persistence.conversion.useTimeZone";
```
lukasj, line 84: I'd prefer to stick with existing naming convention used by other properties - use `eclipselink._whatever_` instead of `org.eclipse.persistence._whatever_`

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -75,6 +84,57 @@
...
+            String tzone = null;
+            try {
+                tzone = PrivilegedAccessHelper.getSystemProperty(SystemProperties.CONVERSION_USE_TIMEZONE);
+                if (tzone != null) {
+                    try {
+                        tzoneid = java.time.ZoneId.of(tzone);
+                    } catch (Throwable t) {
+                        // If an invalid time zone id is supplied, then fall back to checking for checking for using
+                        // either UTC or the system's default time zone.
+                        if (AbstractSessionLog.getLog().shouldLog(SessionLog.WARNING)) {
```
lukasj, line 99: I agree it's better to stay safe but do we really need to do this check 3 times in a row? :-)

...there's the same check in [the log method itself](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/AbstractSessionLog.java#L397) as well as in the method [that log method calls](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/AbstractSessionLog.java#L575)

lukasj:
@jgrassel @dazey3 any chance you're going to look into test failure introduced by this change? Thx

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
jgrassel: This is a port of #801 to master, this should have been raised at that time.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
jgrassel: The first catch-block is for catching a conversion failure with java.time.ZoneId.of(tzone), for which I wanted to log a specific warning:

```
                    try {
                        tzoneid = java.time.ZoneId.of(tzone);
                    } catch (Throwable t) {
                        // If an invalid time zone id is supplied, then fall back to checking for checking for using
                        // either UTC or the system's default time zone.
                        if (AbstractSessionLog.getLog().shouldLog(SessionLog.WARNING)) {
                            AbstractSessionLog.getLog().log(SessionLog.WARNING, "invalid_tzone", 
                                    SystemProperties.CONVERSION_USE_TIMEZONE, tzone);
                        }
                    }
```

The second catch block, surrounding that, is a guard against an (unlikely) exception occurring elsewhere, like say 
```
tzone = PrivilegedAccessHelper.getSystemProperty(SystemProperties.CONVERSION_USE_TIMEZONE);
```
Which is just swallowed.

If either of the above cases threw an Exception, it would fall back to processing `SystemProperties.CONVERSION_USE_DEFAULT_TIMEZONE`

Which has its own Exception catch block, which emits a warning specific to the attempt to consume that specific property.

So yes, it needed that many catches to ensure that both a proper warning message is emitted and that the follow up property is attempted.

jgrassel:
> this introduced a failure on default inmemory derby DB:
> 
> ```
> [EL Fine]: connection: 2020-10-21 23:58:47.044--Thread(Thread[main,5,main])--Detected database platform: org.eclipse.persistence.platform.database.JavaDBPlatform
> [EL Fine]: connection: 2020-10-21 23:58:47.044--ServerSession(846014380)--Connection(1872515144)--Thread(Thread[main,5,main])--connecting(DatabaseLogin(
> 	platform=>JavaDBPlatform
> 	user name=> "root"
> 	datasource URL=> "jdbc:derby://localhost/memory:ecltests;create=true"
> ))
> ```
> 
> when I run `mvn -U -C clean install -Pstaging -pl :org.eclipse.persistence.jpa.jse.test` on my box, I get:
> 
> ```
> [INFO] Results:
> [INFO] 
> [ERROR] Failures: 
> [ERROR]   TestJavaTimeTypeConverter.timeConvertUtilDateToLocalDate:86 expected:<JANUARY> but was:<DECEMBER>
> [INFO] 
> [ERROR] Tests run: 184, Failures: 1, Errors: 0, Skipped: 8
> [INFO] 
> ```

Yeah, this is weird for a number of reasons.  First, this is basically a unit test which drives input and output directly with the ConversionManager class.  The test doesn't create EntityManagerFactories, EntityManagers, nor does it drive any kind of interaction with a database.

Also I tried running this on my own mac (I replied in email) and it passed fine (which is good, as I remember it passing when I was running it locally before creating the pull request!)

jgrassel:
It's possible that there could be some interesting and unexpected things going on because the test bucket has:

```
@RunWith(EmfRunner.class)
public class TestJavaTimeTypeConverter { 
```

I commented out the @RunWith (it doesn't need any EMF processing), and it looks like maven still ran the test bucket.  Does that change make a difference in your sandbox?

PR #943, opened 2020-10-27
[master] Added a config for LGTM.com

### LINKED ISSUE #942: Add a config for LGTM.com
LGTM.com offers static analysis for open-source project. It helps catching issues earlier (including security ones). This is based on CodeQL engine. Currently, a report for EclipseLink is available but it doesn't see to be correct. In particular, it doesn't contain any Java sources which is suspicious:

https://lgtm.com/projects/g/eclipse-ee4j/eclipselink/alerts/?mode=list

Instead, the report shows findings in the Python code that seems to be used only for testing purposes. Therefore, the current grade `E` (which is bad) doesn't look correct.

Adding a config file for LGTM.com should fix this:

https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html#language_block

I think the config may look like the following:

```
path_classifiers:

  # exclude tests
  test:
    - "**/src/test/**"
    - "**/src/it/**"
    - "**/*.test/**"
    - "**/tests/**"
    - performance

  # exclude docs
  docs:
    - about.html
    - license.html
    - readme.html

extraction:
  java:
    index:
      build_command: mvn -DskipTests clean package
```

I can open a pull request that adds a config like above. Once a correct report at LGTM.com is available, I can also triage findings and fix major ones.

artem-smotrakov:
Added a config for LGTM.com

- Excluded tests and docs
- Specified a build command

Fixes #942

Signed-off-by: Artem Smotrakov <artem.smotrakov@sap.com>

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,21 @@
+path_classifiers:
+
+  # exclude tests
+  test:
+  - "**/src/test"
+  - "**/src/it"
+  - "**/*.test"
+  - "**/tests"
+  - "performance"
+
+  # exclude docs
+  docs:
+  - "about.html"
+  - "license.html"
+  - "readme.html"
+
+extraction:
+  java:
+    index:
+      build_command: "mvn -DskipTests clean package"
+


lukasj (approved):
let's see how this is going to work

PR #948, opened 2020-10-31
first exception when reading a class file should be a warning

lukasj:


rfelcman (approved):
LGTM

PR #958, opened 2020-11-11
[master] Bug/New Feature 559307: EclipseLink on all versions can dead-lock forever 2.0

rfelcman:
There is small fix to improve readability of massive dump Page 7 of Concurrency manager state.


lukasj (approved):
LGTM

PR #965, opened 2020-11-19
support of mariadb sequences fixes #883

### LINKED ISSUE #883: MySQLPlatform lacks support of mariadb sequences
MySQLPlatform lacks support of mariadb sequences introduced in https://mariadb.com/kb/en/sequences/

Duplicating Bug 545248 - Support for mariadb 10.3 sequences https://bugs.eclipse.org/bugs/show_bug.cgi?id=545248
 

CREATE SEQUENCE was introduced in MariaDB 10.3.
 
https://mariadb.com/kb/en/sequences/

A sequence is an object that generates a sequence of numeric values, as specified by the CREATE SEQUENCE statement. Sequences are an alternative to AUTO_INCREMENT when you want more control over how sequence numbers are generated. 

Sequences can be used as alternative to table storage, but without locks when multiple threads are accessing id generation.

slavb18:
Add support of mariadb sequences
https://mariadb.com/kb/en/sequences/


slavb18:
SEQUENCES supported only on mariadb 10.3+, tests fails 

```
Internal Exception: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SEQUENCE AC_STUDENT_SEQ START WITH 25 INCREMENT BY 25' at line 1
```

May be there  should be mysql server version check  in method `supportsSequenceObjects` ?

rfelcman (changes requested):
Sorry but it seems, that this PR break compatibility with older MySQL versions (5.6, 5.7). See PR checks (Travis-CI output)
I see there possible solutions:

1. Add to EclipseLink new platform e.g. `org.eclipse.persistence.platform.database.MariaDB` and inherit and override content from `org.eclipse.persistence.platform.database.MySQLPlatform`
2. Check DB server version in origin `org.eclipse.persistence.platform.database.MySQLPlatform` see e.g. https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/SQLServerPlatform.java#L79 and put Your changes in conditions, but as I checked some changes are similar like inherited from `org.eclipse.persistence.internal.databaseaccess.DatabasePlatform#buildSequenceObjectDeletionWriter`

arjantijms:
@slavb18 at the offchance you're still interested in this, what do you think about the feedback from @rfelcman? Could you implement that and rebase this pr? I know this is almost pre-historic, but just wandering.

PR #966, opened 2020-11-26
[2.6] Bug/New Feature 559307: EclipseLink on all versions can dead-lock forever - backport from master

rfelcman:
L2 Cache dead lock diagnostic
This new feature adds diagnostic logic to detect and log thread dead-lock, or potential dead-lock issues related with access to L2 entity cache.
Main purpose of this patch/new feature is identify threads/user code which leads into dead-lock above L2 entity cache.
There are multiple new system properties or persistence properties which control dead-lock detection mechanism and amount of messages produced in the log output. Dead-lock detection is by default disabled.
Most functional changes are in ConcurrencyManager and WriteLockManager classes. Major part of dead-lock detection logic is in the new ConcurrencyUtil class.
Due a JPA.JSE tests there are some new dependencies (jars) in plugins directory.


lukasj (approved):
LGTM

PR #971, opened 2020-12-03
[master] Jenkins build fix - Snapshots publishing

rfelcman:
This fix has two parts:
1. Bash condition `if [ ${CONTINUOUS_BUILD} = "true" ]; then` ensure, that snapshots publishing will be active only during "Nightly build".
2. Maven target change from `... clean deploy` to `... deploy` ensure, that tests results from previous Jenkins stages will remain there for last Jenkins pipeline stage "Proceed test results"


lukasj (approved):
LGTM

PR #983, opened 2020-12-08
#567891: improving fix to pass TCKs

lukasj:
@dazey3 fix in #925 for bug 567891 caused TCK failure, in particular com/sun/ts/tests/jpa/core/criteriaapi/CriteriaBuilder/Client.java#toStringTest_from_standalone test fails with:

```
[javatest.batch] 12-07-2020 14:29:09:  ERROR: java.lang.ClassCastException: java.lang.Character cannot be cast to java.lang.String
[javatest.batch] 	at com.sun.ts.tests.jpa.core.criteriaapi.CriteriaBuilder.Client.toStringTest(Client.java:6345)
[javatest.batch] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[javatest.batch] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[javatest.batch] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[javatest.batch] 	at java.lang.reflect.Method.invoke(Method.java:498)
[javatest.batch] 	at com.sun.ts.lib.harness.EETest.run(EETest.java:596)
[javatest.batch] 	at com.sun.ts.lib.harness.ServiceEETest.run(ServiceEETest.java:115)
[javatest.batch] 	at com.sun.ts.tests.common.vehicle.EmptyVehicleRunner.run(EmptyVehicleRunner.java:40)
[javatest.batch] 	at com.sun.ts.lib.harness.ServiceEETest.run(ServiceEETest.java:105)
[javatest.batch] 	at com.sun.ts.lib.harness.EETest.getPropsReady(EETest.java:486)
[javatest.batch] 	at com.sun.ts.lib.harness.ServiceEETest.run(ServiceEETest.java:209)
[javatest.batch] 	at com.sun.ts.lib.harness.EETest.run(EETest.java:285)
[javatest.batch] 	at com.sun.ts.tests.common.vehicle.VehicleClient.main(VehicleClient.java:38)
[javatest.batch] 
```

The problem in #925 is, IMHO, that it is altering the object to meet user expectations too late - after all processing has been done and right before returning the value to the caller. Apparently, there are cases where customized type is used during result processing but newly introduced code is changing that to some unexpected type and thus causes the failure as seen ie in the TCK test. What should have been done is to follow the path properly - on the way to the DB, boolean literals are being converted to values being understood by the DB (ie `true` -> `1`), on the way back from the DB, these values should be translated back to types eclipselink uses internally (ie `1` -> `true`). I believe this last step is what was missing, since eclipselink knows that the type retrieved from the DB is `Boolean`, it should have been just using proper method to "decode" returned value properly - this PR does exactly that.

@rfelcman as to why the TCK test is failing - the code introduced by @dazey3 explicitly tries to convert the value returned by the DB (which is expected `String`) to the initial type (which is `Character`) ignoring the fact that `CriteriaBuilder.toString()` has been called on that `Character`. The conversion unfortunately succeeds - conversion manager converts `String` to `Character` as `someString.charAt(0)` and the test fails with ClassCastEx because expected type is  `Character` while `String` should have been returned.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -1383,6 +1383,8 @@ public class DatabaseAccessor extends DatasourceAccessor {
         } else if ((fieldType == ClassConstants.SHORT) || (fieldType == ClassConstants.PSHORT)) {
             value = Short.valueOf(resultSet.getShort(columnNumber));
             isPrimitive = ((Short)value).shortValue() == 0;
+        } else if (fieldType == ClassConstants.BOOLEAN) {
+            value = resultSet.getBoolean(columnNumber);
         } else if ((type == Types.TIME) || (type == Types.DATE) || (type == Types.TIMESTAMP)) {
             if (Helper.shouldOptimizeDates) {
                 // Optimize dates by avoid conversion to timestamp then back to date or time or util.date.
@@ -1424,7 +1426,7 @@ public class DatabaseAccessor extends DatasourceAccessor {
             if (value != null) return ((BigDecimal)value).toBigInteger();
         } else if (fieldType == ClassConstants.BIGDECIMAL) {
             value = resultSet.getBigDecimal(columnNumber);
-         }
+        }
 
         // PERF: Only check for null for primitives.
         if (isPrimitive && resultSet.wasNull()) {
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/ReportQueryResult.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/ReportQueryResult.java
@@ -34,7 +34,6 @@ import java.util.Set;
 import java.util.Vector;
 
 import org.eclipse.persistence.descriptors.ClassDescriptor;
-import org.eclipse.persistence.exceptions.ConversionException;
 import org.eclipse.persistence.exceptions.QueryException;
 import org.eclipse.persistence.expressions.ExpressionOperator;
 import org.eclipse.persistence.internal.expressions.FunctionExpression;
@@ -289,21 +288,6 @@ public class ReportQueryResult implements Serializable, Map {
                 }
             } else {
                 value = row.getValues().get(itemIndex);
-
-                // Verify that the expected result type matches the actual value type
-                if(value != null) {
-                    Class valueType = value.getClass();
-                    Class resultType = item.getResultType();
-                    if(!valueType.isInstance(resultType)) {
-                        try {
-                            value = query.getSession().getPlatform().convertObject(value, resultType);
-                        } catch (ConversionException e) {
-                            // If an Exception is thrown while attempting to 
-                            // convert, ignore and return the original value
-                        }
-                    }
-                }
-
                 // GF_ISSUE_395
                 if (this.key != null) {
                     this.key.append(value);


rfelcman (approved):
LGTM

dazey3:
@lukasj 

> as to why the TCK test is failing - the code introduced by @dazey3 explicitly tries to convert the value returned by the DB (which is expected String) to the initial type (which is Character) ignoring the fact that CriteriaBuilder.toString() has been called on that Character. 

Correct me if this is incorrect for this use case: 
1. The query resultset returns a String type
2. The `resultType`, from my fix, is expecting Character type
3. The result is converted then from String -> Character
4. The test fails because the test expects String type returned

> The conversion unfortunately succeeds - conversion manager converts String to Character as someString.charAt(0) and the test fails with ClassCastEx because expected type is Character while String should have been returned.

Ok, so the test is failing because it EXPECTS String type and is making an explicit class cast. Yes?

Two question:
1. Are we sure the test isn't wrong? Maybe they ARE supposed to be getting back a Character type and the explicit cast was added to the test because EclipseLink USED to return a String, even though that was technically incorrect.
2. Are we sure the `resultType` wasnt just wrong (set to Character when that wasnt the right expected type)? It wouldnt be the first time EclipseLink's ExpressionBuilderVisitor implementation set the WRONG type for a specific expression...

lukasj:
> 1. The query resultset returns a String type

Yes. The datatype returned by the database is `VARCHAR`=> `String`

> 2. The resultType, from my fix, is expecting Character type
> 3. The result is converted then from String -> Character
> 4. The test fails because the test expects String type returned

Yes, all above is correct when the issue happens.

> Are we sure the test isn't wrong? Maybe they ARE supposed to be getting back a Character type and the explicit cast was added to the test because EclipseLink USED to return a String, even though that was technically incorrect.

Feel free to file an appeal against tck (in jpa-api repo). The test in question is [here](https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/jpa/core/criteriaapi/CriteriaBuilder/Client.java#L6324) and the most important parts of it are probably:

```
cbuilder.createQuery(String.class); // <- the query result is to be of type String 
...
cquery.select(cbuilder.toString(cbuilder.literal('a'))); // <- this is where character on the input side is changed to string
...
String result = tq.getSingleResult(); // <- this is where CCE happens 
```

...so yes, I strongly believe the test is correct - it defines a query to return String but what we we're returning is Character.


> 2. Are we sure the resultType wasnt just wrong?

Good question. `resultType` is indeed of type `Character`. But I believe this is correct since the `cbuilder.toString` is just a type cast, nothing else, so it cannot change the `javaType` of the expression created by `cbuilder.literal('a')` -> that would be against definition.

dazey3:
@lukasj 
I pulled down the change and tested it against 2.6_WAS/2.7. This change looks good and I agree that it makes more sense in DatabaseAccessor than ReportQueryResult. I didnt see that location when I was originally tracing.


What I am trying to understand now is WHY my original fix failed for this use case. It may have been in the wrong location, but it seems like the conversion logic was right. As long as the `resultType`, in ReportQueryResult, is the correct value, it should have worked the same way as the fix in DatabaseAccessor. 
The reason this use case is failing is because the `resultType` is Character type when it apparently shouldnt be.

> Feel free to file an appeal against tck (in jpa-api repo). The test in question is here
Thanks for supplying a link to the test! I agree that the test appears to be correct. However, after tracing it through a bit, I'm wondering if there is an improvement we could do with CriteraBuilderImpl.

As you pointed out, this appears to be the most important part of the test:
```java 
    cquery.select(cbuilder.toString(cbuilder.literal('a'))); // <- this is where character on the input side is changed to string
```
"cbuilder.literal('a')" is returning `javax.persistence.criteria.Expression<Character>`. The ExpressionImpl.javaType is appropriately set to `java.lang.Character`. Then, what I find kind of weird is the implementation of `org.eclipse.persistence.internal.jpa.querydef.CriteriaBuilderImpl.toString(Expression<Character> character)`. It doesnt do anything... it just has an explicit cast and returns the same object. The same is true for similar methods like toBigInteger, toBigDecimal, toDouble, toFloat, toInteger, ect. Wouldn't it make more sense for the `javaType` to be set to the new type? Or anything to really happen that indicates the type is supposed to be converted?
Instead, `Character` type is set and the resultType is perpetuated all the way to execution.

lukasj:
> WHY my original fix failed for this use case

the only reason is that `ConversionManager.convertObject(...)` call you have added knows how to "cast" `String` to `Character`. It would have failed for casting `A` to `B` with appropriate exception but in `String`/`Char` case you were just an (un)lucky winner :-/


> [`CriteriaBuilderImpl.toString(Expression<Character> character)`] It doesnt do anything... it just has an explicit cast and returns the same object. 

that is exactly how it is defined by the persistence spec/API. [Javadoc](https://jakarta.ee/specifications/persistence/3.0/apidocs/jakarta.persistence/jakarta/persistence/criteria/criteriabuilder#toString(jakarta.persistence.criteria.Expression)) says exactly that:

"Typecast. Returns same expression object."

so the implementation is correct - at least in terms of it does what it is expected to do.

> Wouldn't it make more sense for the javaType to be set to the new type?

Yes, this was the first thought and actually what @rfelcman did initially. But I'm afraid this is something against the definition. Also Imagine one stores result of the `cbuilder.literal('a')` in a variable and uses it on different places applying `cbuilder.toString(..)` call on some of them. Given that an instance cannot change, what should happen in this case?
I believe that proper types are tracked trough metamodel criteriaquery is based on instead of through expressions as we know it from regular, not only eclipselink native, queries.

PR #985, opened 2020-12-08
3.0.0 release

rfelcman:


lukasj (approved):
LGTM

laeubi:
@lukasj @rfelcman 
I'm trying to use https://download.eclipse.org/rt/eclipselink/updates/ but it seems the 3.0 release is not there (I only see 2.x) the zipped bundles contains the 3.0 bundles so it seems the update-site is not yet published?

lukasj:
We don’t plan to publish update site. Is there a reason maven dependency can’t be used?

laeubi:
Quote from the [download page](https://www.eclipse.org/eclipselink/downloads/)

> The EclipseLink bundles can be obtained from a P2 repository for installation in a target platform or Eclipse IDE. 

that's my use-case and its a bit confusing when the page states that one can consume it via update-site but in fact its not there.

lukasj:
then the problem is that the page has not been fully updated for 3.0.0 and the best thing you could have done would be to just file an issue (#996)

PR #984, opened 2020-12-08
CriteriaBuilderImpl fix revert and code improvement in DatabaseAccessor to fix JPA.WDF tests.

rfelcman:
Fix #983 broken JPA.WDF tests see https://ci.eclipse.org/eclipselink/view/Current%20Jobs/job/eclipselink-nightly-master/153/console . This PR removes unnecessary change in CriteriaBuilderImpl and extends fix in DatabaseAccessor.

Signed-off-by: Radek Felcman <radek.felcman@oracle.com>

PR #987, opened 2020-12-09
[2.7] Bug 567891: Boolean literals are replaced with Integer values in CASE queries - backport

rfelcman:
This is backport from #983 and #984 .
This is simplified and more correct fix for 567891 . It partially change #925 .

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -1383,6 +1383,9 @@ public class DatabaseAccessor extends DatasourceAccessor {
         } else if ((fieldType == ClassConstants.SHORT) || (fieldType == ClassConstants.PSHORT)) {
             value = Short.valueOf(resultSet.getShort(columnNumber));
             isPrimitive = ((Short)value).shortValue() == 0;
+        } else if ((fieldType == ClassConstants.BOOLEAN) || (fieldType == ClassConstants.PBOOLEAN))  {
+            value = Boolean.valueOf(resultSet.getBoolean(columnNumber));
+            isPrimitive = ((Boolean)value).booleanValue() == false;
         } else if ((type == Types.TIME) || (type == Types.DATE) || (type == Types.TIMESTAMP)) {
             if (Helper.shouldOptimizeDates) {
                 // Optimize dates by avoid conversion to timestamp then back to date or time or util.date.
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/queries/ReportQueryResult.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/queries/ReportQueryResult.java
@@ -277,21 +277,6 @@ public class ReportQueryResult implements Serializable, Map {
                 }
             } else {
                 value = row.getValues().get(itemIndex);
-
-                // Verify that the expected result type matches the actual value type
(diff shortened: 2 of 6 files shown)


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/queries/ReportQueryResult.java
```diff
@@ -277,21 +277,6 @@ protected Object processItem(ReportQuery query, AbstractRecord row, Vector toMan
                 }
             } else {
                 value = row.getValues().get(itemIndex);
-
-                // Verify that the expected result type matches the actual value type
-                if(value != null && query.getSession().getProject().allowResultTypeConversion()) {
```
lukasj, line 282: corresponding property should be removed, method deprecated and this particular call removed

dazey3: I added the property so that behavior will only change IF the property is set. The property defaults to turn this fix OFF. It would break existing customers if you update 2.7 to change behavior. 

Customers may have written their applications to expect Integer, even though Boolean is TECHNICALLY correct. Without this property around the fix, customers WILL BREAK and have to change their application implementation.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
```diff
@@ -1383,6 +1383,9 @@ protected Object getObjectThroughOptimizedDataConversion(ResultSet resultSet, Da
         } else if ((fieldType == ClassConstants.SHORT) || (fieldType == ClassConstants.PSHORT)) {
             value = Short.valueOf(resultSet.getShort(columnNumber));
             isPrimitive = ((Short)value).shortValue() == 0;
+        } else if ((fieldType == ClassConstants.BOOLEAN) || (fieldType == ClassConstants.PBOOLEAN))  {
```
dazey3, line 1386: I think the `allowResultTypeConversion()` property check needs to be included. This property is meant to protect customers who have written their applications to expect the Integer value, instead of the type Boolean. Without this change, production environments could see breaking, ClassCastException changes after updating to a newer version of 2.7.x. I think this is the safest approach to nonfailing, behavior changes that are not dictated by the spec.

dazey3 (changes requested):

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
rfelcman: In latest commit I decided to remove `eclipselink.allow-result-type-conversion` property and related methods due a my and @lukasj fix https://github.com/eclipse-ee4j/eclipselink/pull/983 . As I checked sources, this property and `org.eclipse.persistence.sessions.Project#allowResultTypeConversion()` related method was used only in removed code.
As this property is not included in any release I think, that we can remove it.
If there is any reason why to keep it there let me know.

lukasj (approved):
LGTM now

dazey3:
@lukasj @rfelcman Umm, is it that no one understands the issue being fixed here that my comment to keep the enabling property was completely ignored?

The original code returned Integer. It didnt throw an exception and the JPA specification doesn't dictate that we SHOULD be returning a Boolean. This change is altering behavior, albeit to correct behavior, but it still will result in broken applications that are written to EXPECT Integer. The property was a fail safe for customers to revert/enable the behavior without having to rewrite their applications. Delivering these kinds of behavior changes, on a released version & without a property guard, is reckless and poorly planned.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
dazey3: #983 is MASTER, which I agree is safe to NOT NEED the property as there had not been a release yet. This is 2.7, which has many releases already. Your reasoning for removing the property is flawed

jgrassel: Customers/users using 2.7.x could have written applications which are dependent on the `eclipselink.allow-result-type-conversion` property.  Updates to a released version should not break existing users.  Is there an exceptionally good reason for removing this property which justifies this risk to existing users?  Otherwise, it should remain.

PR #989, opened 2020-12-16
[master] Concurrency manager logging - extension

rfelcman:
There is some minor change is log output produced by cache dead-lock diagnostic. concurrencyManagerCreationDate format is changed to display time with milliseconds and cache key hash code is added.


lukasj (approved):
LGTM

PR #990, opened 2020-12-17
Ant build - catchup ASM upgrade

rfelcman:
This is fix for the promotion. Ant scripts don't reflect current ASM version. There was a limit into < 9.0 version. It leads into not initialized `asm.version` property in build scripts.


lukasj (approved):
LGTM

PR #994, opened 2020-12-24
[2.6_WAS] Fix TCK failure for java.lang.Character type

dazey3:
You may notice that I did not alter the name of the persistence property for this port. This was intentional as the current property name was already released for WebSphere and is now supported. Rather than change the property name and break customers, the property will remain named the same for the 2.6_WAS release.

for #922


jgrassel (approved):
LGTM

jgrassel:
LGTM

PR #995, opened 2020-12-24
[2.7] Fix TCK failure for Character type

dazey3:
I changed the name of the persistence property for this fix to more accurately describe what it is gating. Since the old property was not released in a 2.7 release, I think renaming the property is fine.

for #922

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2020 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2021 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -2598,6 +2598,14 @@ public class PersistenceUnitProperties {
      * Default: "<code>true</code>".
      */
     public static final String ALLOW_NULL_MAX_MIN = "eclipselink.allow-null-max-min";
+
+    /**
+     * The "<code>eclipselink.sql.allow-convert-result-to-boolean</code>" property allows
+     * eclipselink to convert resultset values to boolean types.
+     *
+     * Default: "<code>true</code>".
+     */
+    public static final String ALLOW_CONVERT_RESULT_TO_BOOLEAN = "eclipselink.sql.allow-convert-result-to-boolean";
 
     /**
      * The "<code>eclipselink.id-validation</code>" property defines
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
@@ -1384,8 +1384,10 @@ public class DatabaseAccessor extends DatasourceAccessor {
             value = Short.valueOf(resultSet.getShort(columnNumber));
             isPrimitive = ((Short)value).shortValue() == 0;
         } else if ((fieldType == ClassConstants.BOOLEAN) || (fieldType == ClassConstants.PBOOLEAN))  {
-            value = Boolean.valueOf(resultSet.getBoolean(columnNumber));
-            isPrimitive = ((Boolean)value).booleanValue() == false;
+            if(session.getProject().allowConvertResultToBoolean()) {
+                value = Boolean.valueOf(resultSet.getBoolean(columnNumber));
+                isPrimitive = ((Boolean)value).booleanValue() == false;
+            }
         } else if ((type == Types.TIME) || (type == Types.DATE) || (type == Types.TIMESTAMP)) {
             if (Helper.shouldOptimizeDates) {
                 // Optimize dates by avoid conversion to timestamp then back to date or time or util.date.
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/sessions/Project.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/sessions/Project.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2020 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -192,6 +192,7 @@ public class Project extends CoreProject<ClassDescriptor, Login, DatabaseSession
     protected Collection<String> structConverters;
 
     protected boolean allowNullResultMaxMin = true;
+    protected boolean allowConvertResultToBoolean = true;
 
     /**
      * PUBLIC:
@@ -1316,6 +1317,14 @@ public class Project extends CoreProject<ClassDescriptor, Login, DatabaseSession
 
     /**
      * INTERNAL:
+     * Return true if ResultSet values should be converted for this project.
+     */
+    public boolean allowConvertResultToBoolean() {
+        return this.allowConvertResultToBoolean;
+    }
+
+    /**
+     * INTERNAL:
      * Return true if SQL calls can defer to EOT on this project.
      */
     public boolean allowSQLDeferral() {
@@ -1375,6 +1384,14 @@ public class Project extends CoreProject<ClassDescriptor, Login, DatabaseSession
      */
     public void setAllowNullResultMaxMin(boolean allowNullResultMaxMin) {
         this.allowNullResultMaxMin = allowNullResultMaxMin;
+    }
+
+    /**
+     * INTERNAL:
+     * Set whether ResultSet values should be converted for this project.
+     */
+    public void setAllowConvertResultToBoolean(boolean allowConvertResultToBoolean) {
+        this.allowConvertResultToBoolean = allowConvertResultToBoolean;
     }
 
     /**
--- a/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/property/TestConvertResultToBoolean.java
+++ b/jpa/eclipselink.jpa.test.jse/src/org/eclipse/persistence/jpa/test/property/TestConvertResultToBoolean.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2021 IBM Corporation. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+package org.eclipse.persistence.jpa.test.property;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import java.util.List;
+
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.Query;
+
+import org.eclipse.persistence.config.PersistenceUnitProperties;
+import org.eclipse.persistence.jpa.test.framework.DDLGen;
+import org.eclipse.persistence.jpa.test.framework.Emf;
+import org.eclipse.persistence.jpa.test.framework.EmfRunner;
+import org.eclipse.persistence.jpa.test.framework.Property;
+import org.eclipse.persistence.jpa.test.property.model.GenericEntity;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(EmfRunner.class)
+/**
+ * Tests the 'eclipselink.sql.allow-convert-result-to-boolean' persistence property. 
+ * By enabling this property, the result type of a case select with Booleans should be an integer
+ */
+public class TestConvertResultToBoolean {
+
+    @Emf(createTables = DDLGen.DROP_CREATE, classes = { GenericEntity.class }, 
+            properties = { @Property(name="eclipselink.logging.level", value="FINE"), 
+                    @Property(name=PersistenceUnitProperties.ALLOW_CONVERT_RESULT_TO_BOOLEAN, value="false")})
+    private EntityManagerFactory emf;
+
+    private static boolean POPULATED = false;
+
+    @Test
+    public void testCASE_DisableConvertResultToBoolean() {
+        if (emf == null)
+            return;
+
+        if(!POPULATED) 
+            populate();
+
+        EntityManager em = emf.createEntityManager();
+        try {
+            em.getTransaction().begin();
+
+            Query query = em.createQuery(""
+                    + "SELECT ("
+                       + "CASE "
+                       + "WHEN g.itemInteger1 = 1 THEN TRUE "
+                       + "ELSE FALSE "
+                       + "END "
+                    + ") "
+                    + "FROM GenericEntity g ORDER BY g.itemInteger1 ASC");
+
+            List<?> intList = query.getResultList();
+            assertNotNull(intList);
+            assertEquals(2, intList.size());
+            assertEquals(new Long(1), intList.get(0));
+            assertEquals(new Long(0), intList.get(1));
+        } finally {
+            if (em.getTransaction().isActive()) {
+                em.getTransaction().rollback();
+            }
+            if(em.isOpen()) {
+                em.close();
+            }
+        }
+    }
+
+    private void populate() {
+        EntityManager em = emf.createEntityManager();
+        try {
+
+            GenericEntity tbl1 = new GenericEntity();
+            tbl1.setKeyString("Key01");
+            tbl1.setItemString1("A");
+            tbl1.setItemInteger1(1);
+
+            GenericEntity tbl2 = new GenericEntity();
+            tbl2.setKeyString("Key02");
+            tbl2.setItemString1("B");
+            tbl2.setItemInteger1(2);
+
+            em.getTransaction().begin();
+            em.persist(tbl1);
+            em.persist(tbl2);
+            em.getTransaction().commit();
+
+            POPULATED = true;
+        } finally {
+            if (em.getTransaction().isActive()) {
+                em.getTransaction().rollback();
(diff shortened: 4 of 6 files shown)


jgrassel:
Thank you for restoring the property Will.  For some a behavior that is not well described in the spec, we don't want to break existing customers by unilaterally changing behavior.  There probably are existing users who have been using this release who would find their applications suddenly broken by changing this behavior by default, and they could rightfully open a complain against the product.

Since the spec doesn't clearly define this behavior, you probably should open an issue on https://github.com/eclipse-ee4j/jpa-api to bring this to the attention of the JPA Spec community.

lukasj:
> There probably are existing users who have been using this release who would find their applications suddenly broken by changing this behavior by default, and they could rightfully open a complain against the product.

I'm willing to bet my salary on that no such customer exists since there is NO official 2.7.x EclipseLink release having this property defined.

[2.7.8 version](https://github.com/eclipse-ee4j/eclipselink/blob/2.7.8/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java)
[2.7.7 version](https://github.com/eclipse-ee4j/eclipselink/blob/2.7.7/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java)
[2.7.6 version](https://github.com/eclipse-ee4j/eclipselink/blob/2.7.6/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java)
[2.7.5 version](https://github.com/eclipse-ee4j/eclipselink/blob/2.7.5/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java)

...do you want me to continue? Since I may be looking at wrong places (....we were checking this 4times already), can you show me already released version having this property defined, please?

lukasj (commented):
this is better, The main question to be answered - do we really want to add a property for every bug fix?

jgrassel:
> this is better, The main question to be answered - do we really want to add a property for every bug fix?

Yeah, the previous statement (which I deleted shortly after) I was under the assumption that it had shipped with the prior release (2.7.7).  Will notified me of my mistaken assumption.

The thing is, we don't know if someone out there using Eclipselink has written their application using the original behavior.  The spec isn't clear on the expected output, so they might have just tested to see what Eclipselink does, and wrote their application to use that method.  If they update from an earlier version of 2.7.x to 2.7.8, their application will no longer work.  That's something we should try to avoid.

However, another user had an alternate interpretation (correct me if I'm wrong here, @dazey3) which isn't necessarily wrong as per the spec.  But to provide this alternate interpretation without breaking existing applications, a property is needed to gate the change.  In 2.7.x, if someone really wants this behavior, then they should have to set a property for it, since it changes the nature of an output, rather than say fixing a NullPointerException.  This isn't a new precedent, we added properties like `eclipselink.allow-null-max-min` and `eclipselink.pessimistic.lock.timeout.unit` which preserved existing behavior while enabling the corrected/desired behavior with a configuration property.

Making the new behavior the default shouldn't be an issue with 3.0.0, since that is a major version update, where there should be the expectation for users moving up to that version to fully test their existing application.

lukasj:
> The spec isn't clear on the expected output.

TBH, I'm a bit lost here as I don't know to which part of the spec and particular (test) case you are referring to. To me it looks pretty clear that every typed query created by `em.createQuery("sth", Boolean.class)` must return boolean type and nothing else - [javadoc](https://jakarta.ee/specifications/persistence/3.0/apidocs/jakarta.persistence/jakarta/persistence/entitymanager#createQuery(java.lang.String,java.lang.Class)) says "resultClass - the type of the query result"; in this case the result class is `Boolean.class`, so the return type must be `Boolean` type and not `Integer`.

Granted, it is a change in behaviour from the wrong one, which is easy to workaround (thus not worth to report for many users), to the correct one. What do we want from those affected by this change? Put in a property or remove their workaround? Both changes require a change in the user code.

wrt `eclipselink.allow-null-max-min` and `eclipselink.pessimistic.lock.timeout.unit` - having special handling of null values as well as ability of setting a time unit for time-outs are handy enhancements and there is nothing against that from my side.

jgrassel:
Will explained the problem to me as a CASE statement involving boolean values which the query result was returning an integer value type instead of a boolean value type.  He didn't mention anything about typed queries being involved in the problem he was investigating, so I assume that the situation was calling for an untyped query, and the user was expecting the return value to be boolean instead of an integer based on the SELECT clause.  JPA 2.2 section 4.6.17.4 indeed is mute about the expected output values of a CASE statement, though it is not unreasonable to expect the return type to be the same type as what is used in the CASE expression (if the scalar_expression is a String, then you expect to get a String back).

So for example (using an older Eclipselink that predates Will's fix):

```
@Entity
public class EntityA {
    @GeneratedValue(strategy = GenerationType.AUTO)
    @Id
    private int id;
```

```
    public static void query(HttpServletRequest request, HttpServletResponse response, EntityManager em, UserTransaction tx) throws Exception {
        final PrintWriter pw = response.getWriter();
        em.clear();
        
        String qStr = "SELECT a.id, CASE WHEN MOD(a.id, 2) = 0 THEN TRUE ELSE FALSE END FROM EntityA a";

        pw.println("<br>");
        pw.println("Executing JPQL: &quot;" + qStr + "&quot;");
        pw.println("<br>");
        
        Query q = em.createQuery(qStr);
        List retList = q.getResultList();

        dumpList(retList, em, pw);
    }
```

```
        for (Object m : retList) {
            Object[] row = (Object[]) m;

            pw.println("<tr>");

            // Primary Key
            pw.print("<td><b>");
            pw.print(row[0]);
            pw.println("</b></td>");

            // CASE Data
            pw.print("<td>");
            pw.print(row[1]);
            pw.println("</td>");
            pw.println("</td>");

            pw.println("</tr>");
        }
```

Produces the following output:
```

Id | CASE result
-- | --
1 | 0
3 | 0
4 | 1
5 | 0
7 | 0
8 | 1
9 | 0
11 | 0
12 | 1
13 | 0

```

The above isn't a typed query, just a regular query, with the CASE's output value as part of the select.  The CASE portion of the SELECT defines TRUE or FALSE as its output.  Eclipselink's returned resultset (<2.7.8) defines the CASE's boolean output as an integer and not a boolean (otherwise in the above output we would see true and false, not 1 and 0).  (Because the JPA spec doesn't specifically state what the CASE's output type should be, this isn't a wrong output.  But going purely off the spec, it's not unreasonable to expect a boolean type either.)

Now, users of Eclipselink < 2.7.8 who accepted this type of output are now going to find their applications breaking with 2.7.8 because row[1] is now going to be boolean instead of int (which could cause type casting exceptions.)  Our user base is going to be justifiably unhappy with this, because they expect to be able to update the level of Eclipselink they use without application breaking behavior changes.

Therefore, changing the output type from integer to boolean should be considered an enhancement, and in a micro release increment (2.7.7 to 2.7.8) this behavior change should be put behind a configuration option, so you only get the behavior change if you specifically ask for it.

lukasj:
First of all, I'm not strongly against this property as such. Even though I don't like it, it may be better to have it. Don't take me wrong, I'm just expressing my disagreement :-), while also trying to fully understand the problem and get both/all of us on the same page. Note that I'm using [Jakarta version of the spec](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html) but there's no difference from JPA 2.2 wrt SELECT/CASE.

It's clear that there was (at least one) misunderstanding between us caused by the query we were looking at - simple typed query vs case statement. I think that we can agree on that for my typed query, current behaviour should be the only correct one.

Let's see if we can get on the same page wrt case statement as well. This is how I see it:

We have: `CASE WHEN MOD(a.id, 2) = 0 THEN TRUE ELSE FALSE` where `case expression` is `scalar_expression` and is defined as (based on 4.6.17.4, shortened by substitutions):

```
CASE case_operand WHEN scalar_expression THEN scalar_expression
    ELSE scalar_expression
    END
```

based on that `TRUE` and `FALSE` are `scalar_expression` being evaluated to `boolean_expression`s -> `boolean_literal`s.

For completeness, `scalar_expression` is defined as:

```
scalar_expression::=
    arithmetic_expression | string_expression | enum_expression | datetime_expression |
    boolean_expression | case_expression | entity_type_expression
```

Section 4.6.17 says that _Numeric, string, datetime, case, and entity type expressions result in scalar values._
To me it looks like the defined `case` statement is expected to return scalar value of the `boolean_expression` (= `TRUE`/`FALSE`).

 Let's wrap the `CASE` expression by the `SELECT` statement to have `SELECT CASE WHEN MOD(a.id, 2) = 0 THEN TRUE ELSE FALSE END FROM EntityA a` (as the standalone case statement as such does not make much sense), we should be able to apply section 4.8.1 which, among other things, says: _The result type of a scalar_expression is the type of the scalar value to which the expression evaluates._ Since our `scalar_expression` is the `case_expression` which has been already shown to be evaluated to the `boolean_expression`, the `SELECT` with the `CASE` statement should return the scalar value of the `boolean_expression` (or better - of the appropriate `boolean_literal`).

Did I make any mistake while reading the spec up to this point?

The thing I haven't found to be explicitly defined and/or mentioned in the spec, is the expected java type of the JPQL's scalar value of the `boolean_expression`. So yes, even when I think it is obvious that `boolean_expression` should be evaluated to corresponding Java type for boolean and section 4.3 very likely supports this in this case through _The Jakarta Persistence query language is a typed language, and every expression has a type. The type of an expression is derived from [...] the types of literals._, (we have `boolean_expression` consisting of `boolean_literal`; type of the `boolean_literal` is Boolean - 4.6.1 - therefore our expected type is Boolean) it doesn't seem to be a bad idea to make this explicit in the spec.

Would proposing sth like adding a section _4.6.17.6 Boolean Expressions_` be appropriate? Can you suggest some wording for it, please?  


Now, to understand what EclipseLink is doing and should be doing - on the way to the DB, it translates `TRUE`/`FALSE` literals to corresponding `1`/`0` values since most DBs did not have boolean type at the time code was written, and the problem was that on the way from the DB, `1`/`0` were not translated back to `TRUE`/`FALSE`. Given that some DBs in their recent versions already support boolean data type, we should probably evaluate this and update the code to use `BOOLEAN` DB datatype when possible.

lukasj (commented):
Since we don't want this to be in 3+, it would make sense to mark all new public fields and methods introduced by this change as `Deprecated`

jgrassel:
I'm fine with the deprecation marking.  Are we in agreement with restoring the property that enables the new behavior, then?  Will's waiting for the review result.

dazey3:
@lukasj @jgrassel 
Hello all! I'm a bit late to this conversation as I have been out on vacation. 

> The main question to be answered - do we really want to add a property for every bug fix?

To summarize my position on this, I think there are two classifications for behavior fixes:
1) Specification defined
2) EclipseLink defined

If EclipseLink is violating a spec defined behavior and needs to be corrected, no persistence property is required to gate this behavior. I think we have a strong leg to stand on in this case for complaining customers as EclipseLink is a spec reference implementation. If EclipseLink is blatantly violating the spec in some way, it must be fixed to correct that error and that change should be the default behavior. Otherwise, we may as well add a new property `eclipselink.enable-changes` to simply gate ALL changes and I think we can all agree that is going way too far.
Of course, it gets a bit weedy when EclipseLink was previously working just fine, albeit wrong value. Customers could be taking advantage of bugs or bad/unintended behaviors that aren't necessarily throwing an exception or failing but I still feel comfortable in this case changing behavior to follow the spec by default.

If EclipseLink is behaving in a way not necessarily defined by the spec (provider specific API, gray spec areas), but decidedly/logically incorrect, then a gated property must be used (on non-master releases) / old API becomes deprecated. I think in this case, original behavior must be preserved we are in a much weaker position to make behavior changes with flimsy reasons like "we didn't like what it was doing". 

As to this specific issue (the Integer -> Boolean behavior change):
My original assertion with this change was that it is not necessarily demanded by the specification and therefore fell into category 2. It also has that weird caveat of not being a bug that results in a runtime failure or even look obviously wrong at first glance.

---

> We have: CASE WHEN MOD(a.id, 2) = 0 THEN TRUE ELSE FALSE where case expression is scalar_expression and is defined as (based on 4.6.17.4, shortened by substitutions):
>
>    CASE case_operand WHEN scalar_expression THEN scalar_expression
>    ELSE scalar_expression
>    END
> based on that TRUE and FALSE are scalar_expression being evaluated to boolean_expressions -> boolean_literals.

I agree to this point.

> Section 4.6.17 says that Numeric, string, datetime, case, and entity type expressions result in scalar values. To me it looks like the defined case statement is expected to return scalar value of the boolean_expression (= TRUE/FALSE).

Maybe that's where we disagree, I don't think the specification defines that the case_expression evaluates to the boolean_expression. Does the specification define the specific result **type** of a case expression? 
The specification defines that the result is a scalar **value**, but I don't see where it defines what the type is as it relates to the THEN and ELSE _scalar_expressions_. TRUE and FALSE are scalar values, but so are 1 and 0. It almost seems like this could be provider or database dependent. This may be a gray area in the spec. The footnote of section 4.6.17.4 even says that Case expressions are not supported by all databases and queries are therefore not considered portable. Does each database define the expected function of it's own CASE result?

I'm not trying to be a contrarian just to be argumentative. I agree that Boolean is the correct result because it makes the most sense. It just seems entirely reasonable to me that a customer could currently be getting 1 or 0 integer types with this query, the spec seems absent/vague on the correct type (according to how I read it), and it's safest to gate with a property.

lukasj:
(@dazey3 np, we were all out since mid Dec; btw: HNY!)

>> Section 4.6.17 says that Numeric, string, datetime, case, and entity type expressions result in scalar values. To me it looks like the defined case statement is expected to return scalar value of the boolean_expression (= TRUE/FALSE).

> Maybe that's where we disagree, I don't think the specification defines that the case_expression evaluates to the boolean_expression. 

Correct, it is not defined explicitly.

Given:
* [4.6.1](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#literals) which says _The boolean literals are TRUE and FALSE._
* [4.14](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#bnf) says that only expression allowed to contain boolean literals is the _boolean_expression_:
```
boolean_expression ::=
    state_valued_path_expression | boolean_literal | input_parameter |
    case_expression | function_invocation | (subquery)
```

I believe it is implied that _TRUE_/_FALSE_ are _boolean_expressions_ and nothing else. I do not see any other expression type allowed having _boolean_literals_ defined by the JPQL BNF.

Therefore at this point we should have:
```
CASE case_operand WHEN scalar_expression THEN boolean_expression
    ELSE boolean_expression
END
```
because _boolean_expression_ is _scalar_expression_ ([4.6.17](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#a5243)) and this is how I got to the _To me it looks like the defined case statement is expected to return scalar value of the boolean_expression_ statement.

As for particular **type** - [4.3](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#abstract-schema-types-and-query-domains) only says: _The Jakarta Persistence query language is a typed language, and every expression has a type. The type of an expression is derived from [...] the types of literals._ and depending on the particular context - for *SELECT* we have [4.8.1](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#result-type-of-the-select-clause) saying _The result type of a scalar_expression is the type of the scalar value to which the expression evaluates._. We have nothing about **type** for the _case_expression_ in ie ORDER BY or other applicable places.

> I'm not trying to be a contrarian just to be argumentative.

I really appreciate it. I'm just trying to find the weak place in the spec to improve it. Ie [MySQL](https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#operator_case) defines the return type as _The return type of a CASE expression result is the aggregated type of all result values._, [Derby](https://db.apache.org/derby/docs/10.9/ref/rrefcase.html) has something like _ThenExpression and elseExpression are both expressions that must be type-compatible. For built-in types, this means that the types must be the same or a built-in broadening conversion must exist between the types._

dazey3:
@Lukas

>Therefore at this point we should have:
>    CASE case_operand WHEN scalar_expression THEN boolean_expression
>          ELSE boolean_expression
>    END
> because boolean_expression is scalar_expression

I agree up to this point, I think the BNF makes it quite clear how TRUE/FALSE are _boolean_literal_ -> _boolean_expression_ and _scalar_expression_ -> _boolean_expression_. 

> I'm just trying to find the weak place in the spec to improve it.

To me, what is missing is the connective argument that the THEN/ELSE _scalar_expression_ define the _case_expression_ return type. Perhaps this is a JDBC defined function of CASE, but given that the JPA spec goes out of its way to say "Note that not all databases support the use of SQL case expressions" [4.6.17.4]; it sounds platform dependent. In fact, not all databases even support Booleans (a comment made in database code comments).
The JPA spec only says `... case expressions result in scalar values` (4.6.17). It doesn't say what type the are or define what a CASE select does/returns. Perhaps that is too detailed for our spec and I can agree, but given that not all databases apparently support Case Expressions, someone needs to define it and we need to reference them then.

dazey3:
@lukasj 
Do you agree with this change then and we can get this merged as is? Or are there additional changes you would like me to make?

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/PersistenceUnitProperties.java
```diff
@@ -2599,6 +2599,14 @@
      */
     public static final String ALLOW_NULL_MAX_MIN = "eclipselink.allow-null-max-min";
 
+    /**
+     * The "<code>eclipselink.sql.allow-convert-result-to-boolean</code>" property allows
+     * eclipselink to convert resultset values to boolean types.
+     *
+     * Default: "<code>false</code>".
```
lukasj, line 2606: If the option is being added for backward compatibility (and I believe that is the case here) then the default should be `true` instead

lukasj:
> The JPA spec only says ... case expressions result in scalar values (4.6.17).

Yes, that is what is immediately visible. But then there are few context specific definitions, see [4.8.1](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#result-type-of-the-select-clause) and [4.8.6](https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#a5574)

dazey3:
@lukasj 

> But then there are few context specific definitions, see 4.8.1 and 4.8.6

I believe you are referring to this part of **4.8.1**, yes?
```
The result type of a scalar_expression [case_expression] is the type of the scalar value to which the expression
evaluates. The result type of a numeric scalar_expression is defined in section 4.8.6.
```
**NOTE:**  `scalar_expression::= case_expression`.

Where does the spec define the scalar value to which a case expression evaluates?

----

I think this is the portion of **4.8.6** that is relevant:
```
The result of a case expression, coalesce expression, nullif expression, or arithmetic expression (+, -, *,
/) is determined by applying the following rule to its operands[72].

[72] In the case of a general or simple case expression, these are the scalar expressions of the THEN and ELSE clauses.
```

Footnote [72] is relevant for this scenario since no operand is being specified. However, **4.8.6** really only lists the result type expectations for numeric result types anyway. In this scenario, the THEN and ELSE clauses evaluate to type Boolean. I don't think **4.8.6** specifies the result type for Boolean operand.

dazey3:
@lukasj I changed the default for the persistence property to be "true" rather than "false" per your suggestion. I don't agree with this change if we agree that this is not necessarily defined in the spec... but WebSphere can set the property to override the behavior and protect our customers *shrug*

lukasj (approved):
LGTM

PR #997, opened 2021-01-07
[2.6] Concurrency manager logging - extension (backport from master)

rfelcman:
There is some minor change is log output produced by cache dead-lock diagnostic. concurrencyManagerCreationDate format is changed to display time with milliseconds and cache key and cached object hash code is added. There are some formatting changes too.


lukasj (approved):
LGTM

Issue #999, opened 2021-01-15
Bug 570378: NullPointerException from Embedded Temporal values as MapKeys


dazey3:
Linking to [BZ #570378](https://bugs.eclipse.org/bugs/show_bug.cgi?id=570378)

PR #1000, opened 2021-01-15
[master] Bug 570378: NullPointerException from MapKey Temporals

dazey3:
#999


rfelcman (approved):
LGTM

PR #1007, opened 2021-01-21
[master] Force deferred locks persistence query property

rfelcman:
New `eclipselink.cache.query-force-deferred-locks` property force all queries and relationships
 to use deferred lock strategy during object building and L2 cache population.
 It helps to solve dead lock issues in multithreaded environments where one L2 cache is accessed/populated
 by multiple threads.
 Threads create objects from the same object graph but from different starting point.
 Dead lock issue should happens in large object graphs and entities connected with relationships with EAGER fetch strategy.


lukasj (approved):
LGTM

PR #1010, opened 2021-01-28
[master] Bug/New Feature 559307: EclipseLink on all versions can dead-lock forever - performance fix

rfelcman:
There is small fix for performance improvement when dead-lock detection/logging is enabled.
`eclipselink.concurrency.manager.allow.readlockstacktrace` default value will be false.
Stack trace collection for read locks consumes so much resources.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/PersistenceUnitProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/PersistenceUnitProperties.java
@@ -3948,33 +3948,49 @@ public class PersistenceUnitProperties {
     public static final String CONCURRENCY_MANAGER_MAX_FREQUENCY_DUMP_MASSIVE_MESSAGE  = "eclipselink.concurrency.manager.maxfrequencytodumpmassivemessage";
 
     /**
-     * true - if we want the to fire up an exception to try to get the current thread to release all of its acquired locks and allow other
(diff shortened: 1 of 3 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/PersistenceUnitProperties.java
```diff
@@ -3972,8 +3972,8 @@
     public static final String CONCURRENCY_MANAGER_ALLOW_CONCURRENCY_EXCEPTION  = "eclipselink.concurrency.manager.allow.concurrencyexception";
 
     /**
-     * true - collect debug/trace information during ReadLock acquisition
-     * false - don't collect debug/trace information during ReadLock acquisition
+     * true - collect debug/trace information during ReadLock acquisition. Has negative impact to the performance.
+     * false - (Default) don't collect debug/trace information during ReadLock acquisition
      * Allowed values are: true/false.
```
lukasj, line 3977: I'd prefer to follow the pattern used for other properties (this applies to other places as well):
```
     * <p>
     * <b>Allowed Values</b> (String)<b>:</b>
     * <ul>
     * <li>"<code>false</code>" (DEFAULT)
     * <li>"<code>true</code>"
     * </ul>
```

lukasj (approved):
I have a small problem with javadoc format but other than that, LGTM

PR #1011, opened 2021-01-28
[2.7] Bug/New Feature 559307: EclipseLink on all versions can dead-lock forever - performance fix

rfelcman:
There is small fix for performance improvement when dead-lock detection/logging is enabled.
"eclipselink.concurrency.manager.allow.readlockstacktrace" default value will be false.
Stack trace collection for read locks consumes so much resources.


lukasj (approved):
I have a small problem with javadoc format but other than that, LGTM

Issue #866, opened 2020-07-27
EclipseLink DB2Platform no longer binds Literals as Parameters (DB2Platform.java#L74 => "this.shouldBindLiterals = false;")


hubmat00:
EclipseLink DB2Platform no longer binds Literals as Parameters (DB2Platform.java#L74 => this.shouldBindLiterals = false;)

We are using EclipseLink packaged within Websphere Liberty Profile (WLP) version 20.0.0.x in combination with DB2. WLP 20.0.0.x packages EclipseLink in the version 2.7.6.

While updating from WLP version 19.0.0.x to 20.0.0.x we experience problems with the new provided EclipseLink version 2.7.6. As changed in the DB2Platform with version 2.7.6 (https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/DB2Platform.java#L74) literals get no longer bound as parameters in the query when using Criteria Builder, literals are bound inline in the SQL-statements. **_This behaviour leads to changed SQL-statements and restricts DB2 to NO longer use prepared statements with parameters._**

Changes that force DB2Platform to "this.shouldBindLiterals = false" were implemented with:
* EclipseLink Issue: https://github.com/eclipse-ee4j/eclipselink/issues/492
* EclipseLink Commit: https://github.com/eclipse-ee4j/eclipselink/commit/418a2e5506c36854e72a5fd5662cdff4ac1cb80e
* IBM related issue:
https://www.ibm.com/support/pages/apar/PH06475

There exists another IBM related Issue (https://www.ibm.com/support/pages/apar/PH14747), but I am not sure if this IBM issue relates to my issue.

In general and independent of known workarounds, **we ask for a global solution that allows to set the property "shouldBindLiterals" via system properties** (similar to "eclipselink.jdbc.force-bind-parameters" and "eclipselink.jdbc.bind-parameters" => https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java#L2828 and L#2833). We are also open for other global solutions or fixes that allow us to bind literals as parameters!

### For WLP Applications with DB2 persistency we found some workarounds:
1) **Parameter-Expressions:** setting parameters as Parameter-Expressions => every query based on Criteria API needs to be adapted, not practicable for large applications with thousands of queries
```java
[...]

PR #1013, opened 2021-02-08
[master] Add 'eclipselink.target-database-properties' as a System Property

dazey3:
for #866


lukasj (approved):
LGTM

Issue #1016, opened 2021-02-12
 EclipseLink doesn't generate SQL UPDATE correctly if @ReturnUpdate is used in InheritanceType.JOINED


rfelcman:
SQL UPDATE command is not correctly generated if `@ReturnUpdate` is used in `InheritanceType.JOINED`. It started from 2.7.7. It is related with fix for #709 .
Test case to simulate it:
DB Tables:
```
CREATE TABLE JPA22_RETURNINSERTUPDATE_MASTER_JOINED  (
    ID     NUMBER(15) PRIMARY KEY ,
    TYPE   VARCHAR2(50) NOT NULL
  );

CREATE TABLE JPA22_RETURNINSERTUPDATE_DETAIL_JOINED  (
    ID                  NUMBER(15) PRIMARY KEY ,
    DETAIL_NR           NUMBER(15),
    DETAIL_NR_VIRTUAL   NUMBER(15) AS ( DETAIL_NR * 10 ) VIRTUAL
);

```

There are two entities:
```
...
@Entity
@Table(name = "JPA22_RETURNINSERTUPDATE_MASTER_JOINED")
@Inheritance(strategy = InheritanceType.JOINED)
@DiscriminatorColumn(name = "type", discriminatorType = DiscriminatorType.STRING)
@SequenceGenerator(name = "AbstractEntityObjectSEQ_STORE", sequenceName = "TEST_RETURNINSERT_MASTER_JOINED_ID_SEQ", allocationSize = 1)
public class ReturnInsertUpdateMasterJoined implements Serializable {
...
```
```
...
@Entity
@Table(name = "JPA22_RETURNINSERTUPDATE_DETAIL_JOINED")
@PrimaryKeyJoinColumn(name = "id")
@DiscriminatorValue("A")
public class ReturnInsertUpdateDetailJoined extends ReturnInsertUpdateMasterJoined {
...
```
Fields in ReturnInsertUpdateDetailJoined
```
@Column(name = "detail_nr")
Long detailNumber = null;

@Column(name = "detail_nr_virtual", nullable = false, unique = true, updatable = false)
@ReturnInsert(returnOnly = true)
@ReturnUpdate
Long detailNumberVirtual = null;
```

There are two possible errors:
1) For field update in JPA22_RETURNINSERTUPDATE_DETAIL_JOINED table only
```
...
[...]

PR #1017, opened 2021-02-15
EclipseLink doesn't generate SQL UPDATE correctly if @ReturnUpdate is used in InheritanceType.JOINED

rfelcman:
Bugfix for #1016 + unit test

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2019 IBM and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
```diff
@@ -835,7 +835,16 @@ protected SQLUpdateStatement buildUpdateStatement(DatabaseTable table) {
         updateStatement.setModifyRow(getModifyRow());
         updateStatement.setTranslationRow(getTranslationRow());
         if (getDescriptor().hasReturningPolicies() && getDescriptor().getReturnFieldsToGenerateUpdate() != null) {
-            updateStatement.setReturnFields(getDescriptor().getReturnFieldsToGenerateUpdate());
+            // In case of RelationalDescriptor only return fields for current table must be used.
+            Vector<DatabaseField> returnFieldsForTable = new NonSynchronizedVector();
```
lukasj, line 839: Would it be possible to start using common `List` (`ArrayList`) where possible instead? I can see no reason for deprecated `Vector` here...

rfelcman: OK replaced by `List` (`ArrayList`)

lukasj (approved):
LGTM

PR #1019, opened 2021-02-16
[2.6] EclipseLink doesn't generate SQL UPDATE correctly if @ReturnUpdate is used in InheritanceType.JOINED - backport from master

rfelcman:
Bugfix for #1016 + unit test


lukasj (approved):
LGTM

PR #1022, opened 2021-02-19
#1021: jaxb-compiler command fails to find jaxb implementation

### LINKED ISSUE #1021: [moxy] jaxb-compiler command fails to find jaxb implementation
running `./jaxb-compiler.sh ../xsds/beanval.xsd ` fails with:

```
Exception in thread "main" java.lang.AssertionError: jakarta.xml.bind.JAXBException: Implementation of Jakarta XML Binding-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: org.glassfish.jaxb.runtime.v2.ContextFactory]
	at com.sun.tools.xjc.reader.xmlschema.bindinfo.BindInfo.getCustomizationContext(BindInfo.java:326)
...
Caused by: jakarta.xml.bind.JAXBException: Implementation of Jakarta XML Binding-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: org.glassfish.jaxb.runtime.v2.ContextFactory]
	at jakarta.xml.bind.ContextFinder.newInstance(ContextFinder.java:255)
	at jakarta.xml.bind.ContextFinder.newInstance(ContextFinder.java:243)
	at jakarta.xml.bind.ContextFinder.find(ContextFinder.java:398)
	at jakarta.xml.bind.JAXBContext.newInstance(JAXBContext.java:691)
	at jakarta.xml.bind.JAXBContext.newInstance(JAXBContext.java:632)
	at com.sun.tools.xjc.reader.xmlschema.bindinfo.BindInfo.getCustomizationContext(BindInfo.java:306)
	... 62 more
Caused by: java.lang.ClassNotFoundException: org.glassfish.jaxb.runtime.v2.ContextFactory
...
```

the problem here is that XJC contains binding classes representing customization schema. When it tries to initialize them through the `JAXBContext.newInstance(Class...)` API call, no xml binding implementation is found because XJC as such does not depend on the specific runtime - it should just find MOXy and use that.

Possibilities:
* add definition to `META-INF/services` - if it is added, then it is likely that this entry would hide or compete with the default/other implementation and we don't want that. We want to allow both implementations to co-exist ie in the server environment
* add `jaxb.properties` to `com.sun.tools.xjc.reader.xmlschema.bindinfo` package - a side effect of this can be that even default/other implementation could read the model through MOXy - again, something we probably don't want
* pass `jakarta.xml.bind.JAXBContextFactory` system option to XJC through the script - probably the best option; the command itself is there for the use from command line, so it is unlikely it will interfere with others when run

lukasj:
fixes #1021

### DIFF
--- a/bundles/eclipselink/src/main/scripts/bin/jaxb-compiler.cmd
+++ b/bundles/eclipselink/src/main/scripts/bin/jaxb-compiler.cmd
@@ -1,5 +1,5 @@
 @REM
-@REM Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
+@REM Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
 @REM
 @REM This program and the accompanying materials are made available under the
 @REM terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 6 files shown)


rfelcman:
LGTM

rfelcman (approved):

PR #1025, opened 2021-03-02
[2.7] Bug 535088 - [test] SubSelectFromClauseTest2 failing randomly on public infra 2.0

rfelcman:
I think, that behind this test failure is some race condition with another test. It seems, that due a parallel test execution there are values in the DB tables which leads, that query in the test returns at least one row. But this test is focused on `ExpressionBuilder` syntax and doesn't expect any rows from DB. Additional condition/criteria ensures, that nothing is selected.


lukasj (approved):
LGTM

PR #1031, opened 2021-03-04
[2.7] Force deferred locks persistence query property - backport from master

rfelcman:
New `eclipselink.cache.query-force-deferred-locks` property force all queries and relationships
to use deferred lock strategy during object building and L2 cache population.
It helps to solve dead lock issues in multithreaded environments where one L2 cache is accessed/populated
by multiple threads.
Threads create objects from the same object graph but from different starting point.
Dead lock issue should happens in large object graphs and entities connected with relationships with EAGER fetch strategy.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java
@@ -1467,6 +1467,12 @@ public class EntityManagerSetupImpl implements MetadataRefreshListener {
         if(typeMap.isEmpty() && sizeMap.isEmpty() && sharedMap.isEmpty()) {
             return;
         }
+        String queryCacheForceDeferredLocks = getConfigPropertyAsStringLogDebug(PersistenceUnitProperties.CACHE_QUERY_FORCE_DEFERRED_LOCKS, m, session);
+        if ((queryCacheForceDeferredLocks != null) && queryCacheForceDeferredLocks.equalsIgnoreCase("true")) {
+            session.getProject().setQueryCacheForceDeferredLocks(true);
+        } else {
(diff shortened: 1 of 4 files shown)


jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java
```diff
@@ -1467,6 +1467,12 @@ protected void updateDescriptorCacheSettings(Map m, ClassLoader loader) {
         if(typeMap.isEmpty() && sizeMap.isEmpty() && sharedMap.isEmpty()) {
             return;
         }
+        String queryCacheForceDeferredLocks = getConfigPropertyAsStringLogDebug(PersistenceUnitProperties.CACHE_QUERY_FORCE_DEFERRED_LOCKS, m, session);
+        if ((queryCacheForceDeferredLocks != null) && queryCacheForceDeferredLocks.equalsIgnoreCase("true")) {
+            session.getProject().setQueryCacheForceDeferredLocks(true);
+        } else {
+            session.getProject().setQueryCacheForceDeferredLocks(false);
+        }
```
lukasj, line 1475: this could be collapsed into `session.getProject().setQueryCacheForceDeferredLocks(queryCacheForceDeferredLocks != null && "true".equalsIgnoreCase(queryCacheForceDeferredLocks));`

lukasj (approved):
LGTM

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java
lukasj: or even `session.getProject().setQueryCacheForceDeferredLocks("true".equalsIgnoreCase(queryCacheForceDeferredLocks));`

lukasj (approved):

Issue #1034, opened 2021-03-09
Bug 571794: Enable Input parameters for Criteria API SELECT clause


dazey3:
Linking to [BZ #571794](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571794)

I noticed a difference in behavior between JPQL and CriteriaBuilder for the allowance of input parameters in the SELECT clause. I'm not sure if this is a bug or a feature, but the code and JPQL behavior implies to me it's a bug even though the JPA Spec seems to indicate this shouldn't work:

```
4.6.4 Input Parameters

    Input parameters can only be used in the WHERE clause or HAVING clause of a query or as the new value for an update item in the SET clause of an update statement.
```

JPQL Example:
```
    Query query = em.createQuery("SELECT s.strVal1, ?1 FROM SimpleEntity s WHERE s.intVal1 = ?2");
    query.setParameter(1, "HELLO");
    query.setParameter(2, 4);
    query.getResultList();
```

This example JPQL does NOT throw an error, instead it generates this SQL query (Oracle, MySQL, Derby, DB2) and runs fine:
```
SELECT STRVAL1, ? FROM SIMPLEENTITY WHERE (INTVAL1 = ?)
	bind => [HELLO, 4]
```

Criteria Builder API Example:
```
    CriteriaBuilder cb = em.getCriteriaBuilder();
    cquery = cb.createQuery(Object[].class);
    root = cquery.from(SimpleEntity.class);
    cquery.multiselect(root.get(SimpleEntity_.strVal1), cb.parameter(String.class, "stringValue"));
    cquery.where(cb.equal(root.get(SimpleEntity_.intVal1), cb.parameter(Long.class, "idValue")));

    query = em.createQuery(cquery);
    query.setParameter("stringValue", "HELLO");
    query.setParameter("idValue", 4);
    query.getResultList();
```

Running this equivalent Criteria API example DOES throw this exception:
```
Exception [EclipseLink-6094] (Eclipse Persistence Services - 2.7.9.qualifier): org.eclipse.persistence.exceptions.QueryException
Exception Description: The parameter name [stringValue] in the query's selection criteria does not match any parameter name defined in the query.
Query: ReportQuery(referenceClass=SimpleEntity sql="SELECT STRVAL1, ? FROM SIMPLEENTITY WHERE (INTVAL1 = ?)")
	at org.eclipse.persistence.exceptions.QueryException.parameterNameMismatch(QueryException.java:1098)
[...]

PR #1037, opened 2021-03-09
[master] Bug 571794: Enable Input parameters for Criteria API SELECT clause

dazey3:
for #1034

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/criteria/TestCriteriaBuilder.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/criteria/TestCriteriaBuilder.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2018, 2019 IBM Corporation. All rights reserved.
+ * Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -30,6 +30,8 @@ import jakarta.persistence.criteria.Join;
 import jakarta.persistence.criteria.Root;
 
 import org.eclipse.persistence.internal.jpa.EntityManagerFactoryImpl;
+import org.eclipse.persistence.jpa.test.criteria.model.CriteriaCar;
+import org.eclipse.persistence.jpa.test.criteria.model.CriteriaCar_;
 import org.eclipse.persistence.jpa.test.criteria.model.L1;
 import org.eclipse.persistence.jpa.test.criteria.model.L1Model;
 import org.eclipse.persistence.jpa.test.criteria.model.L1_;
@@ -47,7 +49,7 @@ import org.junit.runner.RunWith;
 @RunWith(EmfRunner.class)
 public class TestCriteriaBuilder {
 
-    @Emf(createTables = DDLGen.DROP_CREATE, classes = { L1.class, L2.class })
+    @Emf(createTables = DDLGen.DROP_CREATE, classes = { L1.class, L2.class, CriteriaCar.class })
     private EntityManagerFactory emf;
 
     /**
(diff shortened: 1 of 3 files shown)


dazey3:
I have this PR as a draft so that it doesn't get merged without the reviewer reading https://bugs.eclipse.org/bugs/show_bug.cgi?id=571794 first. There may be additional changes that should be brought up in the JPA Spec git if this change is accepted.

lukasj (commented):
@dazey3 The most important question - do you need this, if approved, in 3.0.1 or 2.7.x/2.6.x are more important and master can wait?

In any case, let me paste here your statements from the bug and inline my comments for better readability:

> I'm not sure if this is a bug or a feature, but the code and JPQL behavior implies to me it's a bug even though the JPA Spec seems to indicate this shouldn't work:

As I understand it the spec does not indicate it should nor it should not work. As you pointed out, following is there in the spec:

```
4.6.4 Input Parameters

    Input parameters can only be used in the WHERE clause or HAVING clause of a query or as the new value for an update item in the SET clause of an update statement.
```

Per my understanding, all it says is that in given context, input parameters MUST be supported. It says nothing about what SHOULD or SHOULD NOT happen in other cases, ie the one you're dealing with, so based on that, I believe that the behaviour of input parameter in the SELECT clause is undefined by the spec and thus - from that perspective - current behaviour in both cases is compliant with the spec itself. (There are few more restrictions on input parameters through the spec but I believe none of them is applicable to this particular context)

> TLDR, I have two examples: a JPQL and a Criteria API example. Both are equivalent, but the JPQL runs just fine and the Criteria API fails.

...one is likely a bug, the other a feature. But which one is right and which one is wrong? :-)


> The JPA spec seems to say that neither are valid and it should error!

I do not agree with you here, I think the spec is silent on what exactly should happen in this particular case (as mentioned above)

> My opinion is that since it works in JPQL, keeping Criteria API in sync is more important.

Agree!

> If we want to adhere to the JPA Spec, then we should alter JPQL to throw an exception and I am not in favor of removing this functionality given that databases support it. If anything, the JPA spec should expand to include input parameters in SELECT clause. Thoughts?
 
I would probably not change the spec here and keep it open for interpretation unless there is broader agreement across other vendors that this needs to be addressed.

lukasj:
btw: to make it clear - I'm for approving the change `as is`

dazey3:
@lukasj 
Thanks for looking at this!

> The most important question - do you need this, if approved, in 3.0.1 or 2.7.x/2.6.x are more important and master can wait?

It can wait. 

> As I understand it the spec does not indicate it should, nor it should not work.
> Per my understanding, all it says is that in given context, input parameters MUST be supported.

```
4.6.4 Input Parameters

    Input parameters can only be used in the WHERE clause or HAVING clause of a query or as the new value for an update item in the SET clause of an update statement.
```

I disagree with that interpretation of 4.6.4. It isn't just saying that input parameters are supported in WHERE / HAVING clauses, to me, it's saying the JPA spec ONLY supports the use of input parameters in WHERE / HAVING clauses. That's a pretty big distinction and was the root of my dilemma with just "fixing" this without saying anything. If you "can only do X", then I think that prohibits you from doing Y if Y does not exist as part of X. 

> I would probably not change the spec here and keep it open for interpretation unless there is broader agreement across other vendors that this needs to be addressed.

I'm generally not a fan of keeping things vague when it can be clarified. I have seen customers desired use of parameters in SELECT statements (stand alone, in a constructor, or a case expression) and it seems like this is simple enough to expand the JPA Spec to say "Only in SELECT, WHERE, or HAVING clauses"

lukasj:
>  it's saying the JPA spec ONLY supports the use of input parameters in WHERE / HAVING clauses.

Right. But the spec does not explicitly say, what should happen if input parameters are used on other places. I think it should be read as "what is not forbidden, is allowed" - providers are free to either support such usage or fail and they should be consistent in the behaviour they choose (which was not the case before this PR)

> I'm generally not a fan of keeping things vague when it can be clarified.

the best thing to do is to file an issue in the jpa-api project and deal with it there. I am by no means putting myself into the role of the only decision maker ;-)

dazey3:
Ok, I guess we both read it differently. To me it hinges on the word "only" and to me, the use of that word means it prohibits the use outside of WHERE and HAVING. I agree that the spec does not expressly define what happens in the case where an input parameter is using outside of WHERE or HAVING clauses, but I would think that is because outside of WHERE and HAVING is invalid/prohibited. The spec doesnt describe specific error messages much.

I also created a PR for 3.0 and 2.7. If you want to approve them, I am good with merging. I think we both agree the change is good, even if you feel it is EclipseLink specific behavior. I just wanted to make sure it was not in direct opposition to the spec.

lukasj (approved):
LGTM

PR #1038, opened 2021-03-10
[master] ConcurrencyManager dead-lock detection diagnostic improvement 2.0

rfelcman:
This is extension to current dead-lock detection diagnostic.
Major updates are:

- New system/persistence property `eclipselink.concurrency.manager.build.object.complete.waittime` to control how much time ConcurrencyManager will wait before it will identify, that thread which builds new object/entity instance should be identified as a potential dead lock source and log info.
- Add to current dead-lock detection log output method name (from ConcurrencyManager) where dead-lock should happen
- `org.eclipse.persistence.internal.helper.ConcurrencyManager#isBuildObjectOnThreadComplete` minor changes to log some dead-lock diagnostic output


lukasj (approved):
LGTM

PR #1049, opened 2021-03-25
[2.7] ConcurrencyManager dead-lock detection diagnostic improvement (+ Semaphores) 2.0 - backport from master

rfelcman:
This backport of PRs #1038 and #1041

This is extension to current dead-lock detection diagnostic.
Major updates are:

- New system/persistence property `eclipselink.concurrency.manager.build.object.complete.waittime` to control how much time ConcurrencyManager will wait before it will identify, that thread which builds new object/entity instance should be identified as a potential dead lock source and log info.
- Addition to current dead-lock detection log output with method name (from ConcurrencyManager) where dead-lock should happen
- `org.eclipse.persistence.internal.helper.ConcurrencyManager#isBuildObjectOnThreadComplete` minor changes to log some dead-lock diagnostic output
- Implementation of  `org.eclipse.persistence.internal.helper.ConcurrencySemaphore` which should be used to control/limit how many threads should execute selected code.
In this case is `ConcurrencySemaphore` used to limit parallel access to `org.eclipse.persistence.internal.descriptors.ObjectBuilder` and `org.eclipse.persistence.internal.helper.WriteLockManager` to prevent possible dead-lock issues there.
If ConcurrencySemaphore will be active, there should be negative impact to the performance.
By default ConcurrencySemaphore is disabled and must be enabled by persistence or system properties.


lukasj (approved):
LGTM

Issue #1051, opened 2021-03-26
remove org.eclipse.persistence.asm from the main build


lukasj:
The idea is to allow different project lifecycle for the `org.eclipse.persistence.asm` module from the one used by eclipselink itself.

Motivation:
 with current JDK release cadence, it is hard to catch up and provide new EclipseLink release everytime there is new JDK release just to accommodate new ASM to support it but in most cases, eclipselink as such is able to work with new ASM `as is`. There are rare cases when there are changes in eclipselink code required.

Proposed solution:
* remove `org.eclipse.persistence.asm` from `eclipselink.jar` all-in-one artifact
* create new release job(s) for the `org.eclipse.persistence.asm` artifact
* treat `org.eclipse.persistence.asm` as a regular project dependency (like ie `oracleddlparser`)
* relax OSGi imports for repackaged asm packages (basically remove upper boundary from the version range)
* let `eclipselink.jar` depend on the `org.eclipse.persistence.asm` artifact

Proposed Usage:
* default case: particular EclipseLink release depends on/uses latest available version of the `org.eclipse.persistence.asm` artifact
* usage on newly released JDK: new version of `org.eclipse.persistence.asm` is released. Users can explicitly define dependency on the new version of the `org.eclipse.persistence.asm` artifact (in maven) or put it on the classpath (ant, command-line) and keep using "old" version of EclipseLink till the new one is made available.

Probably key question, if we go this way, is where to put the source-code for the `org.eclipse.persistence.asm` artifact. I see 2 options:
1. keep in the main repository - we can create a job and keep going ourselves - this pattern is used ie in [jaxb-ri](https://github.com/eclipse-ee4j/jaxb-ri/tree/master/jaxb-ri) where some sub projects can be released when needed (ie codemodel or xsom), disadvantage can be seen in number of tags and tag naming
2. move it out to separate repository ie `eclipselink-asm` - this would require involvement of eclipse.org people who we would need to ask to set up the repo for us 

@dazey3 @jgrassel @rfelcman thoughts? This can be seen as incompatible change in service release wrt content of `eclipselink.jar` but I think this is something we may want to have in 2.7.9, not sure about 3.0.1 (it could probably wait for the release after it)

jgrassel:
[...]

PR #1054, opened 2021-03-30
Asm

lukasj:
relates to #1051 in master:
* removes eclipselink asm from the main build
* main build treats eclipselink asm as any other other library

### DIFF
--- a/plugins/org.eclipse.persistence.asm/pom.xml
+++ b/plugins/org.eclipse.persistence.asm/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2019, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0 which is available at
--- a/plugins/org.eclipse.persistence.asm/pom.xml
+++ b/plugins/org.eclipse.persistence.asm/pom.xml
@@ -27,7 +27,7 @@
 
     <groupId>org.eclipse.persistence</groupId>
     <artifactId>org.eclipse.persistence.asm</artifactId>
-    <version>3.0.1-SNAPSHOT</version>
+    <version>9.1-SNAPSHOT</version>
 
     <name>EclipseLink ASM</name>
     <description>Comprehensive and universal persistence framework for Java.</description>
--- a/plugins/org.eclipse.persistence.asm/pom.xml
+++ b/plugins/org.eclipse.persistence.asm/pom.xml
@@ -18,17 +18,105 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <modelVersion>4.0.0</modelVersion>
 
-    <name>EclipseLink ASM</name>
+    <parent>
+        <groupId>org.eclipse.ee4j</groupId>
+        <artifactId>project</artifactId>
+        <version>1.0.6</version>
+        <relativePath/>
+    </parent>
+
     <groupId>org.eclipse.persistence</groupId>
     <artifactId>org.eclipse.persistence.asm</artifactId>
-    <packaging>jar</packaging>
+    <version>3.0.1-SNAPSHOT</version>
 
-    <parent>
-        <groupId>org.eclipse.persistence</groupId>
-        <artifactId>org.eclipse.persistence.parent</artifactId>
-        <version>3.1.0-SNAPSHOT</version>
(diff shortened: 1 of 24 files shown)


rfelcman:
Just a few comments:
Travis-CI fix: maybe add to `.travis.yml` before line 79 ASM build to populate Maven local repo.

plugins/org.eclipse.persistence.asm/pom.xml
```diff
@@ -110,6 +253,63 @@
...
+                            <!-- ASM itself uses 1.8 source level and then retrofits the bytecode
+                            to work on Java SE 5+. We don't use retrofitter, so we need
+                            to fix sources to compile for Java SE 7 (EclipseLink 2.6.x) -->
+                            <target>
+                                <replace file="${project.build.directory}/generated-sources/asm-src/org/eclipse/persistence/internal/libraries/asm/tree/Util.java">
+                                    <replacefilter>
+                                        <replacetoken><![CDATA[new ArrayList<>(1)]]></replacetoken>
+                                        <replacevalue><![CDATA[new ArrayList<T>(1)]]></replacevalue>
+                                    </replacefilter>
+                                </replace>
```
rfelcman, line 276: Interesting I believe the best solution from possible options, but I think, that should be applied to all *.java files in `generated-sources` to cover possible ASM code changes.

lukasj: that would have 2 drawbacks:
* it's faster to change 1 file than XX files
* it's probably better to "track" necessary changes, git won't do it for us here

plugins/org.eclipse.persistence.asm/pom.xml
```diff
@@ -18,17 +18,106 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <modelVersion>4.0.0</modelVersion>
 
-    <name>EclipseLink ASM</name>
+    <parent>
```
rfelcman, line 21: Line 4: Copyright Year

pom.xml
```diff
@@ -161,6 +161,8 @@
         <test.xml.parser>org.eclipse.persistence.platform.xml.jaxp.JAXPParser</test.xml.parser>
 
         <!--Eclipse Dependencies version-->
+        <!-- TODO: FIX ME before merge-->
```
rfelcman, line 164: TODO fix?

plugins/org.eclipse.persistence.asm/src/main/java/org/eclipse/persistence/internal/libraries/asm/EclipseLinkASMClassWriter.java
```diff
@@ -0,0 +1,123 @@
...
+                // ie SecurityException
+                LOG.log(Level.WARNING, "Cannot read 'java.specification.version' property.", t);
+                if (LOG.isLoggable(Level.FINE)) {
+                    LOG.log(Level.FINE, "Generating bytecode for Java SE ''{0}''.", v);
+                }
+            }
+        }
+        Integer version = versionMap.get(v);
+        if (version == null) {
+            // current JDK is iether too new
```
rfelcman, line 103: iether -> either

rfelcman (changes requested):
Almost OK except mentioned minor comments.
Two things which should be optionally included there:

- New Travis-CI target something like

```
...
  matrix:
    - MAVEN_TEST=asm
...
  - if [ "$MAVEN_TEST" == "asm" ]; then
      cd plugins/org.eclipse.persistence.asm/;
      mvn test;
    fi
```

- Maybe make generated Javadoc little bit nicer (title, copyright year)

plugins/org.eclipse.persistence.asm/src/main/java/org/eclipse/persistence/internal/libraries/asm/EclipseLinkASMClassWriter.java
lukasj: done

plugins/org.eclipse.persistence.asm/pom.xml
lukasj: done

pom.xml
lukasj: this is intentional to trigger extra build/tests for "final" asm build once it's ready in staging to double check everything is fine

lukasj:
wrt travis for asm - it is non-trivial to add it correctly - the one built by travis won't be the one used by main eclispelink build there without extra modification of eclipselink's pom - I don't think we want to be doing that

wrt javadoc - I can see correct cp year in the javadoc footer; I don't think we need to customize the default more

rfelcman (approved):
LGTM

PR #1055, opened 2021-04-01
[2.7] use o.e.p.asm from maven central

lukasj:
relates to #1051 in 2.7 branch:

after this change, one should be able to backport asm from master by changing one line


buildsystem/compdeps/pom.xml
```diff
@@ -57,6 +62,7 @@
         <eclipse.drop>2018-12</eclipse.drop>
 
         <!-- DEPENDENCY Versions -->
+        <eclipselink.asm.version>9.1.0</eclipselink.asm.version>
```
rfelcman, line 65: Maybe add there as a comment above CQ number.

rfelcman (approved):
LGTM (just one minor comment)

buildsystem/compdeps/pom.xml
lukasj: this is to be tracked at the library level, not here

Issue #1058, opened 2021-04-08
Bug 570702: Add support for Embeddable fields as JOIN targets


dazey3:
Linking to [BZ #570702](https://bugs.eclipse.org/bugs/show_bug.cgi?id=570702)

The JPA Spec, 4.4.4, defines two queries that it states are "equivalent":
```
In the following example, contactInfo denotes an embeddable class consisting of an address and set of phones. Phone is an entity.

SELECT p.vendor FROM Employee e JOIN e.contactInfo.phones p WHERE e.contactInfo.address.zipcode = '95054'

The following query is equivalent to the query above:

SELECT p.vendor FROM Employee e JOIN e.contactInfo c JOIN c.phones p WHERE e.contactInfo.address.zipcode = '95054'
```

First of all, the spec doesn't specifically define the Employee, ContactInfo, Address, and Phone classes for these queries. There are defined classes in other places of the JPA spec for Employee, ContactInfo, & Address but not for Phone. However, those defined Employee, ContactInfo, & Address classes do not match the attributes defined in this query. For this reason, I am using the following ORM classes that at least follow the definition described in 4.4.4 and 4.4.5.1

```
    @Entity
    public class Employee {
        @Id int id;
        @Embedded private ContactInfo contactInfo;
    }

    @Entity
    public class Phone {
        @Id private int id;
        private String vendor;
    }

    @Embeddable
    public class ContactInfo {
        @Embedded private Address address;
        @ManyToMany private List<Phone> phones;
    }

    @Embeddable
    public class Address {
        private String street;
        private String city;
        private String state;
        private String zipcode;
    }
```

Running both of these "equivalent" queries against EclipseLink, the second query fails with the following exception:
```
Caused by: Exception [EclipseLink-0] (Eclipse Persistence Services - 2.6.8.WAS): org.eclipse.persistence.exceptions.JPQLException
Exception Description: Problem compiling [SELECT e FROM Employee e JOIN e.contactInfo c JOIN c.phones p]. 
[30, 43] The collection-valued path 'e.contactInfo' must resolve to an association field.
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:155)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.validate(HermesParser.java:347)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.populateQueryImp(HermesParser.java:278)
[...]

PR #1059, opened 2021-04-08
[master] Bug 570702: Add support for Embeddable fields as JOIN targets

dazey3:
#1058

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -376,6 +377,11 @@ final class EclipseLinkSemanticValidatorHelper implements SemanticValidatorHelpe
         catch (ClassCastException e) {
             return ((QueryKey) mapping).isCollectionQueryKey();
         }
+    }
+
+    @Override
+    public boolean isEmbeddableMapping(Object mapping) {
+        return (mapping instanceof org.eclipse.persistence.mappings.EmbeddableMapping);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
```diff
@@ -378,6 +379,19 @@ public boolean isCollectionMapping(Object mapping) {
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isEmbeddableMapping(Object mapping) {
+
+        if (mapping == null) {
+            return false;
```
lukasj, line 389: this nulll check can be removed since `null instanceof org.eclipse.persistence.mappings.EmbeddableMapping` is  `false`

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
```diff
@@ -378,6 +379,19 @@ public boolean isCollectionMapping(Object mapping) {
         }
     }
 
+    /**
+     * {@inheritDoc}
```
lukasj, line 383: I'd probably just rely on the `@Override` - can be a matter of taste

jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/tools/GenericSemanticValidatorHelper.java
```diff
@@ -388,6 +389,14 @@ public boolean isCollectionMapping(Object mapping) {
         return (mapping != null) && ((IMapping) mapping).isCollection();
     }
 
+    /**
+     * {@inheritDoc}
```
lukasj, line 393: consider removing and relying on "default" override javadoc (applies to few more places)

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
dazey3: Sounds good, removed

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/jpa/jpql/EclipseLinkSemanticValidatorHelper.java
dazey3: Makes sense. Removed.

jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/tools/GenericSemanticValidatorHelper.java
dazey3: There are a lot of places where @Override + `{@inheritDoc}` is being used and I was just keeping everything uniform. I would have considered making changes to all those places since I agree that @Override should be sufficient, but didn't want to widen the changes here. Maybe a large refactoring can be done to clean up the rest.

lukasj: feel free to cleanup remaining `{@inheritDoc}` through the extra PR. Thanks!

lukasj (approved):
LGTM

PR #1064, opened 2021-04-09
[master] Maven build - release version property change to 3.1.0

rfelcman:
This is small fix to reflect Maven project 3.1.0-SNAPSHOT changed before.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2019, 2021 Oracle and/or its affiliates. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -98,7 +98,7 @@
         <!-- TOOL Properties -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <!-- PROJECT Properties -->
-        <release.version>3.0.1</release.version>
+        <release.version>3.1.0</release.version>
```
lukasj, line 101: Would it make sense to have this set to ‘${project.version}’ instead?

rfelcman: Sorry, but I don't think, that is possible as this property is used in nightly build output directory at Eclipse.org and some POMs/modules in `bundles` subdirectory.

lukasj: I'm afraid we'll keep forgetting to update this property going forward. What about different solution, ie `osgiversion-maven-plugin` can drop defined parts from a version property (see ie [here](https://github.com/eclipse-ee4j/glassfish-ha-api/blob/master/pom.xml#L174)). Would that be useful for this?

lukasj: ...but we need ie `-RC` suffix, right?

rfelcman: I strongly agree with "forgetting" comment. I'll try something based on http://www.mojohaus.org/build-helper-maven-plugin/regex-property-mojo.html plugin.
About `-RC` suffix. It depends. In Nightly dir no, in Javadoc headers, in https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/resources/org/eclipse/persistence/version.properties maybe (but there is qualifier).

lukasj: in the `version.properties`, `version` should be full `${project.version}` instead. Same in javadoc. So it is probably just a folder name, which would be good to figure out ...

rfelcman: Please check latest changes.

lukasj: yup, that's better. Thanks

lukasj (approved):

PR #1067, opened 2021-04-09
[3.0] Extended logging of JPA L2 cache usage with thread info. - backport from master

rfelcman:
This commit contains extension to the EclipseLink logging to print information about persistence cache operations and usage.
It prints info when entity is stored, refreshed, removed, invalidated in the cache and info about cache hits/miss.
This feature is controlled by new "eclipselink.cache.extended.logging property" and generates messages at the FINEST log level.
It prints additional message if one thread stored item into cache and different picking item.
(cherry picked from commit f2c4e5230ec553b0fc5958cee3572b1849588bb8)


lukasj (approved):
LGTM

PR #1076, opened 2021-04-15
[master] Extended Thread logging

rfelcman:
This PR contains extension to logging code.
Main purpose is display info if cached entity is picked by another thread and WorkManager is called in different threads.
This logging is by default disabled and must be enabled by persistence properties
Major changes are:

- new `thread` log category specified in `org.eclipse.persistence.logging.LogCategory` and `org.eclipse.persistence.logging.SessionLog`
- Two new persistence unit properties `eclipselink.thread.extended.logging` to enable this logging and `eclipselink.thread.extended.logging.threaddump` to add to the log output threads stack trace
- new log messages are called from `org.eclipse.persistence.internal.sessions.UnitOfWorkImpl`

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/LogCategory.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/LogCategory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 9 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/LogCategory.java
```diff
@@ -70,8 +71,9 @@
     SEQUENCING( (byte)0x12, SessionLog.SEQUENCING),
     SERVER(     (byte)0x13, SessionLog.SERVER),
     SQL(        (byte)0x14, SessionLog.SQL),
-    TRANSACTION((byte)0x15, SessionLog.TRANSACTION),
-    WEAVER(     (byte)0x16, SessionLog.WEAVER);
+    THREAD(     (byte)0x15, SessionLog.THREAD),
```
lukasj, line 74: do not change existing constants, add a new one instead - consider backwards compatibility

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/logging/LogCategory.java
```diff
@@ -70,6 +71,7 @@
     SEQUENCING( (byte)0x12, SessionLog.SEQUENCING),
     SERVER(     (byte)0x13, SessionLog.SERVER),
     SQL(        (byte)0x14, SessionLog.SQL),
+    THREAD(     (byte)0x17, SessionLog.THREAD),
```
lukasj, line 74: it should be last item, so ordinal values of enum entries do not change

lukasj (approved):

PR #1095, opened 2021-04-27
[2.6_WAS] Update jpa.jse test antbuild.xml  for jars with git hash

jgrassel:
The antbuild.xml for the `eclipselink.jpa.test.jse` test bucket has hard coded entries for the core, jpa, jpa.jpql, jpa.modelgen jars, which doesn't play nicely if you build eclipselink with the `-Dgit.hash=$GIT_REVISION` option.  Modifying the antbuild.xml to allow for more flexibility with the version portion of the file name.


dazey3 (approved):
LGTM

Issue #1094, opened 2021-04-27
Bug 573198: ConcurrentModificationException accessing EntityManagerFactoryProvider.emSetupImpls


dazey3:
Linking to [BZ #573198](https://bugs.eclipse.org/bugs/show_bug.cgi?id=573198)

Intermittent ConcurrentModificationException:
```
Caused by: java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1456)
	at java.util.HashMap$ValueIterator.next(HashMap.java:1485)
	at org.eclipse.persistence.jpa.PersistenceProvider.isLoadedWithReference(PersistenceProvider.java:448)
	at javax.persistence.Persistence$PersistenceUtilImpl.isLoaded(Persistence.java:154)
	at org.apache.bval.jsr.resolver.JPATraversableResolver.isReachable(JPATraversableResolver.java:34)
	at org.apache.bval.jsr.resolver.DefaultTraversableResolver.isReachable(DefaultTraversableResolver.java:58)
	at org.apache.bval.jsr.resolver.CachingTraversableResolver.isReachable(CachingTraversableResolver.java:81)
	at org.apache.bval.jsr.ConstraintValidation.isReachable(ConstraintValidation.java:454)
```

It appears that access to `org.eclipse.persistence.internal.jpa.EntityManagerFactoryProvider.emSetupImpls`static Map is only sometimes properly synchronized.

PR #1096, opened 2021-04-27
[master] Bug 573198: ConcurrentModificationException in PersistenceProvider

dazey3:
for #1094


lukasj (commented):
Is it possible to write a test for this?

dazey3:
@lukasj 
I'm not quite sure how to write a test for these indeterminate, race conditions. We've only seen this issue once reported by a customer in their testing environments, so we aren't quite sure how to recreate it in the first place. However, based on the exception, and the usage of `EntityManagerFactoryProvider.emSetupImpls` everywhere else, it seemed fairly safe to add more synchronization to `EntityManagerFactoryProvider.emSetupImpls` like already exists everywhere else.

I'm open to any suggestions though on maybe existing tests to take a look or testing ideas in general

lukasj (approved):
`java.util.concurrent` together with high(er) amount of spawned threads usually helps. But I understand it is non-trivial.

LGTM.

Issue #1102, opened 2021-04-29
Bug 521402: Add support for Criteria queries with only literals


dazey3:
Linking to [BZ #521402](https://bugs.eclipse.org/bugs/show_bug.cgi?id=521402)

EclipseLink does not currently accept simple CriteriaBuilder queries that contain only literal values in WHERE clause and/or literal selection fields. Both JPQL and CriteriaBuilder should be in sync.

JPQL:
```java
    Query query = em.createQuery("SELECT 'HELLO' FROM SimpleEntity s");
    query.getResultList();
```
This query executes without error.

CriteriaBuilder:
```java
    CriteriaBuilder cb = em.getCriteriaBuilder();
    CriteriaQuery<Object[]> cquery = cb.createQuery(Object[].class);
    cquery.from(SimpleEntity.class);
    cquery.multiselect(cb3.literal("HELLO"));

    query = em.createQuery(cquery);
    query.getResultList();
```
The query results in the same exception:

```
 org.eclipse.persistence.exceptions.QueryException
Exception Description: A reference class must be provided.
Query: ReportQuery()
	at org.eclipse.persistence.exceptions.QueryException.referenceClassMissing(QueryException.java:1064)
	at org.eclipse.persistence.queries.ObjectLevelReadQuery.checkDescriptor(ObjectLevelReadQuery.java:835)
	at org.eclipse.persistence.queries.DatabaseQuery.execute(DatabaseQuery.java:828)
	at org.eclipse.persistence.queries.ObjectLevelReadQuery.execute(ObjectLevelReadQuery.java:1142)
	at org.eclipse.persistence.queries.ReadAllQuery.execute(ReadAllQuery.java:460)
	at org.eclipse.persistence.queries.ObjectLevelReadQuery.executeInUnitOfWork(ObjectLevelReadQuery.java:1230)
	at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.internalExecuteQuery(UnitOfWorkImpl.java:2900)
	at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1866)
	at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1848)
	at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1813)
	at org.eclipse.persistence.internal.jpa.QueryImpl.executeReadQuery(QueryImpl.java:258)
[...]

PR #1104, opened 2021-04-29
[master] Bug 521402: Add support for Criteria queries with only literals

dazey3:
for #1102

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestSimpleQuery.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestSimpleQuery.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2021 IBM Corporation, Oracle, and/or affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+// Contributors:
+//     IBM - Bug 521402: Add support for Criteria queries with only literals
+package org.eclipse.persistence.jpa.test.query;
+
+import java.util.List;
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.EntityManagerFactory;
+import jakarta.persistence.Query;
+import jakarta.persistence.criteria.CriteriaBuilder;
+import jakarta.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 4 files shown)


jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestSimpleQuery.java
```diff
@@ -0,0 +1,206 @@
...
+import org.junit.runner.RunWith;
+
+@RunWith(EmfRunner.class)
+public class TestSimpleQuery {
+    @Emf(name = "SimpleQueryPersistenceUnit", 
+            createTables = DDLGen.DROP_CREATE, 
+            classes = { SimpleQueryEntity.class }, 
+            properties = { 
+                    @Property(name = "eclipselink.logging.level", value = "ALL"),
+                    @Property(name = "eclipselink.logging.level.sql", value = "FINE"),
```
lukasj, line 45: can this logging be configurable from the command line somehow?

dazey3: I have no idea how to hook command line arguments into the JSE bucket test runner. I can remove these logging properties if you are concerned with them. They should not impact the test and are really just there in case there is a failure in the future.

lukasj: `EmfRunnerInjector.inject` could add provided logging related system properties to PU... I have two use-cases:
* travis kills the job after certain amount of output, so in this case we probably want to completely disable logging
* on local box I want to easily increase the log level in case of failures but I don't want to see all details by default (ie info is sufficient in this case)

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/model/SimpleQueryEntity_.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+// Contributors:
+//     IBM - Bug 521402: Add support for Criteria queries with only literals
+package org.eclipse.persistence.jpa.test.query.model;
+
+import jakarta.persistence.metamodel.SingularAttribute;
+import jakarta.persistence.metamodel.StaticMetamodel;
+
+@StaticMetamodel(SimpleQueryEntity.class)
+public class SimpleQueryEntity_ {
```
lukasj, line 20: would be good if this is created on the fly by modelgen

lukasj (approved):
LGTM, my comments can be addressed via separate PR(s) (...if ever...) ;-)

dazey3:
@lukasj Thanks for looking at this PR! I do agree with your comments, but they feel a bit outside the bounds of this bug/testing. I usually detest falling on the statement "we can do that in a future PR" because I realize it probably wont get done... but in this case, it feels like separate testing/enhancement to this specific bug.

PR #1108, opened 2021-05-04
fix test failures for mysql driver 8.0.24

lukasj:
* `RANK` is reserved word in MySQL 8
* Connector/J 8.0.24 returns `LocalDateTime` from `rs.getObject()` for `TimeStamp` type

To test, update MySQL JDBC driver to the latest one (8.0.24)


rfelcman (approved):
LGTM

PR #1112, opened 2021-05-05
[master] Jenkins build - switch to JDK 11

rfelcman:
Default JDK for build and tests will be Open JDK 11 instead of previous Oracle JDK 8.0.
Maven is switched to latest version 3.6.3.
This change is related with implementation/activation Java Module System JPMS  in PR #1093.
There is also notification system enabled in the Jenkins pipeline to send a mail on unsuccessful and fixed builds and some minor release scripts changes.

### DIFF
--- a/etc/jenkins/build.groovy
+++ b/etc/jenkins/build.groovy
@@ -99,6 +99,10 @@ spec:
 """
         }
     }
+    tools {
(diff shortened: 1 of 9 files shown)


lukasj (approved):
LGTM

PR #1116, opened 2021-05-06
eclipselink 3.0 JPA Modelgenerator does not recognize jakarta schemas

### LINKED ISSUE #1078: eclipselink 3.0 JPA Modelgenerator does not recognize jakarta schemas
When I have orm file with headers like this : 
<entity-mappings xmlns="https://jakarta.ee/xml/ns/persistence/orm"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence/orm
          https://jakarta.ee/xml/ns/persistence/orm/orm_3_0.xsd"
        version="3.0">

Then CanonicalModelProcessor doesn not recognize entity-mappings schema and throws 
Exception Description: A descriptor with default root element {https://jakarta.ee/xml/ns/persistence/orm}entity-mappings was not found in the project

While doing remote debug at 
org.eclipse.persistence.internal.jpa.modelgen.objects.PersistenceUnit.addXMLEntityMappings(String)

it is obvious, that you do not consider ORM_3_0 models

lukasj:
Fixes #1078


rfelcman (approved):
LGTM

PR #1120, opened 2021-05-06
[master] Extended Thread logging - Cache thread info change

rfelcman:
There is change in log message severity and log category. This log message is also produced if persistence `eclipselink.cache.extended.logging` is set to true instead of previous `eclipselink.thread.extended.logging`.


lukasj (approved):
LGTM

Issue #1122, opened 2021-05-07
Bug 573435: CriteriaBuilder construct throws argument type mismatch exception with Case Select


dazey3:
Linking to [BZ #573435](https://bugs.eclipse.org/bugs/show_bug.cgi?id=573435)

When attempting to execute a CriteriaBuilder query, EclipseLink throws an exception. The equivalent JPQL query executes fine and is currently part of the JSE testing bucket

Example:
```java
    CriteriaBuilder cb = em.getCriteriaBuilder();
    CriteriaQuery<Dto01> cquery = cb.createQuery(Dto01.class);
    Root<EntityTbl01> root = cquery.from(EntityTbl01.class);

    SimpleCase<String, Object> selectCase = cb.selectCase(root.get(EntityTbl01_.itemString2));
    selectCase.when("J", "Japan")
        .otherwise("Other");

    Expression<Long> selectCase2 = cb.<Long>selectCase()
        .when(cb.equal(root.get(EntityTbl01_.itemString3), "C"), new Long(1))
        .otherwise(new Long(0));

    Expression<Long> selectCase3 = cb.<Long>selectCase()
        .when(cb.equal(root.get(EntityTbl01_.itemString4), "D"), new Long(1))
        .otherwise(new Long(0));

    cquery.select(cb.construct(Dto01.class, 
        root.get(EntityTbl01_.itemString1),
        selectCase,
        cb.sum(selectCase2),
        cb.sum(selectCase3)));

    query = em.createQuery(cquery);
    query.getResultList();
```

Exception:
```
    java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.eclipse.persistence.internal.security.PrivilegedAccessHelper.invokeConstructor(PrivilegedAccessHelper.java:489)
	at org.eclipse.persistence.queries.ReportQueryResult.processConstructorItem(ReportQueryResult.java:149)
[...]

PR #1123, opened 2021-05-07
[master] Bug 573435: CriteriaBuilder construct throws argument type mismatch e…

dazey3:
for #1122

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2018, 2020 IBM Corporation. All rights reserved.
+ * Copyright (c) 2019, 2021 Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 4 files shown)


lukasj (approved):
LGTM

dazey3:
@lukasj Thank you for the code review!

On the main issue, I brought up the idea of possible type validation and error message throwing if the type is not valid. 

For instance, 
```java
    selectCase = cb.selectCase(root.get(EntityTbl01_.itemInteger1))
        .when(1, "A")
        .when(100, 600)
        .otherwise("C");
    cquery.where(cb.equal(root.get(EntityTbl01_.itemString1), selectCase));
```

In this example, `.when(1, "A")` would set the result type as String.class. Since the previous type defaulted to Object.class, this is a more refined type and accepted. However, `.when(100, 600)` will try to set the result type to Integer.class (maybe Long.class). Currently, we are not going to validate this and just accept the changing type. Since `.otherwise("C");` sets the type back to String.class, there should be any issues with this specific example, but hopefully you can see that if the last call WAS the wrong type, this could lead to exceptions.

Of course, I would expect the error thrown to be very similar to the current exception, but never-the-less we have the opportunity here to do some validation and throw an exception when the type is not consistent. Do you have an opinion on whether we should do this or maybe it has little value.

dazey3:
I guess we can handle error throwing or validation in a separate fix later if it is deemed worthwhile. If the customer uses the API incorrectly and sets the wrong object types, then there will be an issue the same as there is today, just without a pretty message.

PR #1124, opened 2021-05-07
[3.0] Bug 573435: CriteriaBuilder construct throws argument type mismatch exception with Case Select

dazey3:
for #1122


lukasj (approved):
LGTM

Issue #1107, opened 2021-05-04
Incorrect content of the bundles at Eclipse.org download pages for versions 3.0.0 and higher


rfelcman:
There is incorrect content of the EclipseLink bundles published at Eclipse.org download pages.
Be precise content (classes, resource files) is OK, but in some cases next release version e.g. 3.0.2 is applied to content in MANIFEST.MF files and there are incorrect file names e.g. _org.eclipse.persistence.asm-3.0.2-SNAPSHOT.jar_  in 3.0.1 bundle.
This bug is valid for all bundles available at https://www.eclipse.org/eclipselink/downloads/nightly.php from version 3.0.0.
Content published at Maven Central is not affected with this issue.

rfelcman:
Fixed in master and 3.0.

dazey3:
@rfelcman When can I expect a rebuild and update on https://www.eclipse.org/eclipselink/downloads/? The artifacts currently listed in `EclipseLink 3.0.1 OSGi Bundles` still have invalid manifest package versions.

rfelcman:
Sorry, but I don't think, that is simply possible to change these published bundles the with same qualifier. We can make new release 3.0.2.

dazey3:
@rfelcman Cutting a new 3.0.2 is fine with me. Please verify the package versions are good as well and thank you for looking into this issue

PR #1127, opened 2021-05-10
[master] Maven build - Plugin bundle content fix

rfelcman:
This is fix for the bug #1107 "Incorrect content of the bundles at Eclipse.org download pages for versions 3.0.0 and higher."
There is also change in Jenkins pipelines from OpenJDK to AdoptOpenJDK as AdoptOpenJDK integrates latest JDK fixes more quickly.


lukasj (approved):
LGTM

PR #1139, opened 2021-05-14
[master] Maven build - JPARS module Manifest generation

rfelcman:
Behind this fix is some issue with `org.apache.felix:maven-bundle-plugin`. This plugin inserts into generated `MANIFEST.MF` file Maven project properties, but only properties in capital characters. Maven properties in lowercase characters are ignored and not included into manifest. I tried to upgrade into latest version of `org.apache.felix:maven-bundle-plugin:5.1.2`, but without any change.


lukasj (approved):
LGTM

PR #1141, opened 2021-05-14
Consolidate Oracle Platforms

lukasj:
* drop specific support for Oracle DB 9 and older


dazey3:
@lukasj Is there a reason for dropping support for Oracle 9 and older? Is it getting in the way of something?
Also, it looks like you are removing Oracle10 too.

lukasj:
> Is there a reason for dropping support for Oracle 9 and older? Is it getting in the way of something?

see https://www.oracle.com/us/support/library/lsp-tech-chart-069290.pdf

Oracle DB 8 - EOL Dec 2006
Oracle DB 9 - EOL Jul 2010
Oracle DB 10 - EOL Jul 2013
Oracle DB 11 - EOL Dec 2020

Also [DB release schedule](https://support.oracle.com/knowledge/Oracle%20Database%20Products/742060_1.html) shows that it can be reasonable to - in master/3.0 - support only version 11 and newer

> Also, it looks like you are removing Oracle10 too.

No, `Oracle10` is now the base class - `OraclePlatform` (which used to be sth like Oracle7Platform or even older). Only 9, 8 and anything older is being removed.

rfelcman:
Sorry, but there are some JPA Test issues. See attached output from failsafe-reports.
[failsafe-reports.tar.gz](https://github.com/eclipse-ee4j/eclipselink/files/6569021/failsafe-reports.tar.gz)

Issue #1119, opened 2021-05-06
Indention not working if custom CharacterEscapeHandler is set and encoding is set to 'UTF-8'


alerosmile:
Indention is not working if encoding is set to 'UTF-8' (not 'utf-8') and a custom CharacterEscapeHandler is installed. The bug is in method OutputStreamRecord.writeValue(String, boolean, boolean, OutputStream). This method does not write to the given Outputstream. Instead it writes directly to the underlying buffer. This way the tab buffer in FormattedOutputStreamRecord.tab() remains empty.

rfelcman:
Hello,

this is just workaround before this bug will be fixed, but if You need some functional solution it's possible use JAXB API way.

See e.g.
```
...
System.setProperty(JAXBContext.JAXB_CONTEXT_FACTORY, "org.eclipse.persistence.jaxb.JAXBContextFactory");
final HashMap<String, Object> contextProperties = new HashMap<>();
contextProperties.put(JAXBContextProperties.MEDIA_TYPE, "application/xml");
jaxbContext = JAXBContext.newInstance(DOMAIN_CLASSES, contextProperties);

//Unmarshall
Unmarshaller unmarshaller = jaxbContext.createUnmarshaller();
Object unmarshallDoc = unmarshaller.unmarshal(Thread.currentThread().getContextClassLoader().getResource(XML_READ_DOCUMENT));

//Marshall  (JAXB API)
ByteArrayOutputStream byteArrayOutputStream1 = new ByteArrayOutputStream();
Writer writer = new OutputStreamWriter(byteArrayOutputStream1, "UTF-8");
Marshaller marshaller1 = jaxbContext.createMarshaller();
CharacterEscapeHandler handler1 = new TestCharacterEscapeHandler();
marshaller1.setProperty(MarshallerProperties.CHARACTER_ESCAPE_HANDLER, handler1);
marshaller1.marshal(unmarshallDoc, writer);
...
```

```
package com.oracle.moxy.bugtest.domain;

import org.eclipse.persistence.oxm.CharacterEscapeHandler;

import java.io.IOException;
import java.io.Writer;

public class TestCharacterEscapeHandler implements CharacterEscapeHandler {

    @Override
    public void escape(char[] buf, int start, int len, boolean isAttValue, Writer out) throws IOException {
        for (int i = start; i < start + len; i++) {
            char ch = buf[i];

            if (ch == '%') {
                out.write("*");
                continue;
            }

            out.write(ch);
[...]

PR #1146, opened 2021-05-19
[master] Bugfix for - Indention not working if custom CharacterEscapeHandler is set and encoding is set to 'UTF-8

rfelcman:
Bugfix + unit test for "Indention not working if custom CharacterEscapeHandler is set and encoding is set to 'UTF-8" bug #1119 .


lukasj (approved):
LGTM

Issue #1150, opened 2021-05-25
Bug 573361: Add support for parameters within CriteriaBuilder case/coalesce expressions


dazey3:
Linking to [BZ #573361](https://bugs.eclipse.org/bugs/show_bug.cgi?id=573361)

When using parameter markers within CriteriaBuilder case expressions, EclipseLink throws an exception.

Example:
```java
    CriteriaBuilder cb = em.getCriteriaBuilder();
    CriteriaQuery<Object[]> cquery = cb.createQuery(Object[].class);
    Root<SimpleEntity> root = cquery.from(SimpleEntity.class);
    cquery.multiselect(root.get(SimpleEntity_.intVal1));

    ParameterExpression<Integer> checkParam1 = cb.parameter(Integer.class);
    ParameterExpression<Integer> checkParam2 = cb.parameter(Integer.class);
    ParameterExpression<Integer> resultParam1 = cb.parameter(Integer.class);
    ParameterExpression<Integer> resultParam2 = cb.parameter(Integer.class);
    ParameterExpression<Integer> resultParam3 = cb.parameter(Integer.class);

    Expression<Object> selectCase = cb.selectCase()
        .when(cb.equal(root.get(SimpleEntity_.intVal2), checkParam1), resultParam1)
        .when(cb.equal(root.get(SimpleEntity_.intVal2), checkParam2), resultParam2)
        .otherwise(resultParam3);

    cquery.where(cb.equal(root.get(SimpleEntity_.intVal1), selectCase));

    Query query = em.createQuery(cquery);
    query.setParameter(checkParam1, 2);
    query.setParameter(checkParam2, 4);
    query.setParameter(resultParam1, 20);
    query.setParameter(resultParam2, 40);
    query.setParameter(resultParam3, 60);
    query.getResultList();
```

Exception:
```
java.lang.IllegalArgumentException: No parameter with name : Parameter[name=null] was found within the query: ReportQuery(referenceClass=SimpleEntity ).
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:548)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:1)
```

It appears that EclipseLink cannot see the parameter markers being passed into the select Case.

dazey3:
This issue also extends to the coalesce expression as well.

```java
    CriteriaBuilder cb = em.getCriteriaBuilder();
[...]

PR #1151, opened 2021-05-25
[master] Bug 573361: Add support for parameters within CriteriaBuilder case/coalesce expressions

dazey3:
for #1150

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
@@ -25,6 +25,7 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaBuilder.SimpleCase;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
+import jakarta.persistence.criteria.ParameterExpression;
 import jakarta.persistence.criteria.Root;
 
 import org.eclipse.persistence.jpa.test.framework.DDLGen;
@@ -34,6 +35,7 @@ import org.eclipse.persistence.jpa.test.framework.Property;
 import org.eclipse.persistence.jpa.test.query.model.Dto01;
 import org.eclipse.persistence.jpa.test.query.model.EntityTbl01;
 import org.eclipse.persistence.jpa.test.query.model.EntityTbl01_;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -46,7 +48,7 @@ public class TestQueryCase {
     private static boolean POPULATED = false;
 
     @Test
-    public void testQueryCase1() {
+    public void testQuery_JPQL_Case_Literals_1() {
         if (emf == null)
             return;
 
@@ -55,6 +57,7 @@ public class TestQueryCase {
 
         EntityManager em = emf.createEntityManager();
         try {
+            // test 1
             TypedQuery<EntityTbl01> query = em.createQuery(""
                     + "SELECT t FROM EntityTbl01 t "
                         + "WHERE t.itemString1 = ( "
@@ -68,6 +71,7 @@ public class TestQueryCase {
             assertNotNull(dto01);
             assertEquals(0, dto01.size());
 
+            // test 2
             query = em.createQuery(""
                     + "SELECT t FROM EntityTbl01 t "
                         + "WHERE t.itemString1 = ( "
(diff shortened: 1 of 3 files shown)


jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
```diff
@@ -328,8 +572,125 @@ public void testQueryCase3() {
         }
     }
 
+//    @Test
```
lukasj, line 575: is this intentional?

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
```diff
@@ -238,7 +359,130 @@ public void testQueryCase2() {
...
+            if (em.getTransaction().isActive()) {
+                em.getTransaction().rollback();
+            }
+            if(em.isOpen()) {
+                em.close();
+            }
+        }
+    }
+
+//    @Test
```
lukasj, line 484: is this intentional?

lukasj (commented):
Too much of commented out code in tests for my taste. Would it make sense to have it in test with @Ignore or non-test method instead?

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQueryCase.java
dazey3: Yes, these were intentional. I commented out these 2 tests to disable them because they fail for an unrelated failure. I wrote the tests, and I believe they are valid tests, but they currently fail. I chose to only comment them out because I did not want to lose the test. I think using @Ignore is a much better idea though (I forgot that exists) and I will update to use that here instead.

dazey3:
@lukasj 
All of the commented out code in these tests is because either the test fails (`//@Test`) or the current JPA API doesn't even support what I feel should be supported (https://github.com/eclipse-ee4j/jpa-api/issues/315).
I chose to comment them out, instead of removing them entirely, because I didn't want the test to be lost or forgotten. All the tests I added are just taking the Literal test version and changing it to use Parameter markers. I feel all the tests are valid, but some either fail for reasons outside the scope of #1150 or are currently unsupported API calls

If you'd like to know more about the test failure, it has something to do with Constructor usage in CriteriaBuilder and type. The tests are failing with an error that EclipseLink cannot find the `org.eclipse.persistence.jpa.test.query.model.Dto01(String.class, Object.class, Object.class, Object.class)` constructor. I think this bug is more related to Constructor than CASE/COALESCE. I just didnt want to lose the test. 

I am happy to change to use @Ignore on `testQueryCaseLiterals3` and `testQueryCaseParameters3`. As to the `// Currently unsupported by the JPA API` commented out code. Would you rather I just delete it? I can be fine with that if you don't want it.

lukasj:
> @lukasj
> All of the commented out code in these tests is because either the test fails (`//@Test`) or the current JPA API doesn't even support what I feel should be supported ([eclipse-ee4j/jpa-api#315](https://github.com/eclipse-ee4j/jpa-api/issues/315)).
> I chose to comment them out, instead of removing them entirely, because I didn't want the test to be lost or forgotten.

I'm fine with keeping it there as long as there is an explanation and/or a pointer to existing issue (if it exists). Imagine someone with limited knowledge comes through it - such person may just delete comments simply because "it was useless comment"...

dazey3:
@lukasj
> Imagine someone with limited knowledge comes through it - such person may just delete comments simply because "it was useless comment"...
That is a fair statement. I will add a comment to help explain better the reason for the commented out code. 
Also, are you ok with adding references to issues in GH? For instance, referencing https://github.com/eclipse-ee4j/jpa-api/issues/315?

dazey3:
@lukasj 
Ok, I updated the test to be perhaps a bit cleaner/clearer. I apologize for using a git force push, but I get errors using a standard git push:
```
    ! [rejected]              573361_master -> 573361_master (non-fast-forward)
    error: failed to push some refs to 'git@github.com:dazey3/eclipselink.git'
    hint: Updates were rejected because the tip of your current branch is behind
    hint: its remote counterpart. Integrate the remote changes (e.g.
    hint: 'git pull ...') before pushing again.
```
Everything is up to date and rebased. No idea why it does this! I just know that force has always worked.

lukasj:
`git rebase master && git push -f` is my workflow to keep my own commits "untouched" and on top

lukasj (approved):
LGTM

dazey3:
I missed an @Ignore on one of the tests and it failed in travis. I fixed it and pushed an update that just contained the new @Ignore

PR #1156, opened 2021-05-25
[2.7] Bugfix for - Indention not working if custom CharacterEscapeHandler is set and encoding is set to 'UTF-8 - backport from master

rfelcman:
Bugfix + unit test for "Indention not working if custom CharacterEscapeHandler is set and encoding is set to 'UTF-8" bug.


lukasj (approved):
LGTM

PR #1173, opened 2021-05-28
[3.0] Extended Thread logging - backport from master

rfelcman:
There is change in log message severity and log category. This log message is also produced if persistence `eclipselink.cache.extended.logging` is set to true instead of previous eclipselink.thread.extended.logging.

This is backport from PRs #1100 and #1120 .


lukasj (approved):
LGTM

Issue #1225, opened 2021-08-04
Eclipselink 3.0 did not update ASM8 keyword to ASM9 in ClassWeaver and MethodWeaver:


jgrassel:
When we update the level of ASM (or at least major versions), we must remember to update the ASM level referenced in ClassWeaver and ByteWeaver, otherwise Eclipselink will not be reaping the benefit of the newer ASM.  It looks like it was updated in MetadataAsmFactory.

The following was found in `43ff860462` on the 3.0 branch:

```
$ grep -r ASM8 *
src/main/java/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java:        super(ASM8, classWriter);
src/main/java/org/eclipse/persistence/internal/jpa/weaving/MethodWeaver.java:        super(ASM8, mv);
$ grep -r ASM9 *
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metadata/accessors/objects/MetadataAsmFactory.java:            super(Opcodes.ASM9);
```

The 2.7 branch is up-to-date, as can be seen here:
```
$ grep -r ASM8 *
$ grep -r ASM9 *
jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java:        super(ASM9, classWriter);
[...]

PR #1226, opened 2021-08-05
[3.0] Eclipselink 3.0 did not update ASM8 keyword to ASM9 in ClassWeaver and MethodWeaver: #1225

jgrassel:
PR to change the ASM8 reference to ASM9 in ClassWeaver and MethodWeaver

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/ClassWeaver.java
@@ -215,7 +215,7 @@ public class ClassWeaver extends ClassVisitor implements Opcodes {
     }
 
     public ClassWeaver(ClassVisitor classWriter, ClassDetails classDetails) {
-        super(ASM8, classWriter);
+        super(ASM9, classWriter);
         this.classDetails = classDetails;
     }
 
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/MethodWeaver.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/weaving/MethodWeaver.java
@@ -44,7 +44,7 @@ public class MethodWeaver extends MethodVisitor implements Opcodes {
     protected boolean methodStarted = false;
 
     public MethodWeaver(ClassWeaver tcw, String methodName, String methodDescriptor, MethodVisitor mv) {
-        super(ASM8, mv);
+        super(ASM9, mv);
         this.tcw = tcw;
         this.methodName = methodName;
         this.methodDescriptor = methodDescriptor;


lukasj (commented):
I'd rather see something helping us avoiding the same problem in the future.

What about:
* create `EclipseLinkClassVisitor extends ClassVisitor`/`EclipseLinkMethodVisitor extends MethodVisitor` in `eclipselink-asm` like:
```
public class EclipseLinkClassVisitor extends ClassVisitor {
    public EclipseLinkClassVisitor(ClassVisitor cv) {
        //whatever version we repackage
        super(ASM9, cv);
    }
}
```
* let current `ClassWeaver`/`MethodWeaver` use that new class as a base, ie like:
```
public class ClassWeaver extends EclipseLinkClassVisitor implements Opcodes {
    public EclipseLinkClassVisitor(ClassVisitor cv) {
        //superclass defined in eclipselink-asm sets appropriate ASM level for us
        super(cv);
    }
    ....
}
```
This way we could eliminate all references to particular ASM bytecode version from the main codebase, possibly leading to less mistakes like this.

What do you think?

jgrassel:
> This way we could eliminate all references to particular ASM bytecode version from the main codebase, possibly leading to less mistakes like this.

> What do you think?

Hmm, that's one way to go about it, but instead of creating new sub-classes, might it be easier just to add EclipselinkASMConstants.java in the `jpa` project, and add constant `LATEST_ASM` (which has to be updated to refer to the newest ASM* constant) which ClassWeaver.java, MethodWeaver.java, and MetadataAsmFactory.java all reference?

lukasj:
> add constant LATEST_ASM (which has to be updated to refer to the newest ASM* constant) [..] The compiler will still unwrap and store the actual integer value in the generated bytecode.

What happens once we have set this constant say to 9 (say our ASM 9.2), release 3.0.3, update our ASM library to say ASM 10, release it and use it instead of the one released/shipped with eclipselink 3.0.3 by default? Will, in this case, EclipseLink 3.0.3 use ASM10 or ASM9? Due to inlining, it seems that ASM9, but what we want is ASM10 to be used instead.

jgrassel:
> > add constant LATEST_ASM (which has to be updated to refer to the newest ASM* constant) [..] The compiler will still unwrap and store the actual integer value in the generated bytecode.
> 
> What happens once we have set this constant say to 9 (say our ASM 9.2), release 3.0.3, update our ASM library to say ASM 10, release it and use it instead of the one released/shipped with eclipselink 3.0.3 by default? Will, in this case, EclipseLink 3.0.3 use ASM10 or ASM9? Due to inlining, it seems that ASM9, but what we want is ASM10 to be used instead.

Yeah, I see what you say.  However, waiting until the eclipselink-asm project goes live is going to delay fixing Eclipselink though.  How long do we want to wait?

lukasj:
> However, waiting until the eclipselink-asm project goes live is going to delay fixing Eclipselink though. How long do we want to wait?

If you can create PR there in ASM project, I can merge it. Since @rfelcman is offline already, o.e.p.asm 9.1.1 release will have to wait for tomorrow, some ~14-15hours from now. Is that reasonable waiting time?

rfelcman:
Hi, I can release updated ASM 9.1.1 later today when PR into https://github.com/eclipse-ee4j/eclipselink-asm will be merged.

jgrassel:
> > However, waiting until the eclipselink-asm project goes live is going to delay fixing Eclipselink though. How long do we want to wait?
> 
> If you can create PR there in ASM project, I can merge it. Since @rfelcman is offline already, o.e.p.asm 9.1.1 release will have to wait for tomorrow, some ~14-15hours from now. Is that reasonable waiting time?

It looks like I'll need permissions, it's not letting me create a new issue on the eclipselink-asm repo.

jgrassel:
I've opened https://github.com/eclipse-ee4j/eclipselink-asm/issues/4 for this feature.

jgrassel:
And put up a PR for the update, https://github.com/eclipse-ee4j/eclipselink-asm/pull/6

jgrassel:
Should I close this PR, or should I use it to update the maven dependency to ecl-asm 9.1.1 when that is out?

lukasj:
That’s up to you. Starting over in new PR mah be easier. You don’t have to wait, just follow https://github.com/eclipse-ee4j/eclipselink-asm/pull/6#issuecomment-893726729

lukasj:
this was covered by #1234

PR #1229, opened 2021-08-05
[master] Minor JSE test failure fixes

dazey3:
1. CacheDeadLockDetectionTest - Failing on DB2
```
Internal Exception: com.ibm.db2.jcc.am.SqlSyntaxErrorException: DB2 SQL Error: SQLCODE=-542, SQLSTATE=42831, SQLERRMC=ID, DRIVER=3.72.24
Error Code: -542
Call: CREATE TABLE cachedeadlock_master (id integer PRIMARY KEY, name varchar(200))
```
This failure is because DB2 expects the "NOT NULL" clause for PRIMARY KEY columns: https://www.ibm.com/docs/en/db2-for-zos/11?topic=codes-542 

2. TestJavaTimeTypeConverter.timeConvertUtilDateToLocalDate - Failing depending on the JDK locale
```
<failure message="expected: '120' but was: '119'" type="junit.framework.AssertionFailedError">junit.framework.AssertionFailedError: expected: '120' but was: '119'
	at org.eclipse.persistence.jpa.test.conversion.TestJavaTimeTypeConverter.timeConvertUtilDateToLocalDate(TestJavaTimeTypeConverter.java:77)
	at org.eclipse.persistence.jpa.test.framework.EmfRunner.run(EmfRunner.java:43)
</failure>
```
What's happening here is that the `java.util.Calendar` is set to `Jan 1, 2020 00:00:00 UTC`, but when `java.util.Calendar.getTime()` is called, it creates a `java.util.Date` object. However, `java.util.Date` has no timezone, so it must be converted.

UTC (0000) gets translated into CST (1800); which is the local timezone for my JDK. But UTC (0000) is 6 hours ahead of CST, its really CST = ( Jan 1, 2020 00:00:00 UTC - 6 ) hours, making it Dec 31, 2019 18:00:00 CST. This conversion to causing the test to fail depending on the JDK locale the test is being run in.

3. TestCoalesceFunction - Failing on Oracle
```
    java.sql.SQLSyntaxErrorException: ORA-00904: : invalid identifier
Error Code: 904
Call: CREATE TABLE COALESCEENTITY (ID NUMBER(10) NOT NULL, BIGDECIMAL NUMBER(38) NULL, DATE DATE NULL, DESCRIPTION VARCHAR2(255) NULL, PRIMARY KEY (ID))
```
Using an attribute name of "date" causes Oracle to fail using a reserved word

4. TestReturnInsert - Failing on Oracle
```
<failure message="expected: '1' but was: '31'" type="junit.framework.AssertionFailedError">junit.framework.AssertionFailedError: expected: '1' but was: '31'
	at org.eclipse.persistence.jpa.returninsert.TestReturnInsert.testCreate(TestReturnInsert.java:153)
	at org.eclipse.persistence.jpa.returninsert.TestReturnInsert.test(TestReturnInsert.java:60)
</failure>
```
This is failing similar to TestJavaTimeTypeConverter. The test is expecting the date value of 1 (Jan 01 1970), but is getting 31 (Dec 31 1969) because the local locale of CST (non-UTC 0)

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/returninsert/TestReturnInsert.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/returninsert/TestReturnInsert.java
@@ -14,7 +14,8 @@
 //     Oracle - initial API and implementation
 package org.eclipse.persistence.jpa.returninsert;
 
-import java.time.Instant;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.Date;
 
 import jakarta.persistence.EntityManager;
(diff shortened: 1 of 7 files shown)


lukasj (approved):
LGTM

PR #1237, opened 2021-08-09
[2.7] Update Eclipselink to use Eclipselink*Visitor types

jgrassel:
Eclipselink*Visitor types have been added to the Eclipselink ASM project, which moves the use of the ASM# constants out of the Eclipselink project and into the Eclipselink ASM project. This should make it easier to avoid missing updating all of the constants when moving up to new major versions of ASM.

(Updating to use Eclipselink ASM 9.1.1)


jgrassel:
seems like we have a good number of infrastructure problems today. @lukasj

lukasj (approved):
LGTM

PR #1243, opened 2021-08-16
[master] Project README.md update

rfelcman:
This commit extends current project `README.md` with info about latest published version in Maven Central, Nightly build status and latest available snapshot (master only).


lukasj (approved):
LGTM

PR #1244, opened 2021-08-16
Maven build issue with Maven 3.8.2 fix

### LINKED ISSUE #1242: Problem with building with maven 3.8.2
Since maven 3.8.2 (which includes https://github.com/apache/maven/commit/5a8997312680a4b9c69a129801524691bc546c08 that I think is the cause of change) the EL can no longer be built due to [directory-maven-plugin](https://github.com/jdcasey/directory-maven-plugin) not being compatible (and BTW probably dead anyway).

It results with:
```
[ERROR] Failed to execute goal org.commonjava.maven.plugins:directory-maven-plugin:0.3.1:highest-basedir (find-project-root) on project org.eclipse.persistence.parent: Cannot find a single highest directory for this project set. First two candidates directories don't share a common root. -> [Help 1]
```

As maven versions are not specified exactly in 
https://github.com/eclipse-ee4j/eclipselink/blob/c0f808e3204abeafeaf9c63e69b16d0e0f75571f/.github/workflows/codeql-analysis.yml#L57  or in https://github.com/eclipse-ee4j/eclipselink/blob/c0f808e3204abeafeaf9c63e69b16d0e0f75571f/etc/jenkins/pr_verify.groovy#L94 they might be upgraded out of control of this project.

Please consider taking action before that happens.

rfelcman:
Fixes #1242


pzygielo:
Thanks.
Could you use closing word (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in very first comment, to have issue automatically closed on merge? I suppose the extra `for` is not supported by GH yet (as GH didn't link the issue with this PR).

michael-o:
There is now a similar case: http://mail-archives.apache.org/mod_mbox/maven-dev/202108.mbox/%3CCAGjJkv0CL63THRrT4w7Fpj%2BNQ5NyjtnAOJ8yW7OAmbMkxGjT4Q%40mail.gmail.com%3E. I'd like to wait for @mickaelistria's response.

mickaelistria:
I think the proper way to check whether a project is in the reactor or not is `session.getProjects().contains(project)`. But maybe I'm missing a part of the Maven API contract. At least, there is no test in Maven that assumed that `project.getBasedir() == null` for a project that's not in the reactor...

pom.xml
```diff
@@ -1192,18 +1187,28 @@
...
-                            <property>project.root.location</property>
+                            <source>
+                                <![CDATA[
+                                    def rootPom = project;
+                                    while (rootPom.parent.groupId != "org.eclipse.ee4j") {
+                                        rootPom = rootPom.parent;
+                                    }
+                                    project.properties.setProperty("project.root.location", rootPom.basedir.toString());
+                                    log.info("EclipseLink Parent project directory [project.root.location]:  " + project.properties['project.root.location']);
+                                ]]>
```
lukasj, line 1210: Why can't we rely on `${maven.multiModuleProjectDirectory}` or `${session.executionRootDirectory}`? Do we really support running `mvn` in submodules (ie without `mvn -pl :...`)?

michael-o: You cannot rely on `${maven.multiModuleProjectDirectory}`. This is private property (implementation detail).

lukasj: What else would you recommend?

michael-o: You basically need to solve: https://issues.apache.org/jira/browse/MNG-7038

lukasj: to keep resources shared by all submodules on one place without a need to define a to be redefined in each submodule property, to avoid having multiple copies of the same file on multiple places through the repo or to avoid having to maintain an extra artifact having them. Examples of such resources: license/notice files from the repo root which every build archive must include or common configuration files for various plugins (can be copyright template, common spotbugs exclusion filter, etc)

michael-o: You name it: avoid having to maintain an extra artifact having them. That is the usecase for a POM module with attached classified artifact. @rfscholte or do you see a better way?

rfscholte: A shared resource doesn't have to be restricted classes, it can be of any type. So the preferred way is indeed a artifact that can be used as dependency. Whether it is a classified artifact of "root" or a new modX, that's up to the user.
I still aim for proper context: Maven modules should never look outside their own directory.

lukasj: right, shared resource can be anything. Not sure if all plugins support reading files from such resource, so it may be required to add additional copy/unpack step into submodules to put them into project.build.directory during the build unless one wants to fix all such plugins...

michael-o: No they don't. You have already described the correct process. This is what we do in Maven and me in my personal OSS projects.

bundles/eclipselink/pom.xml
```diff
@@ -185,6 +185,14 @@
         </dependency>
 
         <!-- included artifacts -->
+        <dependency>
+            <groupId>org.eclipse.persistence</groupId>
+            <artifactId>org.eclipse.persistence.parent</artifactId>
+            <version>${project.version}</version>
+            <type>zip</type>
+            <classifier>common-license.zip</classifier>
```
lukasj, line 193: no need to have `.zip` suffix in the `classifier`, it is already present in the `type`

src/main/assembly/common-license.xml
```diff
@@ -0,0 +1,43 @@
...
+          xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
+    <id>common-license.zip</id>
+    <formats>
+        <format>zip</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <fileSets>
+        <fileSet>
+            <directory>${project.basedir}</directory>
+            <outputDirectory>.</outputDirectory>
```
lukasj, line 27: I'd probably put one set to one folder and use a wildcard (`*`) during unpack if possible - this could avoid potential issues should we need to add/remove something

lukasj (approved):
I have a fix for issue you see available, will push that through the next PR in a minute or two...

PR #1246, opened 2021-08-20
[master] ConcurrencyManager dead-lock detection diagnostic improvement 2.0.2

rfelcman:
Minor fix in logging output to remove `<ENTITY>.toString()` call during logging. It should be potentially not safe (e.g. resources consuming, dead-lock, infinite loop). Class name and primary key is enough there to identify cache object.


lukasj (approved):
LGTM

PR #1271, opened 2021-09-03
[3.0] Jenkins build fix to reflect/use JDK specified in the Jenkins pipeline

rfelcman:
This fix removes `. /etc/profile` call from scripts called by various Jenkins pipelines.
In `. /etc/profile` was hardcoded JDK 8 so Jenkins build/test didn't reflect JDK settings from pipeline.


lukasj (approved):
LGTM

PR #1285, opened 2021-09-06
[3.0] Maven build issue with Maven 3.8.2 fix - backport

### LINKED ISSUE #1242: Problem with building with maven 3.8.2

rfelcman:
Fixes #1242

### DIFF
--- a/bundles/eclipselink/pom.xml
+++ b/bundles/eclipselink/pom.xml
@@ -190,6 +190,14 @@
         <!-- included artifacts -->
         <dependency>
             <groupId>org.eclipse.persistence</groupId>
+            <artifactId>org.eclipse.persistence.parent</artifactId>
+            <version>${project.version}</version>
+            <type>zip</type>
+            <classifier>common-license</classifier>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.persistence</groupId>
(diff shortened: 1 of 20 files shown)


lukasj (approved):
LGTM

PR #1288, opened 2021-09-08
[2.6_WAS] Use eclipselink asm project

jgrassel:
Update Eclipselink to draw upon Eclipselink ASM project instead of harboring asm source in repository.

To use:
1) Change directory to plugins/org.eclipse.persistence.asm
2) Update the `<properties>` block in the `pom.xml` file, specifically `<eclipselink.asm.version>9.1.1</eclipselink.asm.version>`, to use the version number of the new maven resource. 
3) Invoke `mvn verify`
4) Update ASM version ranges as listed by mvn output

This script will update the version number references in .classpath, .MF, and ant.properties files.  However, files which use version ranges (ie, `;version="[7.0,10.0)"` must be updated manually.

Then build as normal.  Good idea to use `git status` and `git diff` to ensure that the maven script updated the relevant files, as well as deleted the old asm bundle in `plugins/`.

This only removes the necessity of having to maintain a copy of the ASM source within the eclipselink repo.  It does not change the fact that the asm binaries are still stored by the git repo.  The 2.6_WAS branch has not been modified to remove this necessity as has been done for 2.7 and later


dazey3 (approved):
LGTM

PR #1292, opened 2021-09-09
[3.0] Bugfix: NullPointerException at CacheIndexMetadata.getField()

ivvlev:
PROBLEM
=======
NullPointerException raising at CacheIndexMetadata.getField() while EntityManager initialising.

CAUSE
=======
Entity metadata XML file contains tag <cache-index><column-name>SOMENAME</column-name></cache-index>.
XML Deserializer call CacheIndexMetadata default constructor. ORMetadata.m_project field stay none initialised.
Call of CacheIndexMetadata.getField() throw to exception.

SOLVING
=======
Set ORMetadata.m_project value after XML deserialization.

Signed-off-by: Vladimir Ivanov <ivvlev@ivvlev.com>


ivvlev:
[Eclipselink3_Bug_CacheMetaIndexNPE.zip](https://github.com/eclipse-ee4j/eclipselink/files/7136012/Eclipselink3_Bug_CacheMetaIndexNPE.zip)
Sample project

rfelcman (approved):
LGTM

PR #1302, opened 2021-09-16
[master] SQL Struct with @Lob (BLOB, CLOB) used as stored procedure parameter type - fix

### LINKED ISSUE #1299: SQL Struct with @Lob (BLOB, CLOB) used as stored procedure parameter type
There is a bug if following SQL Object type
```
CREATE OR REPLACE TYPE FS_INNER_OBJ_BLOB as object
( AuthResult  VARCHAR2(20),
  errorDetails VARCHAR2(250),
  blobvariable BLOB);
```
is used as a parameter in PL/SQL stored procedure like
`CREATE OR REPLACE FUNCTION save_blob_in_struct(first IN VARCHAR2, inner_obj  IN FS_INNER_OBJ_BLOB) RETURN VARCHAR2...`
Generated anonymous PL/SQL code doesn't have `FS_INNER_OBJ_BLOB(null, null, null)` constructor and JDBC Blob is not created correctly.

rfelcman:
Fixes #1299
Bugfix and unit test.

[Oracle_Test_Results.zip](https://github.com/eclipse-ee4j/eclipselink/files/7175620/Oracle_Test_Results.zip)

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/structures/ObjectRelationalDataTypeDescriptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/structures/ObjectRelationalDataTypeDescriptor.java
@@ -17,7 +17,6 @@
 package org.eclipse.persistence.mappings.structures;
 
 import java.sql.Array;
-import java.sql.Blob;
 import java.sql.Ref;
 import java.sql.Struct;
 import java.sql.Types;
@@ -337,16 +336,10 @@ public class ObjectRelationalDataTypeDescriptor extends RelationalDescriptor {
             Object[] fields = new Object[getOrderedFields().size()];
             for (int index = 0; index < getOrderedFields().size(); index++) {
                 DatabaseField field = (DatabaseField)getOrderedFields().elementAt(index);
-                if (row.getField(field) != null && "java.sql.Blob".equals(row.getField(field).getTypeName())) {
-                    Blob blob = connection.createBlob();
-                    blob.setBytes(1L, (byte[]) row.get(field));
-                    fields[index] = blob;
-                } else {
-                    fields[index] = row.get(field);
-                }
+                fields[index] = row.get(field);
             }
 
-            structure = session.getPlatform().createStruct(getStructureName(), fields, session, connection);
+            structure = session.getPlatform().createStruct(getStructureName(), fields, row, getOrderedFields(), session, connection);
         } catch (java.sql.SQLException exception) {
             throw DatabaseException.sqlException(exception, session, false);
         } finally {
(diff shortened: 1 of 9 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/structures/ObjectRelationalDataTypeDescriptor.java
```diff
@@ -336,7 +337,13 @@ public Struct buildStructureFromRow(AbstractRecord row, AbstractSession session,
             Object[] fields = new Object[getOrderedFields().size()];
             for (int index = 0; index < getOrderedFields().size(); index++) {
                 DatabaseField field = (DatabaseField)getOrderedFields().elementAt(index);
-                fields[index] = row.get(field);
+                if (row.getField(field) != null && "java.sql.Blob".equals(row.getField(field).getTypeName())) {
```
lukasj, line 340: check either against `ClassConstants.BLOB`/`DBField.getType()` or  `JDBCType.BLOB`/`DBField.getSqlType` is better thing to do

rfelcman: I agree, but `field` properties behind `field/row.getField(field) .getType()` or `field/row.getField(field) .getSqlType()` are not available at this moment. Only `typeName` is there.

lukasj: Ok then. What about `Clob`? The title mentions it but I don't see it explicitly handled

rfelcman: I'm still not sure if conversion like this should be there. Maybe more correct way should be call something like
`fields[index] = session.getDatasourcePlatform().convertObject(row.get(field), java.sql.Blob.class)...` or something more generic
instead of 
`for (int index = 0; index < getOrderedFields().size(); index++) {....}`?

What I see little problematic in this way is that "live" instance of `java.sql.Connection` for `java.sql.B(C)lob` creation is needed. Is it correct or possible use `java.sql.Connection` in `org.eclipse.persistence.platform.database.*Platform` classes?
But I see `org.eclipse.persistence.platform.database.*Platform` classes (`org.eclipse.persistence.platform.database.oracle.Oracle12Platform` in this particular case) as a better location.

lukasj: there is a call to platform specific method already a line bellow the code you've added - can the logic be moved there, into the `session.getPlatform().createStruct(getStructureName(), fields, session, connection);` method? The _bad_ field is there in the `fields` variable..

lukasj (approved):
This looks way better! LGTM

PR #1313, opened 2021-09-18
#1312: Remove unversioned JPARS resources

### LINKED ISSUE #1312: Remove unversioned JPARS resources
JPARS resources can use no version, version `v1`, `v2` or `latest`. As of 4.0, 'unversioned' access should be dropped as it is available as `v1`

lukasj:
this
* removes unversioned rest endpoints
* makes v2.0 the default one
* makes previous version available only as v1.0

fixes #1312


rfelcman (approved):
LGTM

PR #1322, opened 2021-09-23
[2.6] SQL Struct with @Lob (BLOB, CLOB) used as stored procedure parameter type - backport

### LINKED ISSUE #1299: SQL Struct with @Lob (BLOB, CLOB) used as stored procedure parameter type

rfelcman:
Fixes #1299
Bugfix and unit test.

[test-results.tar.gz](https://github.com/eclipse-ee4j/eclipselink/files/7215822/test-results.tar.gz)


lukasj (approved):
LGTM

PR #1336, opened 2021-10-01
[master] SQL Struct with @Lob (BLOB, CLOB) used as stored procedure parameter type - fix 2.0

rfelcman:
Additional fix in `Oracle12Platform` class to use standard `java.sql.Connection.createStruct(....)` method instead of `org.eclipse.persistence.platform.database.oracleOracle8Platform.createStruct(...)` with deprecated `new oracle.sql.STRUCT(...)` call.


lukasj (approved):
LGTM

PR #1352, opened 2021-11-09
[2.6_WAS] Update to use Eclipselink ASM 9.2

jgrassel:
As per the title, update Eclipselink 2.6_WAS to use ASM 9.2


dazey3 (approved):
lgtm

PR #1355, opened 2021-11-11
[master] IdentifiableTypeImpl throws IllegalArgumentException with EmbeddedId and relations - bugfix + unit test

### LINKED ISSUE #1228: IdentifiableTypeImpl throws IllegalArgumentException with EmbeddedId and relations
I am currently trying to use an Embedded ID together with relations. However, i get always a a error it i would like to getId() because there the MapsId are not ignored. Foo and bar are also identified as PK candidates. But it shouldn't, because MapsId is set.

Example for a many to many relation entity.

```
@Entity
public class FooBar  {

   @EmbeddedId
   private FooBarId id = new FooBarId();

   @ManyToOne
   @JoinColumn( name = "foo_id" )
   @MapsId( "fooId" )
   private Foo foo;

   @ManyToOne
   @JoinColumn( name = "bar_id")
   @MapsId( "barId" )
   private Bar bar;

}
```

```
@Embeddable
public class FooBarId implements Serializable {

   @Column( name = "foo_id" )
   private Long fooId;

   @Column( name = "bar_id")
   private Long barId;

}
```
The `EntityTypeImp<FoorBar>` has now always 3 id attributes with the java typ: `FooBarId, Foo, Bar`
The `hasSingleIdAttribute `return `true`, but the check  `type.getCanonicalName().equals(anAttribute.getJavaType().getCanonicalName()))` cannot be true for the different java class types.

```
 EntityTypeImp<FoorBar> source= ...
 source.getId(FoorBarId) --> throws exception
```

```
    @Override
    public <Y> SingularAttribute<? super X, Y> getId(Class<Y> type) {
        // We assume that there is at most a single EmbeddedId
        SingularAttribute<? super X, Y> idAttribute = null;
        if(!hasSingleIdAttribute()) {
            // Id is part of an IdClass
            throw new IllegalArgumentException(ExceptionLocalization.buildMessage(
                    "metamodel_identifiable_id_attribute_is_incorrect_idclass",
                    new Object[] { this }));
        } else {
            // verify single id attribute type
            for(SingularAttribute<? super X, ?> anAttribute : idAttributes) {
                // Verify type is correct - relax restriction on null and Object.class (from same classLoader)
                if(null == type || Object.class == type ||
                        type.getCanonicalName().equals(anAttribute.getJavaType().getCanonicalName())) {
                    idAttribute = (SingularAttribute<? super X, Y>) anAttribute;
                } else {
                    throw new IllegalArgumentException(ExceptionLocalization.buildMessage(
                        "metamodel_identifiable_id_attribute_type_incorrect",
                        new Object[] { anAttribute, this, type, anAttribute.getJavaType() }));
                }
            }
        }
        return idAttribute;
    }
```

Either the maps should not be included in the id attributes or should be ignored during this check.

rfelcman:
Because there are four new objects (java classes) in org.eclipse.persistence.testing.models.jpa.metamodel package MetamodelMetamodelTest.METAMODEL_ALL_TYPES is increased into 56.
These four objects have eights new attributes. This is why MetamodelMetamodelTest.METAMODEL_ALL_ATTRIBUTES_SIZE is increased into 158.

fixes #1228


lukasj (approved):
LGTM

Issue #1354, opened 2021-11-10
Bug 463042: Concurrency issue with Case expression operator


dazey3:
Linking to [BZ #463042](https://bugs.eclipse.org/bugs/show_bug.cgi?id=463042)

Sample project that reproduces the issue. Sample contains 1 class, 1 persistence.xml (uses H2 in-mem DB) and 1 pom.xml (optional)

When using a NamedQuery that contains the 'CASE' operator, the query is no longer ThreadSafe.

The problem can easily be reproduced by firing a NamedQuery containing a case operator within multiple concurrent threads (see example attached).

I was not able to understand the actual problem, but somehow the org.eclipse.persistence.expressions.ListExpressionOperator is shared across threads and its 'numberOfItems' is modified concurrently which eventually leads to all kinds of problems (either ArrayIndexOutOfBoundsException or NullPointerException).

Here is one of the stacktrace I had:

```
java.lang.ArrayIndexOutOfBoundsException: 11
	at org.eclipse.persistence.expressions.ListExpressionOperator.getDatabaseStrings(ListExpressionOperator.java:75)
	at org.eclipse.persistence.expressions.ExpressionOperator.printCollection(ExpressionOperator.java:2170)
	at org.eclipse.persistence.internal.expressions.ArgumentListFunctionExpression.printSQL(ArgumentListFunctionExpression.java:99)
	at org.eclipse.persistence.internal.expressions.FunctionExpression.writeFields(FunctionExpression.java:717)
	at org.eclipse.persistence.internal.expressions.SQLSelectStatement.writeFieldsFromExpression(SQLSelectStatement.java:2033)
	at org.eclipse.persistence.internal.expressions.SQLSelectStatement.writeFieldsIn(SQLSelectStatement.java:2048)
	at org.eclipse.persistence.internal.expressions.SQLSelectStatement.printSQL(SQLSelectStatement.java:1675)
	at org.eclipse.persistence.internal.databaseaccess.DatabasePlatform.printSQLSelectStatement(DatabasePlatform.java:3178)
	at org.eclipse.persistence.platform.database.H2Platform.printSQLSelectStatement(H2Platform.java:53)
	at org.eclipse.persistence.internal.expressions.SQLSelectStatement.buildCall(SQLSelectStatement.java:782)
	at org.eclipse.persistence.internal.expressions.SQLSelectStatement.buildCall(SQLSelectStatement.java:792)
[...]

PR #1359, opened 2021-11-11
[master] Bug 463042: Concurrency issue with Case expression operator

dazey3:
for #1354


rfelcman (approved):
LGTM

PR #1364, opened 2021-11-24
[master] problems with session customizer when upgrading to 2.7.9 - bugfix + unit test

### LINKED ISSUE #1363: problems with session customizer when upgrading to 2.7.9
Hi folks we've had code running on various versions of eclipse link up thru 2.7.8 that worked fine for using session customizers to inject fields as part of transactions. In 2.7.9 something seems to have broken. Was there any changes around this area that i can look into?

We do (roughly):

ReturningPolicy rp = new ReturningPolicy();
rp.addFieldForInsertReturnOnly(table + "." + JPAEntityUtils.getPhysicalColumnName(doClz, Updateable.CREATE_DATE), Date.class);
rp.addFieldForInsertReturnOnly(table + "." + JPAEntityUtils.getPhysicalColumnName(doClz, Updateable.UPDATE_DATE), Date.class);
rp.addFieldForUpdate(table + "." + JPAEntityUtils.getPhysicalColumnName(doClz, Updateable.UPDATE_DATE), Date.class);

descriptor.setReturningPolicy(rp);

where we do:

create or replace trigger %{TRIGGER_NAME}
before insert or update
on %{BASE_TABLE_NAME}
FOR EACH ROW
BEGIN
:NEW.%{UPDATE_TS_COLUMN} := %{SYSDATE_REFERENCE};
IF INSERTING THEN
:NEW.%{CREATE_TS_COLUMN} := %{SYSDATE_REFERENCE};
END IF;
end;

The issue is, now, when update happens, we clear out the create date column. Setting of the create date should only happen on the create, not update. If we roll back to 2.7.8 it works as it should.

rfelcman:
fixes #1363

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
@@ -837,12 +837,12 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
         if (getDescriptor().hasReturningPolicies() && getDescriptor().getReturnFieldsToGenerateUpdate() != null) {
             // In case of RelationalDescriptor only return fields for current table must be used.
             List<DatabaseField> returnFieldsForTable = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


lukasj (approved):
LGTM

PR #1383, opened 2022-01-05
Bug in the PersistenceUnitProcessor.findPersistenceArchives(ClassLoader loader, String descriptorPath) - bugfix

### LINKED ISSUE #1348: Bug in the PersistenceUnitProcessor.findPersistenceArchives(ClassLoader loader, String descriptorPath)
I hit an issue when i pack an app in WAR with structure:
- /WEB-INF/classes/META-INF/persistence.xml

inside the PersistenceUnitProcessor.findPersistenceArchives(ClassLoader loader, String descriptorPath):

//where the puRootUrl = jar:file:/..../WEB-INF/classes/' and the descriptorPath ( = META-INF/persistence.xml [the default location])

archive = PersistenceUnitProcessor.getArchiveFactory(loader).createArchive(puRootUrl, descriptorPath, null);

where the archive object will end-up be an 'DirectoryInsideJarURLArchive' instance, with method:

 //entryPath = 'META-INF/persistence.xml'
    public InputStream getEntry(String entryPath) throws IOException {
        InputStream is = entries.contains(entryPath) ?
            jarFile.getInputStream(jarFile.getEntry(relativeRootPath + entryPath)) : null;
        return is;
    }

//the entries.contains(entryPath) will always return false (as the entries contains strings with leading '/' but the entryPath isn't)

also the (relativeRootPath + entryPath) should be => (relativeRootPath + "/" + entryPath) 
//as the relativeRootPath = 'WEB-INF/classes' which form a path string '**WEB-INF/classesMETA-INF/persistence.xml**'

---------------------------

This bug which finally lead to NPE and not able to load the persistence.xml file correctly.

is it a bug in the PersistenceUnitProcessor or the DirectoryInsideJarURLArchive class?

---------------------------

source references:
- https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/deployment/DirectoryInsideJarURLArchive.java
- https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/deployment/PersistenceUnitProcessor.java
-

rfelcman:
This small fix allows using/loading persistence.xml from WAR file from `.../WEB-INF/classes/META-INF/persistence.xml` location in JSE environment.
It is still required to point EclipseLink (EntityManagerFactory) to this location in JSE environment by:

`System.setProperty(PersistenceUnitProperties.ECLIPSELINK_PERSISTENCE_XML, "WEB-INF/classes/META-INF/persistence.xml");`
or

``
Properties persistenceUnitProperties = new Properties();
persistenceUnitProperties.setProperty(PersistenceUnitProperties.ECLIPSELINK_PERSISTENCE_XML, "WEB-INF/classes/META-INF/persistence.xml");
EntityManagerFactory entityManagerFactory = Persistence.createEntityManagerFactory("test-jpa-pu2", persistenceUnitProperties);
``

Before this fix ValidationException
e.g.

``
Internal Exception: Exception [EclipseLink-7357] (Eclipse Persistence Services - 4.0.0.v202112161344): org.eclipse.persistence.exceptions.ValidationException
Exception Description: URL [jar:file:/tmp/test1348.war!/WEB-INF/classes/META-INF/persistence.xml] for resource [WEB-INF/classes/META-INF/persistence.xml] does not belong to a valid persistence root (as per JPA Specification)
``

was thrown

Fixes #1348

### DIFF
--- a/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
+++ b/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
--- a/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
+++ b/jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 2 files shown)


lukasj (commented):
It would be probably good to add a (unit) test for this class to cover valid/invalid scenarios to have something for the future reference

rfelcman:
> It would be probably good to add a (unit) test for this class to cover valid/invalid scenarios to have something for the future reference

OK it's done by second commit.

jpa/eclipselink.jpa.test/src/it/java/org/eclipse/persistence/testing/tests/jpa/advanced/PersistenceUnitProcessorTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2021 Oracle and/or its affiliates. All rights reserved.
```
lukasj, line 2: I know it's sad but it is 2022 already ... :-)

lukasj (approved):
LGTM

PR #1389, opened 2022-01-12
Common JSON Support and MySQL/Postgres/Oracle 21c native JSON type support

Tomas-Kraus:
Basic JSON support.
JPA Entity allows JsonValue/JsonObject/JsonArray parameters without any additional annotations.
```
@Entity
public class JsonEntity {

    @Id
    private long id;

    private JsonValue value;

    public JsonEntity() {
        this.id = 0;
        this.value = null;
    }

    public JsonEntity(final long id, final JsonValue value) {
        this.id = id;
        this.value = value;
    }

    public long getId() {
        return id;
    }

    public void setId(final long id) {
        this.id = id;
    }

    public JsonValue getValue() {
        return value;
    }

    public void setValue(final JsonValue value) {
        this.value = value;
    }

}
```
EntityManager now works with such an entity:
```
@Test
public void testCreateJsonField() {

    JsonValue value = Json.createObjectBuilder()
            .add("id", "1001")
            .add("name", "Joe Wright")
            .add("age", 49)
            .build();

    try {
        em.getTransaction().begin();
        JsonEntity e = new JsonEntity(1001, value);
        em.persist(e);
        em.getTransaction().commit();

        JsonEntity dbValue = em.createQuery("SELECT v FROM JsonEntity v WHERE v.id=:id", JsonEntity.class)
                .setParameter("id", e.getId()).getSingleResult();
        Assert.assertEquals(value, dbValue.getValue());

    } finally {
        if (em.getTransaction().isActive()) {
            em.getTransaction().rollback();
        }
        em.close();
    }
}
```
DDL generated database table before MySQL specific type was added:
```
mysql> desc JSONENTITY;
+-------+--------------+------+-----+---------+-------+
| Field | Type         | Null | Key | Default | Extra |
+-------+--------------+------+-----+---------+-------+
| ID    | bigint(20)   | NO   | PRI | NULL    |       |
| VALUE | varchar(255) | YES  |     | NULL    |       |
+-------+--------------+------+-----+---------+-------+
```
MySQL with JSON type support:
```
mysql> desc JSONENTITY;
+-------+------------+------+-----+---------+-------+
| Field | Type       | Null | Key | Default | Extra |
+-------+------------+------+-----+---------+-------+
| ID    | bigint(20) | NO   | PRI | NULL    |       |
| VALUE | json       | YES  |     | NULL    |       |
+-------+------------+------+-----+---------+-------+
```

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2022 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2018 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -26,10 +26,11 @@
 //       - 389090: JPA 2.1 DDL Generation Support
 //     02/19/2015 - Rick Curtis
 //       - 458877 : Add national character support
+//     13/01/2022-4.0.0 Tomas Kraus
+//       - 1391: JSON support in JPA
 package org.eclipse.persistence.internal.databaseaccess;
 
 // javase imports
-import static org.eclipse.persistence.internal.helper.DatabaseField.NULL_SQL_TYPE;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -68,13 +69,14 @@ import org.eclipse.persistence.internal.sessions.AbstractSession;
 import org.eclipse.persistence.internal.sessions.ArrayRecord;
 import org.eclipse.persistence.logging.SessionLog;
 import org.eclipse.persistence.mappings.structures.ObjectRelationalDataTypeDescriptor;
-// EclipseLink imports
 import org.eclipse.persistence.queries.Call;
 import org.eclipse.persistence.queries.DatabaseQuery;
 import org.eclipse.persistence.sessions.DatabaseLogin;
 import org.eclipse.persistence.sessions.DatabaseRecord;
 import org.eclipse.persistence.sessions.Login;
 import org.eclipse.persistence.sessions.SessionProfiler;
+
+import static org.eclipse.persistence.internal.helper.DatabaseField.NULL_SQL_TYPE;
 
 /**
  * INTERNAL:
@@ -1354,7 +1356,12 @@ public class DatabaseAccessor extends DatasourceAccessor {
                 value = Helper.rightTrimString((String) value);
             }
             return value;
-        }else if (type == Types.VARCHAR || type == Types.CHAR || type == Types.NVARCHAR || type == Types.NCHAR) {
+            // } else if (fieldType == JsonValue.class || fieldType == JsonObject.class || fieldType == JsonArray.class) {
+            // - workaround without jakarta.json dependency
+        } else if (fieldType != null && fieldType.getName().contains("json")) {
+            // JSON types have platform specific ResultSet handlers.
+            return platform.getJsonPlatform().getJsonDataFromResultSet(resultSet, columnNumber, Object.class);
+        } else if (type == Types.VARCHAR || type == Types.CHAR || type == Types.NVARCHAR || type == Types.NCHAR) {
             // CUSTOM PATCH for oracle drivers because they don't respond to getObject() when using scrolling result sets.
             // Chars may require blanks to be trimmed.
             value = resultSet.getString(columnNumber);
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -32,9 +32,12 @@
 //       - 540929 : 'jdbc.sql-cast' property does not copy
 //     12/06/2018 - Will Dazey
 //       - 542491: Add new 'eclipselink.jdbc.force-bind-parameters' property to force enable binding
+//     13/01/2022-4.0.0 Tomas Kraus
+//       - 1391: JSON support in JPA
 package org.eclipse.persistence.internal.databaseaccess;
 
 // javase imports
+
 import java.io.ByteArrayInputStream;
 import java.io.CharArrayReader;
 import java.io.IOException;
@@ -63,7 +66,6 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.Vector;
 
-// EclipseLink imports
 import org.eclipse.persistence.descriptors.ClassDescriptor;
 import org.eclipse.persistence.exceptions.DatabaseException;
 import org.eclipse.persistence.exceptions.ValidationException;
@@ -271,6 +273,10 @@ public class DatabasePlatform extends DatasourcePlatform {
     protected Boolean useJDBCStoredProcedureSyntax;
     protected String driverName;
 
+    // DatabaseJsonPlatform has lazy initialization
+    /** JSON support for ResultSet data retrieval. */
+    private volatile DatabaseJsonPlatform jsonPlatform;
+
     /**
      * Creates an instance of default database platform.
      */
@@ -299,6 +305,7 @@ public class DatabasePlatform extends DatasourcePlatform {
         this.endDelimiter = "\"";
         this.useJDBCStoredProcedureSyntax = null;
         this.storedProcedureTerminationToken = ";";
+        this.jsonPlatform = null;
     }
 
     /**
@@ -772,6 +779,8 @@ public class DatabasePlatform extends DatasourcePlatform {
         fieldTypeMapping.put(java.time.LocalTime.class, new FieldTypeDefinition("TIME"));
         fieldTypeMapping.put(java.time.OffsetDateTime.class, new FieldTypeDefinition("TIMESTAMP"));
         fieldTypeMapping.put(java.time.OffsetTime.class, new FieldTypeDefinition("TIME"));
+        // Mapping for JSON type.
+        getJsonPlatform().updateFieldTypes(fieldTypeMapping);
 
         return fieldTypeMapping;
     }
@@ -3812,6 +3821,25 @@ public class DatabasePlatform extends DatasourcePlatform {
             return true;
         } catch (Exception notFound) {
             return false;
+        }
+    }
+
+    // Eager initialization in constructor causes CORBA Extension tests to fail.
+    /**
+     * Get JSON support extension instance.
+     * This instance is initialized lazily with 1st JSON support request.
+     *
+     * @return JSON support extension instance
+     */
+    public DatabaseJsonPlatform getJsonPlatform() {
+        if (jsonPlatform != null) {
+            return jsonPlatform;
+        }
+        synchronized (this) {
+            if (jsonPlatform == null) {
+                jsonPlatform = JsonPlatformManager.getInstance().createPlatform(this.getClass());
+            }
+            return jsonPlatform;
         }
     }
 
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
@@ -17,6 +17,8 @@
 //       - 445546: NullPointerException thrown when an Array of Bytes contains null values
 //     05/11/2020-2.7.0 Jody Grassel
 //       - 538296: Wrong month is returned if OffsetDateTime is used in JPA 2.2 code
+//     13/01/2022-4.0.0 Tomas Kraus
+//       - 1391: JSON support in JPA
 package org.eclipse.persistence.internal.helper;
 
 import java.io.ByteArrayOutputStream;
@@ -170,8 +172,10 @@ public class ConversionManager extends CoreConversionManager implements Serializ
                 return null;
             }
         }
-
-        if ((sourceObject.getClass() == javaClass) || (javaClass == null) || (javaClass == ClassConstants.OBJECT) || (javaClass == ClassConstants.BLOB) || (javaClass == ClassConstants.CLOB)) {
+        if (sourceObject.getClass() == javaClass || javaClass == null || javaClass == ClassConstants.OBJECT
+                || javaClass == ClassConstants.BLOB || javaClass == ClassConstants.CLOB
+                // JSON has its own default converter registered. Direct jakarta.json class reference can't be used in core.
+                || javaClass.getName().contains("json")) {
             return (T) sourceObject;
         }
 
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     13/01/2022-4.0.0 Tomas Kraus - 1391: JSON support in JPA
+package org.eclipse.persistence.platform.database;
+
+import java.sql.CallableStatement;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.Hashtable;
+import java.util.Map;
+
+import org.eclipse.persistence.internal.databaseaccess.FieldTypeDefinition;
+import org.eclipse.persistence.internal.sessions.AbstractSession;
+import org.eclipse.persistence.logging.DefaultSessionLog;
+import org.eclipse.persistence.logging.SessionLog;
+
+/**
+ * Postgres 10 database platform extension.
+ * <p>
+ * <b>Purpose</b>: Provides Postgres 10 specific behavior.
+ * <p>
+ * <b>Responsibilities</b>:
+ * <ul>
+ * <li>Native JSON support added in version 10.
+ * </ul>
+ * This class requires Postgres JDBC driver on the classpath.
+ */
+public class PostgreSQL10Platform extends PostgreSQLPlatform {
+
+    /**
+     * Add extended JSON functionality dependent on PostgreSQL JDBC driver.
+     */
+    public interface PostgreSQL10JsonExtension {
+        /**
+         * Check whether provided instance is an instance of {@code PGobject}.
+         *
+         * @param parameter an instance to check
+         * @return value of {@code true} when provided instance is an instance
+         *         of {@code PGobject} or {@code false} otherwise
+         */
+        boolean isPgObjectInstance(final Object parameter);
+    }
+
+    // PostgreSQL10JsonExtension implementation: PostgreSQL10JsonPlatform instance if available or null
+    private final PostgreSQL10JsonExtension postgreSQL10JsonExtension;
+
+    /**
+     * Creates an instance of Postgres 10 platform.
+     */
+    public PostgreSQL10Platform() {
+        super();
+        // Eager PostgreSQL10JsonExtension initialization from Postgres 10 specific platform
+        // does not break the CORBA Extension tests.
+        if (this.getJsonPlatform() instanceof PostgreSQL10JsonExtension) {
+            postgreSQL10JsonExtension = (PostgreSQL10JsonExtension) this.getJsonPlatform();
+        // Missing PostgreSQL10JsonPlatform from org.eclipse.persistence.pgsql module.
+        // This will cause JSON related functionality to fail.
+        } else {
+            DefaultSessionLog.getLog().log(SessionLog.SEVERE, "pgsql10_platform_missing_json_extension", null);
+            postgreSQL10JsonExtension = null;
+        }
+    }
+
+    /**
+     * Build the mapping of database types to class types for the schema framework.
+     *
+     * @return database types to class types {@code Map} for the schema framework
+     */
+    @Override
+    protected Map<String, Class<?>> buildClassTypes() {
+        final Map<String, Class<?>> classTypeMapping = super.buildClassTypes();
+        // Mapping for JSON type.
+        getJsonPlatform().updateClassTypes(classTypeMapping);
+        return classTypeMapping;
+    }
+
+    /**
+     * Build the mapping of class types to database types for the schema framework.
+     *
+     * @return {@code Hashtable} mapping class types to database types for the schema framework
+     */
+    @Override
+    protected Hashtable<Class<?>, FieldTypeDefinition> buildFieldTypes() {
+        final Hashtable<Class<?>, FieldTypeDefinition>fieldTypeMapping = super.buildFieldTypes();
+        // Mapping for JSON type.
+        getJsonPlatform().updateFieldTypes(fieldTypeMapping);
+        return fieldTypeMapping;
+    }
+
+    /**
+     * INTERNAL
+     * Set the parameter in the JDBC statement at the given index.
+     * This support a wide range of different parameter types, and is heavily optimized for common types.
+     * Handles Postgres specific PGobject instances.
+     *
+     * @param parameter the parameter to set
+     * @param statement target {@code PreparedStatement} instance
+     * @param index index of the parameter in the statement
+     * @param session current database session
+     */
+    @Override
+    public void setParameterValueInDatabaseCall(
+            final Object parameter, final PreparedStatement statement,
+            final int index, final AbstractSession session
+    ) throws SQLException {
+        if (postgreSQL10JsonExtension != null && postgreSQL10JsonExtension.isPgObjectInstance(parameter)) {
+            statement.setObject(index, parameter);
+        } else {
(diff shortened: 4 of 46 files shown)


Tomas-Kraus:
Let's get this in. I'll continue with MySQL native JSON type support as next PR.

Tomas-Kraus:
@lukasj Don't push it in. I'm working on some extension to delegate few things to DB platforms and implementing native JSON type on MySQL.

rfelcman:
Hmm maybe Oracle DB too.
https://docs.oracle.com/en/database/oracle/oracle-database/21/jajdb/oracle/sql/json/package-summary.html
Or I can implement it later. It seems, that native support started from 20c or 21c.

Tomas-Kraus:
@rfelcman Good JSON support is since 21c. But you need latest JDBC for that. MySQL is a good one to start with. Everything is there in 5.7 and I already have local environment working.

Tomas-Kraus:
I just found out that this design with default Converter is a dead end. Will redesign it using conversion in `ConversionManager#convertObject`.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
```diff
@@ -1354,7 +1355,10 @@ protected Object getObjectThroughOptimizedDataConversion(ResultSet resultSet, Da
                 value = Helper.rightTrimString((String) value);
             }
             return value;
-        }else if (type == Types.VARCHAR || type == Types.CHAR || type == Types.NVARCHAR || type == Types.NCHAR) {
+        } if (fieldType == ClassConstants.JSON_VALUE || fieldType == ClassConstants.JSON_ARRAY || fieldType == ClassConstants.JSON_OBJECT) {
```
lukasj, line 1358: `else if` instead?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -3767,4 +3774,47 @@ public boolean checkTableExists(final DatabaseSessionImpl session,
...
+
+    /**
+     * Retrieve JSON data from JDBC ResultSet.
+     *
+     * @param resultSet source JDBC ResultSet
+     * @param columnNumber index of column in JDBC ResultSet
+     * @return JSON data from JDBC ResultSet
+     * @throws SQLException if data could not be retrieved
+     */
+    public Object getJsonDataFromResultSet(ResultSet resultSet, int columnNumber) throws SQLException {
```
lukasj, line 3816: I do not think these methods belongs here. Isn't there better/different way? How is the same problem being solved for other datatypes from external libraries like Oracle specific datatypes?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ClassConstants.java
```diff
@@ -157,6 +163,10 @@
     //XML Classes
     public static final Class<Document> DOCUMENT = Document.class;
 
+    // JSON Classes
+    public static final Class<JsonValue> JSON_VALUE = JsonValue.class;
+    public static final Class<JsonArray> JSON_ARRAY = JsonArray.class;
+    public static final Class<JsonObject> JSON_OBJECT = JsonObject.class;
```
lukasj, line 169: this is sort of dangerous thing to do - even usage of the logger will require json api on the classpath. Is this what we really want?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -169,8 +171,10 @@ public Object clone() {
                 return null;
             }
         }
-
-        if ((sourceObject.getClass() == javaClass) || (javaClass == null) || (javaClass == ClassConstants.OBJECT) || (javaClass == ClassConstants.BLOB) || (javaClass == ClassConstants.CLOB)) {
+        if (sourceObject.getClass() == javaClass || javaClass == null || javaClass == ClassConstants.OBJECT
+                || javaClass == ClassConstants.BLOB || javaClass == ClassConstants.CLOB
+                // JSON has its own default converter registered
+                || javaClass == ClassConstants.JSON_VALUE || javaClass == ClassConstants.JSON_ARRAY || javaClass == ClassConstants.JSON_OBJECT) {
             return (T) sourceObject;
```
lukasj, line 178: this is the place where the conversion should happen with no need for additional methods in the dbplatform

Tomas-Kraus: Well, that's not that easy. Conversion will be platform specific for some databases, e.g. Oracle 21c:
https://docs.oracle.com/en/database/oracle/oracle-database/21/jajdb/oracle/sql/json/package-summary.html
```
OracleJsonObject pear = rs.getObject(1, OracleJsonObject.class);
```
All Oracle specific JSON types implement javax.json.JsonValue interface so I have to overwrite both - ResultSet retrieval and conversion code in new Oracle21cPlatform class.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: Well, this has to be in platform, see this in https://docs.oracle.com/en/database/oracle/oracle-database/21/jajdb/oracle/sql/json/package-summary.html
```
OracleJsonObject pear = rs.getObject(1, OracleJsonObject.class);
```
I have to use Oracle 21c specific call using platform specific data type (`oracle.sql.json.OracleJsonValue`). This must be in Platform clas, not in common EclipseLink code.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java
Tomas-Kraus: It's a typo. Fixed.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ClassConstants.java
Tomas-Kraus: I know. It caused a lot of build issues and `pom.xml` changes. But this was also the _systematic_ way of extending the code.
I still have to keep jakarta.json references in ConversionManager and few other classes. But I can remove them from ClassConstants and use direct JsonValue/Array/Object.class references in the code.
Will also go trough the build/tests and remove jakarta.json from poms where possible.

Tomas-Kraus: Bad news is, that I was able to remove jakarta.json only from OSGITestHelper. All new POMs dependencies are still required.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,198 @@
...
+     * JSON data retrieved from Postgres JDBC {@code ResultSet} are returned as {@code PGobject} instance.
+     * It must be converted to {@code String} first to be accepted by common {@code JsonTypeConverter}.
+     *
+     * @param resultSet source JDBC {@code ResultSet}
+     * @param columnNumber index of column in JDBC {@code ResultSet}
+     * @return JSON data from JDBC {@code ResultSet} as {@code String} to be parsed by common {@code JsonTypeConverter}
+     * @throws SQLException if data could not be retrieved
+     */
+    @Override
+    public Object getJsonDataFromResultSet(ResultSet resultSet, int columnNumber) throws SQLException {
```
Tomas-Kraus, line 178: This method working with `PGobject` would be nice without reflection. But I'm wondering that it may be performance regression compared to default code as it is now.
Default getJsonDataFromResultSet code with `resultSet.getString(columnNumber);` works too with Postgres so maybe I can delete this method.

Tomas-Kraus:
PostgreSQL change set broke the tests. Removed it from PR.

Tomas-Kraus:
PostgreSQL platform issues are resolved now.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
Tomas-Kraus: Also Postgres10+ platform has platform specific code to read JSON from ResultSet.

foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
```diff
@@ -0,0 +1,198 @@
...
+     * @param jsonValue source JSON value field
+     * @return converted JDBC statement type
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> T convertJsonValueToDataValue(final JsonValue jsonValue) throws PersistenceException {
+        if (jsonValue == null) {
+            return null;
+        }
+        final StringWriter sw = new StringWriter(128);
```
Tomas-Kraus, line 151: It would be great to have create methods with jakarta.json.JsonValue/Object/Array on OracleJsonFactory.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java
```diff
@@ -169,8 +175,10 @@ public Object clone() {
                 return null;
             }
         }
-
-        if ((sourceObject.getClass() == javaClass) || (javaClass == null) || (javaClass == ClassConstants.OBJECT) || (javaClass == ClassConstants.BLOB) || (javaClass == ClassConstants.CLOB)) {
+        if (sourceObject.getClass() == javaClass || javaClass == null || javaClass == ClassConstants.OBJECT
+                || javaClass == ClassConstants.BLOB || javaClass == ClassConstants.CLOB
+                // JSON has its own default converter registered
+                || javaClass == JsonValue.class || javaClass == JsonObject.class || javaClass == JsonArray.class) {
```
rfelcman, line 181: I think, that `JsonValue, JsonObject, JsonArray` should be described in `ClassConstants` to be consistent with other types.

Tomas-Kraus: I already did it before, but Lukas asked me to remove them from `ClassConstants`.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+                throw new PersistenceException("Database PGobject conversion failed.", e);
+            }
+        // Fallback option when String value is returned.
+        } else if (rawData instanceof String) {
+            return rawData;
+        }
+        throw new PersistenceException("Unknown JSON type returned from database");
+    }
+
+}
```
rfelcman, line 242: PostgreSQL looks like one big workaroud due JDBC drive limits.
Any chance to call DB JSON as a String like MySQL platform?

foundation/org.eclipse.persistence.extension/pom.xml
```diff
@@ -54,6 +54,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+            <scope>test</scope>
+        </dependency>
```
rfelcman, line 61: Why is dependency there? I don't see any JSON test in this Maven module.

foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
```diff
@@ -0,0 +1,198 @@
...
+            return null;
+        }
+        if (jdbcValue instanceof OracleJsonValue) {
+        // Depends on javax.json, not jakarta.json!
+        //    return ((OracleJsonValue) jdbcValue).wrap(JsonValue.class);
+            try (final JsonReader jr = Json.createReader(new StringReader(jdbcValue.toString()))) {
+                return jr.readValue();
+            }
+        }
+        throw new PersistenceException("Could not convert JDBC ResultSet type to JsonValue");
```
rfelcman, line 180: What about resource bundle usage like e.g.
`throw new PersistenceException(ExceptionLocalization.buildMessage("json_cant_convert", null));`

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/converters/JsonTypeConverter.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param jsonValue source JSON field value
+     * @param session current database session
+     * @return target String to be stored as JDBC VARCHAR
+     */
+    @Override
+    public Object convertObjectValueToDataValue(Object jsonValue, Session session) {
+        if (jsonValue instanceof JsonValue) {
+            return session.getPlatform().convertJsonValueToDataValue((JsonValue)jsonValue);
+        }
+        throw new IllegalArgumentException("Source object is not an instance of JsonValue");
```
rfelcman, line 49: What about resource bundle usage like e.g.
`throw new IllegalArgumentException(ExceptionLocalization.buildMessage("json_.....", null));`

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+                this.constructor = null;
+                this.setValue = null;
+                this.setType = null;
+                this.getValue = null;
+            }
+        }
+
+        private boolean isInstance(final Object o) {
+            if (classReference == null) {
+                throw new PersistenceException("Missing " + this.className + " on classpath");
```
rfelcman, line 83: Exception resource bundle.....

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+            }
+            return classReference.isInstance(o);
+        }
+
+        private Object newPgObject(final String value)
+                throws InvocationTargetException, IllegalAccessException, InstantiationException {
+            if (classReference == null) {
+                throw new PersistenceException("Missing " + this.className + " on classpath");
```
rfelcman, line 91: Exception resource bundle.....

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+    public <T> T convertJsonValueToDataValue(final JsonValue jsonValue) throws PersistenceException {
+        if (jsonValue == null) {
+            return null;
+        }
+        final String jsonAsString = super.convertJsonValueToDataValue(jsonValue);
+        // Following code is called through reflection to avoid PGobject dependency
+        try {
+            return (T) PgObjectAccessor.PG_OBJECT_ACCESSOR.newPgObject(jsonAsString);
+        } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            throw new PersistenceException("JSON value to database type conversion failed.", e);
```
rfelcman, line 209: Exception resource bundle.....

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+    @Override
+    public Object getJsonDataFromResultSet(final ResultSet resultSet, final int columnNumber) throws SQLException {
+        // ResultSet returns an instance of PGobject.
+        final Object rawData = resultSet.getObject(columnNumber);
+        // Following code is called through reflection to avoid PGobject dependency.
+        if (PgObjectAccessor.PG_OBJECT_ACCESSOR.isInstance(rawData)) {
+            try {
+                return PgObjectAccessor.PG_OBJECT_ACCESSOR.getValue(rawData);
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                throw new PersistenceException("Database PGobject conversion failed.", e);
```
rfelcman, line 233: Exception resource bundle.....

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
```diff
@@ -0,0 +1,242 @@
...
+            }
+        // Fallback option when String value is returned.
+        } else if (rawData instanceof String) {
+            return rawData;
+        }
+        throw new PersistenceException("Unknown JSON type returned from database");
```
rfelcman, line 239: Exception resource bundle.....

foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
```diff
@@ -0,0 +1,198 @@
...
+            return null;
+        }
+        final StringWriter sw = new StringWriter(128);
+        try (final JsonWriter jw = Json.createWriter(sw)) {
+            jw.write(jsonValue);
+        }
+        try (final Reader jr = new StringReader(sw.toString())) {
+            return (T) factory.createJsonTextValue(jr);
+        } catch (IOException e) {
+            throw new PersistenceException("Could not convert JsonValue to OracleJsonValue", e);
```
rfelcman, line 158: Exception resource bundle.....

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java
```diff
@@ -0,0 +1,225 @@
...
+            Assert.assertTrue(usedToEntityAttribute);
+        } finally {
+            if (em.getTransaction().isActive()) {
+                em.getTransaction().rollback();
+            }
+            em.close();
+        }
+    }
+
+}
```
rfelcman, line 225: Is it possible to add there test like `SELECT * FROM table WHERE json_column = :json_parameter`?

rfelcman (commented):
Is it possible to add there test into `foundation/eclipselink.core.test/src/it/java/org/eclipse/persistence/testing/tests/types/TypeTestModel.java` like #1418.

Tomas-Kraus:
> Is it possible to add there test into `foundation/eclipselink.core.test/src/it/java/org/eclipse/persistence/testing/tests/types/TypeTestModel.java` like #1418.

Why into core? I made set of JPA tests where implementation is much more simple.

foundation/org.eclipse.persistence.corba/pom.xml
```diff
@@ -103,6 +103,11 @@
             <version>1.3.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
Tomas-Kraus: [ERROR] Errors: 
[ERROR]   CorbaTestModel>TestCollection.run:330 » NoClassDefFound jakarta/json/JsonValue
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -3767,4 +3778,54 @@ public boolean checkTableExists(final DatabaseSessionImpl session,
...
+     * Convert JDBC {@code ResultSet} type to JSON value field.
+     *
+     * @param jdbcValue source classification type value from JDBC
+     * @return converted JSON field value
+     */
+    public JsonValue convertDataValueToJsonValue(Object jdbcValue) {
+        if (jdbcValue == null) {
+            return null;
+        }
+        try (final JsonReader jr = Json.createReader(new StringReader((String)jdbcValue))) {
```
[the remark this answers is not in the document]
Tomas-Kraus: Method `public JsonValue convertDataValueToJsonValue(Object jdbcValue)` uses value returned by `Object getJsonDataFromResultSet(ResultSet resultSet, int columnNumber)` which returns `resultSet.getString(columnNumber)`
So yes, default implementation works with JSON as String values.
Platform dependent code may relay on this, like in `PostgreSQL10Platform` where only `getJsonDataFromResultSet` is overwritten and still returns String, or implement it's own types by owerwriting both methods like in `Oracle21Platform`.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
Tomas-Kraus: Unfortunately not. At least `public <T> T convertJsonValueToDataValue(final JsonValue jsonValue) throws PersistenceException` with all those reflection calls is a must because `PGobject` must be set into statement.
Theoretically I can  remove `public Object getJsonDataFromResultSet(final ResultSet resultSet, final int columnNumber) throws SQLException` because default code from DatabasePlatform superclass works too. But I'd like to have Statement set and ResultSet get code consistent.

foundation/org.eclipse.persistence.extension/pom.xml
Tomas-Kraus: [ERROR] Errors: 
[ERROR]   SLF4JLoggerTest.testGetLevel » NoClassDefFound jakarta/json/JsonValue
[ERROR]   SLF4JLoggerTest.testLogExceptionStackTrace » NoClassDefFound jakarta/json/Json...
[ERROR]   SLF4JLoggerTest.testLogExceptionWithoutStackTrace » NoClassDefFound jakarta/js...
[ERROR]   SLF4JLoggerTest.testLogMessage » NoClassDefFound jakarta/json/JsonValue
[INFO] 
[ERROR] Tests run: 4, Failures: 0, Errors: 4, Skipped: 0

foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
Tomas-Kraus: OK, I'm changing this. But I don't believe it makes sense in EE4J days. Did you ever saw any localization for ExceptionLocalizationResource?
Also messages bundle is in core, but it's being used everywhere (oracle extension in this case).

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
Tomas-Kraus: Fixed.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
Tomas-Kraus: Fixed.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java
Tomas-Kraus: Fixed.

foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
Tomas-Kraus: Fixed.

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java
Tomas-Kraus: Added this test. Everything except MySQL native `JSON` type usage passed.

Tomas-Kraus:
Added jakarta.json dependency removal from core. Rebased on top of the master.  Still need to do some final cleanup in POMs.

Tomas-Kraus:
It's ready for review.

lukasj (approved):
LGTM

guymac:
When using annotations, @Convert seems to be required, otherwise there is an exception. Not sure if that is a bug however, just thought I'd point it out.

E.g.
```
    // this works
    @Column(name = "JSON_DATA")
    @Convert
    private JsonValue jsonData;
```    
But
```
    // this throws an exception
    @Column(name = "JSON_DATA")
    private JsonValue jsonData;
```    

> Internal Exception: jakarta.persistence.PersistenceException: Exception [EclipseLink-28018] (Eclipse Persistence Services - 4.0.0-RC2.v202209091122): org.eclipse.persistence.exceptions.EntityManagerSetupException
> Exception Description: Predeployment of PersistenceUnit [JPA] failed.
> Internal Exception: java.lang.IllegalStateException: Missing JSON type converter on classpath
> 	at org.eclipse.persistence.exceptions.PersistenceUnitLoadingException.exceptionSearchingForPersistenceResources(PersistenceUnitLoadingException.java:129)
>     
>

PR #1390, opened 2022-01-13
[master] Cannot set FOR UPDATE in conjunction with ORDER BY and / or LIMIT (OracleDB & Derby) - bugfix + unit test

### LINKED ISSUE #1369: Cannot set FOR UPDATE in conjunction with ORDER BY and / or LIMIT (OracleDB & Derby)
Running against OracleDB or Derby, if ORDER BY is set on CriteriaQuery, and / or setMaxResults is called on TypedQuery object, then a lock instruction applied using

```
final TypedQuery<Long> query = em.createQuery(cq);
query.setLockMode(LockModeType.PESSIMISTIC_WRITE);
```
results in syntax error on OracleDB, while against Derby the instruction appears to be ignored. Running against SQLServer this works as expected.

Using `query.unwrap(JpaQuery.class).getDatabaseQuery().getSQLString()` to obtain the resultant SQL statement.

Running the following:
```
            cq.orderBy( cb.asc( jpaTable.get(JPATABLE_.getId()) ) );
            final TypedQuery<Long> query = em.createQuery(cq);
            query.setMaxResults(pageSize);
            query.setLockMode(LockModeType.PESSIMISTIC_WRITE);
            final List<Long> ids = query.getResultList();
```
Although the row limit doesn't appear to be applied in the SQL statement, it is enforced somehow, because we do get correct result: `SELECT ID FROM JPATABLE WITH (UPDLOCK) WHERE (TENANT = ?) ORDER BY ID ASC`

Against Derby, the FOR UPDATE instruction seems to be ignored - `SELECT ID AS a1 FROM JPATABLE WHERE (TENANT = ?) ORDER BY ID ASC OFFSET ? ROWS FETCH NEXT ? ROWS ONLY`

Against OracleDB, this results in an exception:
    _javax.persistence.PessimisticLockException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.8.v20201217-ecdf3c32c4): org.eclipse.persistence.exceptions.DatabaseException
    Internal Exception: java.sql.SQLSyntaxErrorException: ORA-00933: SQL command not properly ended
    Error Code: 933_

Running a query with row limit and ORDER BY, but with no locking specified works correct against all DB types. Running a query with locking specified, and without ORDER BY and row limit also works fine.

rfelcman:
Fixes #1369 .

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
@@ -1722,9 +1722,9 @@ public class SQLSelectStatement extends SQLStatement {
     /**
      * Print the SQL representation of the statement on a stream.
      */
-    public Vector printSQL(ExpressionSQLPrinter printer) {
+    public Vector<DatabaseField> printSQL(ExpressionSQLPrinter printer) {
         try {
-            Vector selectFields = null;
+            Vector<DatabaseField> selectFields = null;
             selectFields = printSQLSelect(printer);
             printSQLWhereKeyWord(printer);
             printSQLWhereClause(printer);
@@ -1740,8 +1740,8 @@ public class SQLSelectStatement extends SQLStatement {
         }
     }
 
-    public Vector printSQLSelect(ExpressionSQLPrinter printer) throws IOException {
-        Vector selectFields = null;
+    public Vector<DatabaseField> printSQLSelect(ExpressionSQLPrinter printer) throws IOException {
+        Vector<DatabaseField> selectFields = null;
         printer.setRequiresDistinct(shouldDistinctBeUsed());
 
         if (hasUnionExpressions()) {
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2022 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


rfelcman:
There are tests results CORE, JPA, JPASE against Oracle 12c.
[test-reports.tar.gz](https://github.com/eclipse-ee4j/eclipselink/files/7861837/test-reports.tar.gz)

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
```diff
@@ -1722,69 +1722,102 @@ protected void normalizeOrderBy(Expression builder, List<Expression> allExpressi
     /**
      * Print the SQL representation of the statement on a stream.
      */
-    public Vector<DatabaseField> printSQL(ExpressionSQLPrinter printer) {
+    public Vector printSQL(ExpressionSQLPrinter printer) {
```
lukasj, line 1725: why generics were removed?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/OraclePlatform.java
```diff
@@ -951,14 +951,16 @@ public boolean useJDBCStoredProcedureSyntax() {
     protected String MAX_ROW = "WHERE ROWNUM <= ";
     protected String MIN_ROW = ") WHERE rnum > ";
     // Bug #453208
-    protected String LOCK_START_PREFIX = " AND (";
+    protected String LOCK_START_PREFIX_AND = " AND (";
```
lukasj, line 954: is this rename required? Keep in mind that during the backport, one will have to keep the original name (compatibility)

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
```diff
@@ -1722,69 +1722,102 @@ protected void normalizeOrderBy(Expression builder, List<Expression> allExpressi
...
 
-            if (hasGroupByExpressions()) {
-                appendGroupByClauseToWriter(printer);
-            }
-            if (hasHavingExpression()) {
-                //appendHavingClauseToWriter(printer);
-                printer.printString(" HAVING ");
-                printer.printExpression(getHavingExpression());
-            }
+    public void printSQLWhereKeyWord(ExpressionSQLPrinter printer) throws IOException {
```
lukasj, line 1773: do these new methods have to be public instead of private/protected?

rfelcman: Yes because they are called from `org.eclipse.persistence.platform.database.OraclePlatform` (different package).

lukasj (approved):
LGTM

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
lukasj: ok

Issue #1398, opened 2022-01-19
Bug 578262: Update queries reuse the same version locking value


dazey3:
Linking to [BZ #578262](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578262)

EclipseLink creates an EJBQuery, prepares that EJBQuery, and then resuses that query whenever another query is created with a matching string. During prepare, EJBQuery obtains the version locking value (assuming the entity uses version locking) and stores that version locking value in the update clauses! This is the issue. Since the query is only prepared once (even if the same query is created again later) then the first time it was prepared, the version locking value is obtained and it then never changes again since the query is never re-prepared.

```java
@Entity
public class EcallRegistration {
    @Id private String pk;
    private int ecallAvailableFlag;
    @Version @Column(name = "sys_update_timestamp")
    private Timestamp sysUpdateTimestamp;
}
```

--- Test ---

```java
    EntityManager em = emf.createEntityManager();
    try {
        int flag = 0;
        System.out.println("****CREATING QUERY 1 ****");
        Query query = em.createQuery("UPDATE EcallRegistration e SET e.ecallAvailableFlag = :flag", EcallRegistration.class);
        query.setParameter("flag", flag);

        System.out.println("****EXECUTING QUERY 1 ****");
        query.executeUpdate();
    } finally {
        if(em.isOpen()) {
            em.clear();
            em.close();
        }
    }
...
    Thread.sleep(10000);
...
    em = emf.createEntityManager();
    try {
        int flag = 1;
        System.out.println("****CREATING QUERY 2 ****");
        Query query = em.createQuery("UPDATE EcallRegistration e SET e.ecallAvailableFlag = :flag", EcallRegistration.class);
        query.setParameter("flag", flag);

        System.out.println("****EXECUTING QUERY 2 ****");
        query.executeUpdate();
    } finally {
        if(em.isOpen()) {
            em.clear();
            em.close();
        }
    }
```

Output:
```
****CREATING QUERY 1 ****
[EL Finest]: query: 2022-01-18 10:37:10.884--UnitOfWork(443713699)--Thread(Thread[main,5,main])--Execute query ValueReadQuery(sql="SELECT NOW()")
[EL Fine]: sql: 2022-01-18 10:37:10.884--ClientSession(1563053805)--Connection(1082411691)--Thread(Thread[main,5,main])--SELECT NOW()

****EXECUTING QUERY 1 ****
[...]

PR #1402, opened 2022-01-19
[master] Bug 578262: Bulk update queries reuse the same version locking value

dazey3:
for #1398

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/TimestampLockingPolicy.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/TimestampLockingPolicy.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2022 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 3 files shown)


dazey3:
The fix for this issue changes the query behavior for bulk updates with timestamp version locking values

Current:
```
****CREATING QUERY****
Execute query ValueReadQuery(sql="SELECT NOW()")
****EXECUTING QUERY****
Execute query UpdateAllQuery(referenceClass=EcallRegistration sql="UPDATE ECALLREGISTRATION SET ECALLAVAILABLEFLAG = ?, sys_update_timestamp = ? WHERE (PK = ?)")
```
Currently, EclipseLink executes two separate queries: one during query prepare that gets a TimeStamp value from the database and one during query execute that uses the value obtained during create/prepare

Changed:
```
****CREATING QUERY****
****EXECUTING QUERY****
Execute query UpdateAllQuery(referenceClass=EcallRegistration sql="UPDATE ECALLREGISTRATION SET ECALLAVAILABLEFLAG = ?, sys_update_timestamp = CURRENT_TIMESTAMP WHERE (PK = ?)")
```
With this change, create/prepare for UpdateAllQuery no longer executes a separate query to obtain timestamp values. Instead, create/prepare now uses CURRENT_TIMESTAMP(`ExpressionOperator.Today`), which is defined in the JPA spec and already supported by existing platforms, to add the platform specific CURRENT_TIMESTAMP function to the query. This should be fine for bulk updates as the specification doesn't expect updates (and EclipseLink doesnt currently do so) as stated in section 4.10

lukasj (commented):
can this be rebased on master, please?

dazey3:
Rebased and tests rerunning. I assume this is so that the new tests delivered under https://github.com/eclipse-ee4j/eclipselink/pull/1403 are included?

lukasj:
> I assume this is so that the new tests delivered under #1403 are included?

it's about getting rid of unrelated changes to make the code easier to review

lukasj (approved):
LGTM

PR #1404, opened 2022-01-20
[master] JPA API and Parsson dependencies upgrade

rfelcman:

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/tools/weaving/jpa/StaticWeaveProcessor.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/tools/weaving/jpa/StaticWeaveProcessor.java
@@ -29,8 +29,6 @@ import java.util.Iterator;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 
-import jakarta.persistence.spi.TransformerException;
-
 import org.eclipse.persistence.exceptions.StaticWeaveException;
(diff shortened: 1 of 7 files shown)


lukasj (approved):
LGTM

rfelcman:
It seems, that there is some incompatibility between
`jakarta.persistence.spi.ClassTransformer#transform(....)`
and 
`java.lang.instrument.ClassFileTransformer#transform()`
due change of possible exception `IllegalClassFormatException -> TransformerException`

It's presented by https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/deployment/JavaSECMPInitializer.java#L258
compilation error. 
Maybe `jakarta.persistence.spi.TransformerException` should extend `java.lang.instrument.IllegalClassFormatException`?

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/tools/weaving/jpa/StaticWeaveProcessor.java
```diff
@@ -318,7 +319,7 @@ private void process() throws IOException,URISyntaxException{
                         } else {
                             swoh.addEntry(entryInputStream, newEntry);
                         }
-                    } catch (IllegalClassFormatException e) {
+                    } catch (TransformerException e) {
```
lukasj, line 322: this can be `Throwable` and the other catch below can be removed

lukasj (approved):
LGTM

PR #1438, opened 2022-02-21
Modified getLogString with more NULL Safety

rostup:
The old implementation causes a NPE if the `session` or the `project` is null in line 1102.  This new implementation of will not.

The session being NULL is plausible, because the `appendLogParameters` function, which is called in line 1111, starts with a NULL check.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/StoredProcedureCall.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/StoredProcedureCall.java
@@ -1116,12 +1116,12 @@ public class StoredProcedureCall extends DatabaseCall {
         }
     }
 
-	private boolean isIndexBased(List<String> procedureArgs, AbstractSession session) {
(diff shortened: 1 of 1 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/StoredProcedureCall.java
```diff
@@ -1115,4 +1115,13 @@ public String getLogString(Accessor accessor) {
             return getSQLString();
         }
     }
+
+	private boolean isIndexBased(List<String> procedureArgs, AbstractSession session) {
```
lukasj, line 1119: tabs are not allowed, use spaces

lukasj (commented):
commit has to be signed - did you sign ECA? See [the failure](https://accounts.eclipse.org/legal/eca/validation/94555)

rostup:
I changed the Tabs to 4 Spaces each and signed the ECA.

lukasj (approved):
LGTM

rostup:
THX

PR #1443, opened 2022-03-03
Implement New JPA API 3.1.0 Features

Tomas-Kraus:


lukasj (approved):
LGTM

lukasj:
see https://github.com/eclipse-ee4j/jpa-api/pull/351#discussion_r799518651
see https://github.com/eclipse-ee4j/jpa-api/pull/351#issuecomment-1030029167

dazey3:
@lukasj  Thanks!

I am just investigating the test issues for `TestMathFunctions(test-jpa-jse)` when running on Derby:
```
Exception [EclipseLink-4002] (Eclipse Persistence Services - 4.0.0.v202206081518): org.eclipse.persistence.exceptions.DatabaseException Internal Exception: java.sql.SQLSyntaxErrorException: Syntax error: Encountered "e" at line 1, column 29. Error Code: 30000 Call: SELECT FLOOR((DOUBLEVALUE)*1e?+0.5)/1e? FROM NUMBERENTITY WHERE (ID = ?) bind => [3 parameters bound] Query: ReportQuery(referenceClass=NumberEntity sql="SELECT FLOOR((DOUBLEVALUE)*1e?+0.5)/1e? FROM NUMBERENTITY WHERE (ID = ?)")
```

It seems like this set of math operations is not formed in such a way that supports "?" parameter markers.

dazey3:
@Tomas-Kraus 
```sql
    SELECT FLOOR((DOUBLEVALUE)*1e?+0.5)/1e? FROM NUMBERENTITY WHERE (ID = ?)
```

Can we not change this to:
```sql
    SELECT FLOOR((DOUBLEVALUE) * EXP(?) + 0.5) / EXP(?) FROM NUMBERENTITY WHERE (ID = ?)
```

https://db.apache.org/derby/docs/10.13/ref/rreffuncexp.html

Tomas-Kraus:
@dazey3 There was discussion on this topic in [JPA API isse](https://github.com/eclipse-ee4j/jpa-api/pull/351#issuecomment-1055451419). Problem with EXP(?) is, that it's floating point function and we need FLOOR to produce integer result. The solution is taken is from Hibernate code and it seems to work, but it has some limitations.

Issue #1429, opened 2022-02-11
Bug 578686: ParameterExpression in ORDER BY clause doesn't register


dazey3:
Linking to [BZ #578686](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578686)

Using ParameterExpressions in the CriteriaBuilder orderby clause doesn't register the parameter and causes an exception to be thrown. This might be because ordinal variables aren't supported in the ORDER BY clause for EclipseLink. However, ordinal variables work just fine in equivalent JPQL.

JPQL Test:
```java
    Query query = em.createQuery("SELECT s.intVal1 FROM SimpleEntity s ORDER BY ?1");
    query.setParameter(1, 1);
    query.getResultList();
```

Runs successfully on platforms that support it. DB2 and Derby don't support parameters in the ORDER BY clause... but Oracle and MySQL do support it and run successfully.

Criteria Test:
```java
    CriteriaBuilder cb = em.getCriteriaBuilder();
    CriteriaQuery<Object[]> cquery = cb.createQuery(Object[].class);
    Root<SimpleEntity> root = cquery.from(SimpleEntity.class);
    ParameterExpression<Integer> intParam1 = cb.parameter(Integer.class);
    cquery.multiselect(root.get(SimpleEntity_.intVal1));
    cquery.orderBy(cb.desc(intParam1));

    Query query = em.createQuery(cquery);
    query.setParameter(intParam1, 1);
    query.getResultList();
```

Exception:
```
java.lang.IllegalArgumentException: No parameter with name : Parameter[name=null] was found within the query: ReportQuery(referenceClass=SimpleEntity ).
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:548)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:1)
```

PR #1445, opened 2022-03-03
[2.6_WAS] Bug 578686: Remove OrderBy parameter binding use case

dazey3:
for #1429

For DB2/Derby, it is illegal to include untyped parameter markers in the ORDER BY clause. There are two JPQL scenarios where these untyped parameters can occur:
1) via JPQL "... ORDER BY ?1"
2) via JPQL "... ORDER BY 1" with `shouldBindLiterals=true`

My previous fix for 578686 included a minor fix to allow databases, that do not support untyped parameters, to disable parameter binding in the event that a marker is used within the ORDER BY clause. This fix was flawed and I ultimately decided it would be best to remove completely. 

The issue is that EclipseLink does not have the ability at this time to disable parameter binding per variable instance and instead disables binding for the whole query. The change that I made, and am now removing, altered existing behavior with Constant values and `shouldBindLiterals=true` that I did not intend. I have come to the conclusion that it would be best to continue to allow Derby/DB2 to fail with the 2 scenarios I outlined above until a more fine grained parameter binding strategy can be delivered (I am working on this at the moment)


jgrassel (approved):
lgtm

PR #1446, opened 2022-03-04
Fixed TestJava8DateTime failures.

Tomas-Kraus:
This reverts LocalTime DB type change I made and changes it's ExpressionOperator to be TIMESTAMP too.
Let's wait for tests to be sure everything is OK.


lukasj (approved):
LGTM

PR #1452, opened 2022-03-09
Added missing EXTRACT(QUARTER FROM date/datetime) for Derby platform.

Tomas-Kraus:
All EXTRACT TCK tests added as jUnits.


lukasj (approved):
LGTM

PR #1456, opened 2022-03-18
Changed second fraction precision of LocalDateTime and OffsetDateTime MySQL related type.

Tomas-Kraus:
DATETIME precision modified to DATETIME(6) to strore microsecond fraction.
Some tests modified to be more stable.
Added test with LocalDateTime.now() in WHERE condition. This scenario was failing in TCK.


rfelcman (approved):
LGTM

Issue #1457, opened 2022-03-21
Bug 579327: ParameterExpression in CriteriaBuilder.isNull() doesn't register


dazey3:
Linking to [BZ #579327](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579327)

Passing a ParameterExpression to CriteriaBuilder.isNull() results in an exception when setting the ParameterExpression value on the Query.

Test:
```
    CriteriaBuilder cb = em.getCriteriaBuilder();
    CriteriaQuery<Integer[]> cquery = cb.createQuery(Integer[].class);
    Root<SimpleEntity> root = cquery.from(SimpleEntity.class);
    cquery.multiselect(root.get(SimpleEntity_.intVal1));

    ParameterExpression<String> strParam1 = cb.parameter(String.class);
    cquery.where(cb.isNull(strParam1));

    Query query = em.createQuery(cquery);
    query.setParameter(strParam1, "HELLO");
    query.getResultList();
```

Exception:
```
java.lang.IllegalArgumentException: No parameter with name : Parameter[name=null] was found within the query: ReportQuery(referenceClass=SimpleEntity ).
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:548)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.setParameter(EJBQueryImpl.java:1)
```

rfelcman:
Sorry but, there are some test errors in continues builds (Derby DB) see:
https://ci.eclipse.org/eclipselink/job/eclipselink-cb-master/lastCompletedBuild/testReport/
https://ci.eclipse.org/eclipselink/job/eclipselink-cb-3.0/152/
I think, that in 2.7 too, but JPA.JSE are not executed in continues builds.

dazey3:
@rfelcman 

Thanks for letting me know! It seems that one of my tests is invalid for Derby and I need to account for that.

```
SELECT ITEM_STRING1 FROM SIMPLE_TBL01 WHERE (NULL IS NOT NULL)
```
This query works on Oracle, DB2, SQLServer, Postgres. But it fails on Derby and I missed that. It seems to be illegal.

PR #1458, opened 2022-03-21
[master] Bug 579327: ParameterExpression in CriteriaBuilder.isNull() doesn't register

dazey3:
for #1457

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/querydef/CriteriaBuilderImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/querydef/CriteriaBuilderImpl.java
@@ -82,7 +82,7 @@ public class CriteriaBuilderImpl implements JpaCriteriaBuilder, Serializable {
      */
     @Override
     public CriteriaQuery<Object> createQuery(){
-        return new CriteriaQueryImpl(this.metamodel, ResultType.UNKNOWN, ClassConstants.OBJECT, this);
+        return new CriteriaQueryImpl<Object>(this.metamodel, ResultType.UNKNOWN, ClassConstants.OBJECT, this);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/querydef/CriteriaBuilderImpl.java
```diff
@@ -82,7 +82,7 @@ public CriteriaBuilderImpl(Metamodel metamodel){
      */
     @Override
     public CriteriaQuery<Object> createQuery(){
-        return new CriteriaQueryImpl(this.metamodel, ResultType.UNKNOWN, ClassConstants.OBJECT, this);
+        return new CriteriaQueryImpl<Object>(this.metamodel, ResultType.UNKNOWN, ClassConstants.OBJECT, this);
```
lukasj, line 85: would it make sense to turn the return type to `<T> CriteriaQuery<T>` from `CriteriaQuery<Object>` when you're here?

dazey3: @lukasj 
1) CriteriaBuilderImpl isn't a generic class, so I don't know that I see the value in such a change. True, we have `createQuery` that returns <Object> and `createTupleQuery` that returns <Tuple>. That could possibly be reduced down to just one method with generics, except for #2
2) `CriteriaBuilderImpl.createQuery()` overrides `jakarta.persistence.criteria.CriteriaBuilder.createQuery()`:
```java
public interface CriteriaBuilder {

    CriteriaQuery<Object> createQuery();

    CriteriaQuery<Tuple> createTupleQuery();
```
We couldn't make such a change without first changing the spec

dazey3: @lukasj Did I understand your point correctly? Are there any other review comments?

lukasj: it's not about CriteriaBuilderImpl being a generic class, it's about the ability to rely on/use type inference in "user" code. Since there is already similar method with the same erasure, the change does not make sense

lukasj (approved):
LGTM

PR #1463, opened 2022-03-25
[2.7] JAXB tests fix - Java memory heap increase

rfelcman:
Tests fix to avoid "Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded".
This is solution based on resources (Java Heap) increase.


Tomas-Kraus (approved):
LGTM
Let's see if this helps.

Issue #858, opened 2020-07-07
EclipseLink wrong query for left join generated


RuiBaiXYZ:
I had two tables with no relationship defined in entity. Under logic requirement, has to query some results from table _RcmSpFunctionalFailure_ with left join to table  _RcmSpComponent_.

Here is the join details.

```
RcmSpFunctionalFailure rsff
left join RcmSpComponent rsc
on rsc.id.tenantId = rsff.id.tenantId
and rsc.id.companyId = rsff.id.companyId
and rsc.id.projectId = rsff.id.projectId
and rsc.id.msiId = rsff.id.msiId
and rsc.id.inWorkNumber = rsff.id.inWorkNumber
and rsc.id.issueNumber = rsff.id.issueNumber
and rsc.deleteFlag = rsff.deleteFlag
and rsc.componentId = rsff.failureCauseComponentId
```
EclipseLink translated SQL is
`SELECT t0.tenant_id, t0.company_id, t0.project_id, t0.msi_id, t0.issue_number, t0.in_work_number, t0.function_failure_instance_id, t0.parent_function_failure_instance_id, t0.function_failure_cd, t0.function_failure_type_code, t0.function_failure_description, t0.failure_cause_prefix, t0.failure_cause_component_id, t0.failure_cause_postfix, t1.manual_added_flag, t1.manual_added_component_name, t0.msi_serial_number FROM proam.rcm_sp_functional_failure t0 LEFT OUTER JOIN proam.rcm_sp_component t1 ON ((((((((t1.tenant_id = t0.tenant_id) AND (t1.company_id = t0.company_id)) AND (t1.project_id = t0.project_id)) AND (t1.msi_id = t0.msi_id)) AND (t1.in_work_number = t0.in_work_number)) AND (t1.issue_number = t0.issue_number)) AND (t1.delete_flag = t0.delete_flag)) AND `
[...]

PR #1465, opened 2022-04-01
Left Join Fetch on Embeddable ElementColection fails & NamedQuery with QueryHints.REFRESH fails to execute 

jGauravGupta:
**Fixes issue 1:** 
JPA Left Join Fetch on Embeddable ElementColection fails
https://github.com/eclipse-ee4j/eclipselink/issues/858

**Fixes issue 2:**
When defining a NamedQuery that has EclipseLink hints, the query fails with an exception

````
Exception [EclipseLink-7081] (Eclipse Persistence Services - 2.7.6.payara-p1): org.eclipse.persistence.exceptions.ValidationException
Exception Description: The aggregate object [class be.rubus.payara.support.ticket1455.model.Tag] cannot be directly registered in the UnitOfWork.  It must be associated with the source (owner) object.
	at org.eclipse.persistence.exceptions.ValidationException.cannotRegisterAggregateObjectInUnitOfWork(ValidationException.java:622)
````
The offending NamedQuery, org.eclipse.persistence.config.QueryHints
````
@NamedQuery(
        name = "getAllBlogPosts_fail",
        query = "SELECT i FROM BlogPost i",
        hints = {
                @QueryHint(
                        name = QueryHints.REFRESH,
                        value = HintValues.TRUE
                ),
                @QueryHint(
                        name = QueryHints.LEFT_FETCH,
                        value = "i.tags"
                )
        }
)
````

Sample test app:
[fish-406.zip](https://github.com/eclipse-ee4j/eclipselink/files/8395492/fish-406.zip)

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java
@@ -3021,7 +3021,7 @@ public class UnitOfWorkImpl extends AbstractSession implements org.eclipse.persi
             return null;
         }
         if (descriptor.isDescriptorTypeAggregate()) {
-            throw ValidationException.cannotRegisterAggregateObjectInUnitOfWork(object.getClass());
+           return null;
         }
         Object registeredObject = checkIfAlreadyRegistered(object, descriptor);
         if (registeredObject == null) {
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metamodel/ManagedTypeImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metamodel/ManagedTypeImpl.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2011, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020 Payara Services Ltd.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -87,6 +88,7 @@ import org.eclipse.persistence.internal.queries.ContainerPolicy;
 import org.eclipse.persistence.internal.security.PrivilegedAccessHelper;
 import org.eclipse.persistence.internal.security.PrivilegedGetDeclaredField;
 import org.eclipse.persistence.internal.security.PrivilegedGetDeclaredMethod;
+import org.eclipse.persistence.internal.sessions.AbstractSession;
 import org.eclipse.persistence.logging.AbstractSessionLog;
 import org.eclipse.persistence.logging.SessionLog;
 import org.eclipse.persistence.mappings.CollectionMapping;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by lukasj — CHANGES_REQUESTED** [review]
Needs a test, reliance on NPE being thrown is not an acceptable solution.

File: jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/metamodel/ManagedTypeImpl.java [hunk 1]
```diff
@@ -1329,6 +1331,18 @@ protected void initialize() { // Future: Check all is*Policy() calls
             this.members.put(mapping.getAttributeName(), member);
         }
     }
+    
+    public void preinitaliseMappings(AbstractSession session) {
+        for (DatabaseMapping mapping : getDescriptor().getMappings()) {
+            try {
+                mapping.preInitialize(session);
+            } catch (NullPointerException npe) {
+                // A NPE gets thrown if the expected method is not present for the mapping
```
t-beckmann, line 1340: Should be fixed at the origin of failure.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java
```diff
@@ -3021,7 +3021,7 @@ public Object internalRegisterObject(Object object, ClassDescriptor descriptor,
             return null;
         }
         if (descriptor.isDescriptorTypeAggregate()) {
-            throw ValidationException.cannotRegisterAggregateObjectInUnitOfWork(object.getClass());
+           return null;
```
t-beckmann, line 3024: Can you elaborate why this is fixing the issue? Technically I see you are stopping the exception from being thrown, of course. Is this a try-and-error fix, or some deeper understanding? To me it seems the method should not have been called in the first place for those aggregates.

robertpatrick:
@jGauravGupta @arjantijms I reproduced the failures on current master, and your complete patch fixes them in the local tests. I’d still like to see the fix narrowed before merging.

| ---- | ---- | ---- | 
| LEFT JOIN FETCH, then edit an embedded value and flush | Uninitialized-mapping error | Correct update |
| Named query with REFRESH and LEFT_FETCH | Aggregate-registration exception | Correct results |
| Refresh externally changed owner/tags, then modify the collection and flush | Fails | Fresh values returned; subsequent writes correct |
| Ordinary reads/writes and empty-collection controls | Correct | Unchanged |

The complete patch passed all 12 query/lifecycle scenarios on H2 and Derby, including an additional woven-model run on H2. I found no introduced regression in those cases.

The underlying problem appears to be joined embeddables entering the entity-registration path. The patch leaves that registration in place, suppresses its exception, and reinitializes mappings globally to avoid the later flush failure.  A diagnostic change confined to joined-embeddable construction also fixed the original read/refresh/write cases without either workaround. That prototype isn’t production-ready, but it gives us a concrete direction.

What we need:

| Action | Reason |
| ---- | ---- |
| Fix joined-embeddable construction/refresh at its source | Keep values owned by their entity; preserve registration checks and avoid global mapping reinitialization with blanket NPE suppression |
| Add regressions based on the supplied application | Include SharedCacheMode.NONE, named hints, refreshed values, and read/edit/flush checks—not just successful query execution |
| Rebase and obtain passing existing CI | Currently conflicting, 651 commits behind, with no current-head checks/status results reported |

These are worthwhile fixes. I would support a focused revision with those tests and clean CI.

arjantijms:
@jGauravGupta hi Guarav, any chance to look at this one again?

PR #1466, opened 2022-04-01
Generate correct Informix SQL when parsing Concat Operator

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>

EclipseLink generates SQL, that can't be parsed by the database. The problem is located in the handling of the Concat operator (JPQL and the criteria api are both affected).

Informix expectes the concat operation to be: `<op1> || <op2>`

Eclipselink generates: `<op1> + <op2>`

This is rejected by the DB server. The concat operator can be overridden for the db platform and if done so, the correct SQL is generated.
The test program is attached to demonstrate this issue. The program (de.persona.dev.blaesing.eclipselink.jpa.generation.test.Test) generates tables for a Demo Entity and then queries the concatenation of lastname and firstname.

Lines 17-18 instantiate the entitymanager 
Lines 20-21 demonstrate a fix (if commented out, the error exists, with this fix the query works) 
Lines 23-31 insert demo data 
Lines 35-43 reproduce the problem with the criteria api 
Lines 46-49 reproduce the problem with JPQL

The current result with line 20-21 commented throws an exception:

`Caused by: java.sql.SQLException: A character to numeric conversion process failed`

[test-app.zip](https://github.com/eclipse-ee4j/eclipselink/files/8395591/test-app.zip)

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/InformixPlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/InformixPlatform.java
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2015 IBM Corporation and/or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


lukasj (changes requested):
ECA check failed

lukasj (approved):
LGTM

PR #1467, opened 2022-04-01
Do not log a stack trace when creating a sequence if it does not already exist

jGauravGupta:
Signed-off-by: Jonathan Coustick <jonathan.coustick@payara.fish>


lukasj (changes requested):
[SchemaManager](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/tools/schemaframework/SchemaManager.java#L412) says stacktraces are not to be logged which is in contrast with the subject of this PR. If the subject is correct, then the change does not look right as it bypasses possible problem in the flag of suppression of the stacktrace

lukasj:
superseded by #1666

Issue #1480, opened 2022-04-12
Bug 579409: Add support for Statement.getGeneratedKeys for Identity


dazey3:
Linking to [BZ #579409](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579409)

When persisting entity values, that use IDENTITY generation, EclipseLink may obtain an incorrect IDENTITY on SQLServer. If the SQLServer database has added TRIGGERs that generate IDENTITY values before EclipseLink is able to query for the correct value, the wrong value is obtained.

When persisting an entity, that uses IDENTITY generation, EclipseLink performs an INSERT followed by a separate query to obtain the ID value:

```
    [EL Fine]: sql: Connection(1940526571)--Thread(Thread[main,5,main])--INSERT INTO COFFEE(NAME, PRICE) VALUES (?, ?)
	bind => [Capa, 4.2]
...
    [EL Finest]: query: Thread(Thread[main,5,main])--Execute query ValueReadQuery(name="SEQ_GEN_IDENTITY" sql="SELECT @@IDENTITY")
    [EL Fine]: sql: 2022-03-24 11:30:15.807--ClientSession(1647497427)--Connection(1940526571)--Thread(Thread[main,5,main])--SELECT @@IDENTITY

```
Notice that in this trace from SQLServer, EclipseLink executes `SELECT @@IDENTITY` to obtain the ID value. The problem is that this value returned from `SELECT @@IDENTITY` may not be correct. If another process generates an IDENTITY value BEFORE EclipseLink can execute this separate query, then the value returned will not be the one matching the actual entity inserted into the table.

------------

Table:
```sql
    CREATE TABLE COFFEE (ID NUMERIC(19) IDENTITY NOT NULL, NAME VARCHAR(255) NULL, PRICE FLOAT(32) NULL, PRIMARY KEY (ID))
```

For instance, consider the following example. If the SQLServer database contains the following additional, non-JPA table that also uses IDENTITY generation:
```sql
    CREATE TABLE COFFEE_AUDIT (AUDIT_ID NUMERIC(19) IDENTITY NOT NULL, COFFEE_ID NUMERIC(19) NOT NULL, NAME VARCHAR(255) NULL, PRICE FLOAT(32) NULL, PRIMARY KEY (AUDIT_ID))
```

and a TRIGGER to insert into that table:
```sql
[...]

Issue #1393, opened 2022-01-13
IDENTITY in H2 2.0.202 and later is removed in some cases; H2Platform may need to be changed


ljnelson:
I just ran across this and didn't want it to get lost.  I'm not sure what the proper solution is.

H2 [changed](https://github.com/h2database/h2database/pull/3067) certain details about `IDENTITY` in their grammar.

A quick search shows me it is still used here:

https://github.com/eclipse-ee4j/eclipselink/blob/3e2c7e55986794d23c83c2b0cb9bbf9103af04f7/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/H2Platform.java#L143-L155

Maybe that is OK, and maybe it is not but I suspect there will be problems here.

katzyn:
`printFieldIdentityClause()` is implemented incorrectly, it should return `" GENERATED BY DEFAULT AS IDENTITY"` for all versions of H2 unconditionally. This change is trivial.

Situation with `buildSelectQueryForIdentity()` is complicated. `IDENTITY()` is obliviously a wrong function. It doesn't actually return a generated identity value from the last `INSERT` operation executed by application in this session. If table has a trigger and that trigger performs, for example, an another `INSERT` operation in some different table with identity column, a value of that identity column will be returned instead.

For H2 1.x `SCOPE_IDENTITY()` can be used, it doesn't return unexpected values from triggers.

H2 2.x doesn't have these functions. There are only two ways to get the inserted values:
1. `Statement.getGeneratedKeys()` (collection of keys needs to be enabled first by providing names or ordinal number of columns or simply with `Statement.RETURN_GENERATED_KEYS`).
2. Standard data change delta tables (`SELECT ID FROM FINAL TABLE (INSERT INTO TEST(A, B) VALUES (?, ?))`).

Hibernate ORM uses the first method with all drivers that support that. From my point of view EclipseLink should do the same instead of tricks with various vendor-specific functions.

We added these weird legacy functions to `LEGACY` compatibility mode in H2 2.1.210, but this mode should be considered only as a temporary workaround. In this version both `IDENTITY()` and `SCOPE_IDENTITY()` work like `SCOPE_IDENTITY()` in H2 1.x.

MelleD:
@katzyn is there something similar like in SQL SERVER `SELECT @@IDENTITY` for H2 2.1.210?

dazey3:
[...]

PR #1481, opened 2022-04-12
[master] Bug 579409: Add support for Statement.getGeneratedKeys for Identity Generation

dazey3:
#1480


dazey3:
For `master`, I decided to enable `supportsReturnGeneratedKeys` for MySQL, Derby, DB2, and SQLServer platforms as those were the only ones I tested. Other databases may support `Statement.getGeneratedKeys()`, but someone will need to add that support and the appropriate testing as well.

For service branches (`2.6_WAS`, `2.7`, & `3.0`), I elected to continue having `supportsReturnGeneratedKeys` disabled for all platforms. Support can be enabled by users by setting a persistence property:
```xml
    <property name="eclipselink.target-database-properties" value="supportsReturnGeneratedKeys=true" />
```
This will preserve existing behavior as not all databases that support it have issues with the current `DatasourcePlatform.buildSelectQueryForIdentity()` strategy. For instance, DB2 uses `SELECT IDENTITY_VAL_LOCAL() FROM SYSIBM.SYSDUMMY1` to obtain IDENTITY values. This documented strategy does not return IDENTITY values generated in other levels (like TRIGGERS or stored procedures). Enabling `supportsReturnGeneratedKeys` for DB2 should only be a performance benefit as 2 queries may not be necessary

dazey3:
I would appreciate a code review or even just feed back on this PR. This may be a fix for #1393

dazey3:
One alternative implementation I looked into was passing the ResultSet to the AbstractSession, or possibly even the DatabaseAccessor, so that it would be available both during INSERT (`DatabaseAccessor.basicExecuteCall()`) and later when getting the Sequence value (`ObjectBuilder.assignSequenceNumber()`).
The benefit of this strategy was that I didn't need to change any of the API calls that currently expect Integer result types! However, I just didn't feel like the Sessions or Accessors should be used like that; the Call seemed more appropriate as the Call is a disposable clone and Calls themselves, I believe, are more like "execution receptacles" where storing results made more sense. Also, the current CURSOR implementation does something very similar, just with queries that actually return results.

One issue I faced was that INSERT queries are viewed as `DatasourceCall.NO_RETURN` and a result is not expected. The CURSOR implementation actually has it's own `org.eclipse.persistence.internal.databaseaccess.DatasourceCall.returnType`! I thought about implementing a new `returnType` as well... but it felt like it changed INSERTs much more than just changing some (Integer) explicit casts and internal API methods.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
```diff
@@ -215,7 +219,14 @@ public boolean isJPQLCallQueryMechanism() {
      * This should be overridden by subclasses.
      * @return the row count.
      */
-    public abstract Integer executeNoSelect() throws DatabaseException;
+    public abstract Object executeNoSelect() throws DatabaseException;
```
lukasj, line 222: On a first look, this is sort of the change (Int -> Object) I don't like. I'm not saying it's wrong but isn't there some other alternative like ie introducing new method returning the right type instead of generic Object? From the diff, it looks like the return value will always be either Integer or DatabaseCall and having Object there may lead to headaches during maintenance..

lukasj:
I'm sure you are aware of https://wiki.eclipse.org/EclipseLink/Development/Architecture/EclipseLink

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
dazey3: Sure, I could just create new methods, copy everything over, and even deprecate the old ones. However, I figured this was safe because it was internal API (also, this is HEAD where backwards compatibility isn't necessarily too important).

I went with the implementation that required the least amount of code duplication. It isn't that the overall execution strategy of INSERT queries is changing, just that it isn't necessarily ALWAYS returning an Integer up the call stack. Creating new API methods to call would be cleaner, but determining when to make those calls over the current defaults is trickier.

If it's a hard pass, I can take a look at creating new methods and what that would entail.

dazey3:
> I'm sure you are aware of https://wiki.eclipse.org/EclipseLink/Development/Architecture/EclipseLink

Sure, I'm just not sure what there is to get out of it. The JavaDoc on most classes is fairly good at describing the class function and how to use it. This seems more... historical and high level.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
lukasj: I'm not sure yet, I'm still checking the codebase if there is something solving similar problem like handling multiple resultsets through one call and then somehow merging results already - what caught my attention is the comment about `DatasourceCall.NO_RETURN` - looking at [prepareInsertObject](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatasourceCallQueryMechanism.java#L612)  it seems to me, that at the point the query is being prepared, it may be already known that identity is used, so why not start by properly configuring the call itself?

lukasj: [#350487](https://bugs.eclipse.org/bugs/show_bug.cgi?id=350487) added handling of multiple resultsets through one call, so that might be good source

dazey3: > looking at [prepareInsertObject](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatasourceCallQueryMechanism.java#L612) it seems to me, that at the point the query is being prepared, it may be already known that identity is used

The order of operations is roughly this:
1) `NativeSequence` is created. It determines that IDENTITY is being used and that the Platform class supports GeneratedKeys. So, `NativeSequence.shouldUseGeneratedKeysIfPlatformSupports` is set `true`
2) InsertObjectQuery is created. It checks the descriptors sequence (possibly being the NativeSequence from step 1) and sets `DatabaseQuery.shouldReturnGeneratedKeys` during prepare.
3) A SQLCall is created for the InsertObjectQuery. That SQLCall is prepared, which copies the `DatabaseQuery.shouldReturnGeneratedKeys` to `DatabaseCall.shouldReturnGeneratedKeys`
4) The InsertObjectQuery is executed..
5) The SQLCall is cloned and executed.
6) The SQLCall clone creates a Statement and executes it.
7) The GeneratedKeys ResultSet is stored in the cloned SQLCall, which is returned up to `DatasourceCallQueryMechanism.insertObject()`
8) The SQLCall is passed down into `DatabaseQueryMechanism.updateObjectAndRowWithSequenceNumber()` where it reads the key from the ResultSet as the ID value
9) The ID value is passed to `ObjectBuilder.assignSequenceNumber()`, where the Sequence is not used, but rather the value passed in is

> at the point the query is being prepared, it may be already known that identity is used

Right, so during query prepare, we can know that. We could change the `org.eclipse.persistence.internal.databaseaccess.DatasourceCall.returnType` from `DatasourceCall.NO_RETURN` to a new type. I tried implementing that, but I didn't see much of a benefit and it may potentially involve more code changes. One issue is that at a couple places in the code, it appears to be assumed that `call.isNothingReturned()` == INSERT. That logic would need to be addressed.

I'm not saying this is the wrong fix or a bad idea, but I opted for the implementation that tried to reuse the INSERT execution code path, rather than creating a new parallel path.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
```diff
@@ -215,7 +219,14 @@ public boolean isJPQLCallQueryMechanism() {
      * This should be overridden by subclasses.
      * @return the row count.
```
rfelcman, line 220: I'd like extend comment there with additional possible return type.

rfelcman (approved):
LGTM
With following note.
I tried to enable this in Oracle platform but at JDBC level I found following issue:

When I used in my JDBC testcase
`pStmt = conn.prepareStatement("INSERT INTO TEST_IDENTITY_TAB (NAME) VALUES ( ? )", Statement.RETURN_GENERATED_KEYS);`
line line in
`org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor`
`statement = nativeConnection.prepareStatement(call.getSQLString(), Statement.RETURN_GENERATED_KEYS);`
Oracle JDBC driver always return ROWID (internal DB row id) instead of PRIMARY KEY column.
It's a question if this is correct JDBC driver behavior or not, but maybe we should modify in some next change into something like this to narrow JDBC driver to the "right" DB columns
`pStmt = conn.prepareStatement("INSERT INTO TEST_IDENTITY_TAB (NAME) VALUES ( ? )", new String[] {"ID"});`
It leads me into idea to extend scope of `@GeneratedValue` to other entity fields, not only `@Id`s. Future versions of JPA spec? It should little bit decrease communication with database.
I know, that we have `@ReturnInsert` in the EclipseLink, but this is our additional property.

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/sequence/TestIdentityGeneration.java
```diff
@@ -0,0 +1,215 @@
...
+import org.eclipse.persistence.jpa.test.sequence.model.Tea;
+import org.eclipse.persistence.jpa.test.sequence.model.TeaShop;
+import org.eclipse.persistence.platform.database.DatabasePlatform;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(EmfRunner.class)
+public class TestIdentityGeneration {
```
rfelcman, line 39: Maybe tests there should be limited to the platforms with `this.supportsReturnGeneratedKeys == true` .

dazey3: I have `org.eclipse.persistence.jpa.test.sequence.TestIdentityGeneration.testRefreshWithTriggers()` gated at the beginning with 
```java
    Assume.assumeTrue(platform.supportsReturnGeneratedKeys());
```

I don't have `org.eclipse.persistence.jpa.test.sequence.TestIdentityGeneration.testPersistWithSecondaryTables()` gated with any check because there isn't anything inherently platform specific for that test. It should run successfully against any database, regardless of their support for "RETURN_GENERATED_KEYS". The test was included to exercise the code paths I changed and test for failures.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
dazey3: Sure. I'll add a comment to `org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeNoSelect()`, rather than the abstract class `DatabaseQueryMechanism`. Does that work?

dazey3:
@rfelcman 
Thanks so much for taking a look and providing feedback

> Oracle JDBC driver always return ROWID (internal DB row id) instead of PRIMARY KEY column.
It's a question if this is correct JDBC driver behavior or not

So, I didn't end up testing this against Oracle for 2 reasons:
1) The existing `org.eclipse.persistence.internal.databaseaccess.DatasourcePlatform.buildSelectQueryForIdentity()` seems to already work fine on Oracle (or rather, I didn't see any issues opened against Oracle that described the same issue there). If we don't set `DatasourcePlatform.supportsReturnGeneratedKeys = true` for OraclePlatform, then Oracle should just keep using the current method of executing a second query to obtain IDENTITY values.
2) TBH, I was being a bit lazy. I didn't have time to lookup how to create TRIGGERs on Oracle. I tested against SQLServer, MySQL, Derby, and DB2. SQLServer was the database that my customer reported this issue against and I have seen Microsoft pushing for this support as well. MySQL is a common platform we run our testing against for EclipseLink. Derby is a common platform we at IBM run testing against and DB2 is very similar to Derby so I threw it in. 
Only SQLServer seems to NEED this fix, as the other platforms appear to work as is but this support may be a slight performance improvement (being 1 INSERT vs 1 INSERT + 1 SELECT). H2 DB may also benefit from this, but I was unable to get a H2 DB setup for testing. Also, H2 users can set the persistence property themselves and assist with testing. Someone else could also submit a PR to add it in for H2Platform if that testing works out well.

> maybe we should modify in some next change into something like this to narrow JDBC driver to the "right" DB columns
pStmt = conn.prepareStatement("INSERT INTO TEST_IDENTITY_TAB (NAME) VALUES ( ? )", new String[] {"ID"});

Sure! I was aware that that exists, but I was unsure how much complexity that would add. I didn't do a deep investigation on how to obtain the PK fields in order to implement that, so that may be a "better" implementation if that information can cleanly be passed around.
If the JPA Spec adds more support in the future for this, I guess this whole feature could be scraped for a more embedded implementation with further capabilities. Let me know if you plan on working on that and I'll take a look!

rfelcman:
> Platform to process the ResultSet. It seems like Oracle needs to use the Integer value returned + the generatedKeys ResultSet to calculate the IDENTITY value... I'm not sure though where I could add that step in. Suggestions?

I think, that at this moment we will keep Oracle platform untouched. It should be solved in next change to obtain PK columns. I think, that should leads into some more bigger changes to promote information about `@Id` fields/PK columns to `foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabaseAccessor.java`.

dazey3:
@lukasj Did you have any suggested changes before I merge this and create the backports?

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/DatabaseQueryMechanism.java
```diff
@@ -873,6 +884,64 @@ protected void updateObjectAndRowWithSequenceNumber() throws DatabaseException {
         writeQuery.getDescriptor().getObjectBuilder().assignSequenceNumber(writeQuery);
     }
 
+    /**
+     * Update the object's primary key by getting the generated keys from the call
+     * If there are no generated keys or the value is NULL, then default back to the {@link #updateObjectAndRowWithSequenceNumber()}
+     */
+    protected void updateObjectAndRowWithSequenceNumber(DatabaseCall call) throws DatabaseException {
```
lukasj, line 891: Why do you need to have an instance of the DatabaseCall and cannot use the one obtained via `writeQuery.getCall()`? Would that allow one to get rid of this new method and the change in the return type?

dazey3: @lukasj Sorry, I have been a bit busy and didn't get to this

> Why it is required to have an instance of the DatabaseCall here instead of using writeQuery.getCall()?

The issue is that `writeQuery.getCall()` returns the QueryMechanism's Call object, which is the original and not the clone. Query's don't actually store Call objects.

It's all about execution Calls vs querymechanism Calls. The former is just a clone of the latter. You can see where this occurs in the stack here: `org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCall databaseCall)`

During `org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.basicExecuteCall(...)`, all I have available is a Call, no Query. But it's here, at this depth in the call stack, that we have the `java.sql.Statement` so we can get the GeneratedKeys ResultSet. This Call object is the execution clone. So, I am placing the GeneratedKeys ResultSet on the cloned Call and returning it up the stack. I think this makes sense functionally as Call's are basically just Action objects and the clone is disposable once the action is completed. 

The cloned Call then returns back up the stack, eventually arriving to `org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.insertObject()` and passing it to `org.eclipse.persistence.internal.queries.DatabaseQueryMechanism.updateObjectAndRowWithSequenceNumber(DatabaseCall call). In order to use `writeQuery.getCall()`, I would need to set the ResultSet on the original Call object. Which I think is the wrong place to do that.

lukasj (approved):
LGTM; @rfelcman will file followup issue

dazey3:
@lukasj Sorry to make this an even longer running review! But it just came to my attention that setting the persistence property on 3.0 and 2.7 was not working as expected:
```xml
 <property name="eclipselink.target-database-properties" value="supportsReturnGeneratedKeys=true" />
```
The reason for that is due to WHERE `eclipselink.target-database-properties` are being applied to the Platform class. Currently, it is being done after `EntityManagerFactoryProvider.login()` is called. This is the correct order to do this in since the Platform class isn't set until `DatabaseSessionImpl.connect()` is called.

However, the NativeSequence is being initialized when the Descriptors are initialized, as part of `DatabaseSessionImpl.postConnectDatasource()`, and `DatasourcePlatform.supportsReturnGeneratedKeys` needs to be set before that happens (`NativeSequence.shouldUseGeneratedKeysIfPlatformSupports`)

For that reason, I just added a new commit that moves the initialization of `eclipselink.target-database-properties` from `EntityManagerSetupImpl.deploy()` to `DatabaseSessionImpl.postConnectDatasource()`. I left it as a separate commit for ease of review.

dazey3:
> LGTM; @rfelcman will file followup issue

Yes, I see issue 1526 was opened to address adding support for generated keys on Oracle, since usage of `Statement.RETURN_GENERATED_KEYS` isn't allowed.

lukasj:
feel free to merge this to master (and other branches)

PR #1495, opened 2022-04-21
[3.0] Eclipselink.jar bundle - version.properties fix

rfelcman:
This is fix for `version.properties` file content included in `eclipselink.jar` bundle.
Before this fix version information related properties like `${release.version}, ${build.*}` wasn't replaced by correct values.
`version.properties` file content in `org.eclipse.persistence.core` is replaced correctly.
It happens in 3.0 branch only.

### DIFF
--- a/bundles/eclipselink/pom.xml
+++ b/bundles/eclipselink/pom.xml
@@ -437,8 +437,15 @@
                                     <directory>${gen.src.dir}</directory>
                                     <excludes>
                                         <exclude>**/*.java</exclude>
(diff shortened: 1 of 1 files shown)


bundles/eclipselink/pom.xml
```diff
@@ -438,6 +438,7 @@
                                     <excludes>
                                         <exclude>**/*.java</exclude>
                                     </excludes>
+                                    <filtering>true</filtering>
```
lukasj, line 441: should rather filter one particular resource to avoid surprises

lukasj (approved):
LGTM

PR #1521, opened 2022-05-27
Fixed SQL generation of JPQL IS EMPTY with printInnerJoinInWhereClause=false

Tomas-Kraus:


lukasj (commented):
control behaviour via property

Tomas-Kraus:
@lukasj I'd prefer to not merge this now. It's very specific use-case and solution is not a real fix, but just some kind of workaround.

Tomas-Kraus:
Closing. This won't be merged.

Issue #1504, opened 2022-05-04
Improve support for query parameter binding 


dazey3:
Currently, EclipseLink disables parameter binding for queries that contain certain functions (ABS, SQRT, MOD, ect ), when running against DB2. However, usage of un-typed parameter markers within many of these functions is actually permitted on DB2. The first part of this issue adds support to EclipseLink so that SQL functions that allow un-typed parameters on DB2 will use parameter markers; if the JPQL and/or Criteria API use parameters.

JPQL Example:
```java
 Query query = em.createQuery("SELECT s.id FROM SimpleEntity s WHERE s.intVal1 = ABS(?1) AND s.intVal2 = ?2");
 query.setParameter(1, -36);
 query.setParameter(2, 42);
 query.getResultList();
```

Current SQL syntax for DB2:
```sql
SELECT ID FROM SIMPLEENTITY WHERE ((INTVAL1 = ABS(-36)) AND (INTVAL2 = 42))
```

EclipseLink parses the "ABS" function and disables parameter binding for DB2. However, it is legal for an untyped parameter marker to be passed to the "ABS" function on DB2.

NEW SQL syntax for DB2:
```sql
SELECT ID, INTVAL1, INTVAL2, STRVAL1, STRVAL2 FROM SIMPLEENTITY WHERE ((INTVAL1 = SQRT(ABS(?))) AND (INTVAL2 = ?))
```

-----------

Additionally, EclipseLink currently only supports disabling or allowing parameter binding for the entire query. This means that if a single, untyped parameter marker is not allowed within a query, then EclipseLink can only disable binding for the entire query. The second part of this issue adds support for EclipseLink to allow partial binding of parameters; making the determination on when to bind/not to bind on an individual level and not a query level.

JPQL Example:
```java
 Query query = em.createQuery("SELECT ?1 FROM SimpleEntity s WHERE s.intVal1 = ?2");
 query.setParameter(1, 1);
 query.setParameter(2, 3);
 query.getResultList();
```

Current SQL syntax for DB2:
```sql
SELECT 1 FROM SIMPLEENTITY WHERE (INTVAL1 = 3)
```

EclipseLink parses the SELECT clause and since there is a ParameterExpression as a SELECT argument, all parameter binding is disabled for the entire query. While it is correct that on DB2 it is illegal to pass an untyped parameter marker as a SELECT argument, that is not true for the "=" usage in the WHERE clause.

NEW SQL syntax for DB2:
```sql
[...]

PR #1534, opened 2022-06-09
[master] Fix JSE tests failing on derby

dazey3:
for #1443

There were several test failures in the `TestMathFunctions` tests on Derby. These failures were because using untyped parameters with the new "ROUND" operator syntax is invalid

```sql
SELECT FLOOR((DOUBLEVALUE)*1e?+0.5)/1e? FROM NUMBERENTITY WHERE (ID = ?)
```
I updated the implementation in DerbyPlatform to that issue by not allowing parameter binding. Also, I cleaned up the implementation.

for #1504

The tests delivered by for this issue didn't respect that `DB2Platform.shouldBindLiterals` was changing the default from "false" -> "true". As a result, almost all the `TestQuerySyntax*Tests` failed as the expected default query syntax was not correct for DB2 & Derby

Also, there were failures with the "IN" operator for Derby and DB2/zOS. Those failures were because it is invalid on those platforms for ALL arguments to be untyped parameters (ie "... ? IN (?, ?, ?)". So, I updated DerbyPlatform and DB2ZPlatform to not allow that syntax.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/SQLSelectStatement.java
@@ -776,6 +776,11 @@ public class SQLSelectStatement extends SQLStatement {
                     } else if(expression.isConstantExpression() && printer.getPlatform().shouldBindLiterals()) {
                         ((ConstantExpression) expression).setCanBind(false);
                     }
+                } else if (printer.getPlatform().isDynamicSQLRequiredForFunctions()) {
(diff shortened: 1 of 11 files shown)


Tomas-Kraus (approved):
Looks good.

lukasj:
Down to 1 failure: https://ci.eclipse.org/eclipselink/job/eclipselink-cb-master/577/testReport/

dazey3:
@lukasj 

> Down to 1 failure: https://ci.eclipse.org/eclipselink/job/eclipselink-cb-master/577/testReport/

https://github.com/eclipse-ee4j/eclipselink/pull/1523#issuecomment-1146477503

PR #1536, opened 2022-06-10
[master] Fix issue 1504 JSE test failure on Derby pt2

dazey3:
A test fix was missed in https://github.com/eclipse-ee4j/eclipselink/pull/1534

`org.eclipse.persistence.jpa.test.property.TestParameterBinding.testIN_ForceBindJPQLParameters()` 
```
java.sql.SQLSyntaxErrorException: It is not allowed for both operands of 'IN' to be ? parameters.
Error Code: 30000
Call: SELECT 2 FROM GENERICENTITY WHERE (? IN (?, ?, ?))
 bind => [a, a, b, c]
Query: ReportQuery(referenceClass=GenericEntity sql="SELECT 2 FROM GENERICENTITY WHERE (? IN (?, ?, ?))")
```

This test was failing on Derby because the test is testing the persistence property `eclipselink.jdbc.force-bind-parameters`. With this property set, all parameters are bound as parameters, which is illegal on Derby. I recently changed `DB2Platform.shouldBindLiterals` to default to "true", making the literals in the test to now also bind as parameters. This causes a failure on Derby as it is invalid syntax. The test needs updated to respect that this test will fail on Derby

The property `eclipselink.jdbc.force-bind-parameters` should be deprecated really. The latest property `eclipselink.jdbc.allow-partial-bind-parameters` is much more comprehensive as it doesn't completely "force" parameter binding and respects platform syntax legality.

### DIFF
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/property/TestParameterBinding.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/property/TestParameterBinding.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2019, 2021 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2019 IBM Corporation. All rights reserved.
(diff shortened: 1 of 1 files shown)


rfelcman (approved):
LGTM (except copyright year).

Issue #1538, opened 2022-06-13
Type Conversion error with the CEILING/FLOOR Operations


jgrassel:
Given entity:
```
@Entity
public class QueryEntity {
    @Id
    private int id;

    private int intVal;
    private long longVal;
    private float floatVal;
    private double doubleVal;
```

The execution of an operation with the ceiling()/floor() criteria API operation:
```
            CriteriaBuilder cb = em.getCriteriaBuilder();
            CriteriaQuery cq = cb.createQuery(Float.class);
            Root<QueryEntity> root = cq.from(QueryEntity.class);
            cq.select(cb.ceiling(root.get("floatVal")));
            cq.where(cb.equal(root.get("id"), id));
            Float fResult = (Float)  em.createQuery(cq).getSingleResult();
```

Results in the following failure (using Oracle):
```
java.lang.ClassCastException: class java.math.BigDecimal cannot be cast to class java.lang.Float (java.math.BigDecimal and java.lang.Float are in module java.base of loader 'bootstrap') 
at io.openliberty.jpa.tests.jpa31.web.TestNewQueryMathFunctionsServlet.testFunction_CEILING_Criteria(TestNewQueryMathFunctionsServlet.java:136) 
```

The spec says `The ABS, CEILING, and FLOOR functions accept a numeric argument and return a number (integer, float, or double) of the same type as the argument.`, so it should be returning a Float type.

I noticed that this only affects criteria API, the JPQL version of this query works correctly.

Tomas-Kraus:
Thanks for reporting this Jody. I thought that there was good enough test coverage, but I was wrong.
Databases return this as Double and because field type set in `CriteriaBuilderImpl#ceiling(Expression<N>)` is just Number, it passes trough `DatabaseAccessor#getObjectThroughOptimizedDataConversion(...)` method without any conversion.

So we have field mappings set as:
- field.type :: Number
- fielt.typeName :: "java.lang.Float"
this requires additional handling in `DatabaseAccessor#getObjectThroughOptimizedDataConversion(...)`

PR #1542, opened 2022-06-14
Issue #1538: Type Conversion error with the CEILING/FLOOR Operations

Tomas-Kraus:
Fixes DB ResultSet conversion when returned type is common Number and specific type is stored only in `typeName`.
Fixes issue #1538


jgrassel (approved):
lgtm!  Thanks!

lukasj (approved):
LGTM

PR #1553, opened 2022-06-22
Fix creation of dynamic entities on SE 11+

lukasj:
moving jpa.beanvalidation and jpa.beanvalidation.dynamic tests from the jpa.test module to "standalone apps" with simple junit tests on SE environment revealed the problem with creating dynamic entities.

moved tests are tests for this bug therefore they are part of this fix and removed from the "monolithic" jpa.test module. Server side runs are still TBD


rfelcman (commented):
Test execution should be included into _etc/jenkins/pr_verify.groovy_ to `stage('JPA Modelgen, JPA JSE, WDF, JPARS, DBWS, DBWS Builder, Distribution') {`

rfelcman (commented):
Maybe execution should be near other JPA tests not at the end.

rfelcman (approved):
LGTM

PR #1555, opened 2022-06-27
JPA test framework updates:

lukasj:
- let all test runners implement same (remote) interface
- if server.testrunner.context is set, lookup and run all test runnners found there, if not, fallback to current behaviour
- add temporary profile to run tests on the server to testapps


rfelcman (approved):
LGTM

Issue #1554, opened 2022-06-26
JPA 3.1 CriteriaBuilder Query exception


hantsy:
My sample project used the following software.
* Windows 10 64bit
* Java 17
* Glassfish 7.0M4 with built-in DB

The complete codes to reproduce the error: https://github.com/hantsy/jakartaee10-sandbox/blob/master/jpa/src/test/java/com/example/it/JPQLCriteriaBuilderTest.java#L155

```java
@Entity
public class Person {
    @Id
    @Column(name = "id", nullable = false)
    @GeneratedValue(strategy = GenerationType.UUID)
    private UUID id;

```

Query:

```java 
var cb = em.getCriteriaBuilder();
var query = cb.createTupleQuery();
var root = query.from(Person.class);

query.multiselect(root.get("name"),
        cb.localTime(),
        cb.localDateTime(),
        cb.localDate()
);
query.where(cb.equal(root.get("id"), id));

var resultList = em.createQuery(query).getResultList();
```

The query execution caused the following exceptions.

```bash 
Local Exception Stack: 
Exception [EclipseLink-4002] (Eclipse Persistence Services - 4.0.0-M3.v202203111216): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLSyntaxErrorException: Syntax error: Encountered "cf3" at line 1, column 118.
Error Code: 20000
Call: SELECT NAME, CAST(CURRENT_TIME AS TIMESTAMP), CURRENT_TIMESTAMP, CURRENT_DATE FROM PERSON WHERE (id = d04a98a2-4408-4cf3-b232-5cfaab58afc9)
Query: TupleQuery(referenceClass=Person sql="SELECT NAME, CAST(CURRENT_TIME AS TIMESTAMP), CURRENT_TIMESTAMP, CURRENT_DATE FROM PERSON WHERE (id = d04a98a2-4408-4cf3-b232-5cfaab58afc9)")
	at org.eclipse.persistence.exceptions.DatabaseException.sqlException(DatabaseException.java:343)
	
```

The complete server log file is here.
[server.log](https://github.com/eclipse-ee4j/eclipselink/files/8986828/server.log)

lukasj:
provide full stacktrace, please

hantsy:
@lukasj Attached the complete `server.log` at the end of the original post.

Tomas-Kraus:
[...]

PR #1561, opened 2022-06-28
Issue #1554: JPA 3.1 CriteriaBuilder Query exception

Tomas-Kraus:
1. Added missing UUID value handling in `setParameterValueInDatabaseCall` and `appendParameterInternal` methods.

2. Did small refactoring in DatabasePlatform:
moved `public int appendParameterInternal(Call call, Writer writer, Object parameter)` and all related helper methods right after `setParameterValueInDatabaseCall` methods and added comments explaining close relation between them.

It was not for the first time when we forgot to implement some feature in both code paths for `eclipselink.jdbc.bind-parameters` PU property so hope that having both methods close together may avoid repeating the same mistake next time.


lukasj (approved):
LGTM

PR #1571, opened 2022-07-04
fix loading mapping files for members of the composite PU

lukasj:
found and tested via xml.extended.composite tests on WildFly


rfelcman (approved):
LGTM

PR #1576, opened 2022-07-07
avoid the need to set ejb.lookup property

lukasj:
on the server by introducing the ability to provide it directly to the test itself
via env-entry in ejb-jar.xml;
provide temporary hook to NOT prefix member PU names for extended.xml.composite members


rfelcman (approved):
LGTM

PR #1584, opened 2022-07-12
let generic runner handle PUs bound via ejb-jar.xml,

lukasj:
move TestDataSource and TestContext from core.tests to framework


rfelcman (approved):
LGTM

Issue #1506, opened 2022-05-04
BatchFetchPolicy.clone() cause memory leak and ConcurrentModifcationException


fataOT:
The latest EL update to 2.7.10 introduced a memory leak within our application.
We think the root cause for this leads back to an old bug in BatchFetchPolicy.clone() (https://bugs.eclipse.org/bugs/show_bug.cgi?id=491349)

In 2.7.10 the initialization of the "dataResults" map has been shiftet to the constructor
```
    public BatchFetchPolicy(BatchFetchType type) {
        this.type = type;
        this.dataResults = new HashMap<Object, List<AbstractRecord>>();
        this.dataResults.put(this, new ArrayList<AbstractRecord>());
    }
```

So every created BatchFetchPolicy will have a dataResults-Map with the BatchFetchPolicy as an entry.

Now let's look at the clone() method
```
    public BatchFetchPolicy clone() {
        BatchFetchPolicy clone = null;
        try {
            clone = (BatchFetchPolicy)super.clone();
        } catch (CloneNotSupportedException error) {
            throw new InternalError(error.getMessage());
        }
        if (clone.dataResults != null) {
            clone.dataResults.put(clone, clone.dataResults.get(this));
        }
        return clone;
    }
```

The clone()-method will create a BatchFetchPolicy which will use the same map reference as the original BatchFetchPolicy.
The result of one clone() call is, that the dataResults map will have 2 entries, then 3, 4, ....
As a result of this, the garbage collection is unable to remove those objects.

The ConcurrentModifcationException from the referenced bug has the same origin.
The attributeExpressions list ist not a deep copy so all cloned BatchFetchPolicies will use the same list which can cause the ConcurrentModifcationException in case of heavy usage of QueryHints.BATCH within different threads/EntityManagers

**Open questions:**
- What is the benefit for using the clone() method at all? Which data is necessary to remember?
- Is there any reason to not deep copy the data? 
- Why is the dataResults map always initialized but only used for BatchFetchType.IN (we do not use BatchFetchType.IN at all, so our workaround for the memory leak is to only initialize the dataResults for BatchFetchType.IN)

fataOT:
The memory leak has already been fixed by #1245

Issue #1245, opened 2021-08-19
Find all query with full qualified class name is recognized as ReportQuery


TJDev:
EclipseLink Version: 2.6.8.WAS

I have tested the BatchFetch feature with our current implementation and got a NullPointerException when executing a query like the following:
```sql
select e from my.package.Entity e
```

> Caused by: java.lang.NullPointerException
	at org.eclipse.persistence.queries.BatchFetchPolicy.getDataResults(BatchFetchPolicy.java:251)
	at org.eclipse.persistence.mappings.ForeignReferenceMapping.extractResultFromBatchQuery(ForeignReferenceMapping.java:562)
	at org.eclipse.persistence.internal.indirection.NoIndirectionPolicy.valueFromBatchQuery(NoIndirectionPolicy.java:297)
	at org.eclipse.persistence.mappings.ForeignReferenceMapping.batchedValueFromRow(ForeignReferenceMapping.java:275)
	at org.eclipse.persistence.mappings.ForeignReferenceMapping.valueFromRow(ForeignReferenceMapping.java:2176)
	at org.eclipse.persistence.mappings.ForeignReferenceMapping.buildCloneFromRow(ForeignReferenceMapping.java:336)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildAttributesIntoWorkingCopyClone(ObjectBuilder.java:2002)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildWorkingCopyCloneFromRow(ObjectBuilder.java:2255)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildObjectInUnitOfWork(ObjectBuilder.java:853)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildObject(ObjectBuilder.java:740)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildObject(ObjectBuilder.java:694)
	at org.eclipse.persistence.queries.ReportQueryResult.processItem(ReportQueryResult.java:251)

For my understanding the exception is thrown because the query is recognized as a ReportQuery and not a ReadAllQuery. If I delete the package definition, the query is recognized as ReadAllQuery.
```sql
select e from Entity e
```

My question is: Is this an bug of EclipseLink or should a query with package and class name be a ReportQuery?

Regards,
Thomas

dazey3:
Hello @TJDev !

[...]

PR #1585, opened 2022-07-13
Fix ConcurrentModificationException in BatchFetchPolicy

fataOT:
See #1506

The memory leak mentioned in the issue has already been solved by #1245 
This PR will fix the ConcurrentModificationException of the attributeExpression list, when running the same (batch)query with different entity managers.


foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/queries/BatchFetchPolicy.java
```diff
@@ -77,6 +77,10 @@ public BatchFetchPolicy clone() {
             dataResults.put(clone, list);
         }
         clone.setDataResults(dataResults);
+
+        if(this.attributeExpressions != null) {
+            this.attributeExpressions = new ArrayList<>(this.attributeExpressions);
```
lukasj, line 82: shouldn't it be `clone.attributeExpressions = ...` instead of `this`?

rfelcman (approved):
LGTM

PR #1600, opened 2022-07-25
buildBase64StringFromBytes more memory efficient

jbescos:
Makes this method more memory efficient getting rid of a byte[] having the base64 encoded.

Although that byte[] will be garbage collected when the method finishes, this is relevant when there are many bytes to encode and multiple tasks like this in parallel.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/XMLConversionManager.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/XMLConversionManager.java
@@ -1641,12 +1641,7 @@ public class XMLConversionManager extends ConversionManager implements org.eclip
 
     @Override
     public String buildBase64StringFromBytes(byte[] bytes) {
-        byte[] convertedBytes = Base64.base64Encode(bytes);
-        StringBuilder buffer = new StringBuilder();
-        for (int i = 0; i < convertedBytes.length; i++) {
-            buffer.append((char) convertedBytes[i]);
-        }
-        return buffer.toString();
+        return Base64.base64EncodeToString(bytes);
     }
 
     public String buildBase64StringFromObjectBytes(Byte[] bytes) {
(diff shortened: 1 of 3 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/oxm/XMLConversionManager.java
```diff
@@ -1641,12 +1641,7 @@ protected Byte[] convertSchemaBase64ToByteObjectArray(Object sourceObject) throw
 
     @Override
     public String buildBase64StringFromBytes(byte[] bytes) {
-        byte[] convertedBytes = Base64.base64Encode(bytes);
-        StringBuilder buffer = new StringBuilder();
-        for (int i = 0; i < convertedBytes.length; i++) {
-            buffer.append((char) convertedBytes[i]);
-        }
-        return buffer.toString();
+        return java.util.Base64.getEncoder().encodeToString(bytes);
```
lukasj, line 1644: I'd prefer see new method in `org.eclipse.persistence.internal.oxm.conversion.Base64` which is called from here over spreading usage of `java.util.Base64` through the codebase leading to non-explicit imports (ie when one sees Base64 in the code, can it be determined which package it comes from without looking at import statements?)

jbescos: I am just checking that the JDK implementation is also creating a new byte array anyway, so this fix will not help.

jbescos:
JDK implementation does not make it better. Closing.

jbescos:
I reopen it, because anyway current implementation relies in JDK Base64 plus the byte[].

Possibly I will fix it in the way it is processed in chunks.

jbescos:
JAXB-API DatatypeConverter#printBase64Binary is not memory efficient either. I will try other solution.

lukasj (approved):
LGTM

Issue #1540, opened 2022-06-13
EXTRACT failure with Oracle


jgrassel:
With Oracle (Derby is fine), the following code:
```
// Verify EXTRACT(YEAR) from a LocalDate field
q = em.createQuery("SELECT EXTRACT(YEAR FROM qdte.localDateData) FROM QueryDateTimeEntity qdte WHERE qdte.id = 1");
result = q.getSingleResult();
```

fails with the following Exception:

```
jakarta.persistence.PersistenceException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 4.0.0.v202206101604): org.eclipse.persistence.exceptions.DatabaseException Internal Exception: java.sql.SQLException: ORA-30076: invalid extract field for extract source Error Code: 30076 Call: SELECT EXTRACT(YEAR FROM LOCALDATEDATA) FROM QUERYDATETIMEENTITY WHERE (ID = ?) bind => [1] Query: ReportQuery(referenceClass=QueryDateTimeEntity sql="SELECT EXTRACT(YEAR FROM LOCALDATEDATA) FROM QUERYDATETIMEENTITY WHERE (ID = ?)") 
at org.eclipse.persistence.internal.jpa.QueryImpl.getDetailedException(QueryImpl.java:392) at org.eclipse.persistence.internal.jpa.QueryImpl.executeReadQuery(QueryImpl.java:265) 
at org.eclipse.persistence.internal.jpa.QueryImpl.getSingleResult(QueryImpl.java:531) at org.eclipse.persistence.internal.jpa.EJBQueryImpl.getSingleResult(EJBQueryImpl.java:400) 
at io.openliberty.jpa.tests.jpa31.web.TestNewQueryTimeFunctionsServlet.testEXTRACTFunction_OnQueryResults_JPQL(TestNewQueryTimeFunctionsServlet.java:445) 

```

Similarly, the following:
```
tq = em.createQuery("SELECT qdte FROM QueryDateTimeEntity qdte WHERE EXTRACT(YEAR FROM qdte.localDateData) = 2022", QueryDateTimeEntity.class);
tResList = tq.getResultList();
```
fails with:
```
[...]

PR #1602, opened 2022-07-26
Issue 1540: Fixed EXTRACT(WEEK FROM DATEVALUE) in Oracle platorm.

Tomas-Kraus:
This shall fix EXTRACT(WEEK FROM DATEVALUE) problem which is still there after Issue #1540 fix.


lukasj (approved):
LGTM

dazey3:
@Tomas-Kraus This should be backported to EclipseLink 3.0 and 2.7, right? The `java.time.*` support you added to OraclePlatform is applicable to JPA 2.2 and JPA 3.0, right?

PR #1615, opened 2022-08-02
#1612: DDL generation grabs column info for the wrong table

### LINKED ISSUE #1612: DDL generation grabs column info for the wrong table
Noticed this in the `DDLGenerationExtendTablesTestSuite` running sequentially  against two different schemas on mysql.

assuming there are two empty DB schemas `t1` and `t2` on mysql and the test is to end up with a table `Countries`, with columns `iso_code` and `name` where the `name` column is added by the `create-or-extend-tables` DDL option 

First run on `t1` always passes - the particular table gets created with one column, later extended with the new, `name`, column by the test and all following tests are passing.

Second run on `t2` fails with:
```
Error persisting new country with city, states and zips : jakarta.persistence.PersistenceException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 4.0.0.v202208012021): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLSyntaxErrorException: Unknown column 'name' in 'field list'
Error Code: 1054
Call: INSERT INTO Countries (iso_code, name) VALUES (?, ?)
	bind => [840, USA]
Query: InsertObjectQuery(org.eclipse.persistence.testing.models.jpa.ddlgeneration.Country@705fe9f2)
	at deployment.org.eclipse.persistence.jpa.testapps.ddlgeneration.ear//junit.framework.Assert.fail(Assert.java:57)
	at deployment.org.eclipse.persistence.jpa.testapps.ddlgeneration.ear//junit.framework.TestCase.fail(TestCase.java:223)
	at deployment.org.eclipse.persistence.jpa.testapps.ddlgeneration.ear.org.eclipse.persistence.jpa.testapps.ddlgeneration_ejb.jar//org.eclipse.persistence.testing.tests.jpa.ddlgeneration.DDLGenerationTest.testPrimaryKeyJoinColumns(DDLGenerationTest.java:276)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
...
```

the problem is, that the table in the `t2` schema is not extended with the `name` column.

The cause is that the schemaframework (the part responsible for DDL) does not take into account current schema and catalog when querying the DB for the column names of the particular table, so when it asks for columns of the `Countries` table, what gets returned is actually info about the `Countries` table from schema `t1` where all columns are already present instead of the same table from schema `t2` where the to be created `name` column is missing, so in the end, the query checking the state of the `Countries` table in schema `t2` fails.

lukasj:
fixes #1612

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/ClassDescriptor.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 1998, 2021 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 1998, 2021 IBM Corporation. All rights reserved.
+ * Copyright (c) 1998, 2022 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2022 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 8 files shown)


rfelcman (approved):
LGTM

PR #1629, opened 2022-08-03
make the jpa dependency on json optional

lukasj:
+ removed unneeded executions from testapps build
+ improved stability of tests on the server
+ cleanup in testapps


rfelcman (approved):
LGTM

PR #1631, opened 2022-08-04
FS layers squash turned off (commented out) in Docker build.

Tomas-Kraus:
Looks like this feature is broken in current maven Docker plugin.


rfelcman (approved):
LGTM

PR #1638, opened 2022-08-05
move inherited/inheritance models to testapps

lukasj:
+ move CMP3TestModel and EntityContainerTestBase to jpa.test.framework,
+ minor cleanup through testapps
+ disable failing tests


rfelcman (commented):
LGTM

Issue #1535, opened 2022-06-10
UUID Generator sometimes doesn't initialize correctly


jgrassel:
I've noticed this, hit and miss, in my own testing with Eclipselink master (4.0.x) when testing the UUID Generator feature.  Half of the time (after app start) it works , half of the time it doesn't seem to initialize properly and any attempt to use it results in a NullPointerException:

```
junit.framework.AssertionFailedError: 2022-06-03-10:24:00:974 ERROR: Caught exception attempting to call test method testBasicUUIDIdentity_UUID_Generator_JTA_XML on servlet io.openliberty.jpa.tests.jpa31.web.TestUUIDEntityIDServlet 
java.lang.RuntimeException: java.lang.NullPointerException 
at io.openliberty.jpa.tests.jpa31.web.TestUUIDEntityIDServlet.testBasicUUIDIdentity_UUID_Generator_JTA_XML(TestUUIDEntityIDServlet.java:428) 
<snip>
Caused by: java.lang.NullPointerException 
at org.eclipse.persistence.internal.sequencing.SequencingManager.getNextValue(SequencingManager.java:1142) 
at org.eclipse.persistence.internal.sequencing.ClientSessionSequencing.getNextValue(ClientSessionSequencing.java:73) 
at org.eclipse.persistence.internal.descriptors.ObjectBuilder.assignSequenceNumber(ObjectBuilder.java:435) 
at org.eclipse.persistence.internal.descriptors.ObjectBuilder.assignSequenceNumber(ObjectBuilder.java:394) 
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.assignSequenceNumber(UnitOfWorkImpl.java:546) 
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.registerNotRegisteredNewObjectForPersist(UnitOfWorkImpl.java:4446) 
at org.eclipse.persistence.internal.sessions.RepeatableWriteUnitOfWork.registerNotRegisteredNewObjectForPersist(RepeatableWriteUnitOfWork.java:540) 
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.registerNewObjectForPersist(UnitOfWorkImpl.java:4391) 
at org.eclipse.persistence.internal.jpa.EntityManagerImpl.persist(EntityManagerImpl.java:599) 
[...]

PR #1644, opened 2022-08-10
[Master] ECL1535: UUIDGenerator intermittently fails to initialize

jgrassel:
Fix for #1535


dazey3 (approved):
It's unfortunate that the `Sequence.equals()` method wasn't better implemented to do
```
if (this.getClass() != obj.getClass())
    return false;
```
but this does follow the same pattern as the other Sequence classes. LGTM

PR #1654, opened 2022-08-17
Update to ojdbc11

lukasj:
fix oracle db driver configuration
moved oracle db specific test framework parts from oracle tests to the core test framework
cleanup


rfelcman (commented):
LGTM

Issue #1625, opened 2022-08-03
AUTO for UUID generation does not work


jgrassel:
The following test case:

``` java
@Entity
@Table(name = "UUIDAutoGenEntity")
public class UUIDAutoGenEntity {

    @Id
    @GeneratedValue(strategy = GenerationType.AUTO)
    private UUID id;
}

@Test
public void testBasicUUIDIdentity_AUTO_Generator() {
    EntityManager em = uuidEmf.createEntityManager();
		
    try {
        em.getTransaction().begin();
            
        UUIDAutoGenEntity entity = new UUIDAutoGenEntity();
        entity.setName("Ultron");
        em.persist(entity);
            
        em.getTransaction().commit();
            
        UUID id = entity.getId();
        assertNotNull(id);

    } catch (AssertionError ae) {
        throw ae;
    } catch (Exception e) {
        e.printStackTrace();
        throw new RuntimeException(e);
    } finally {
        if (em.getTransaction().isActive()) {
            em.getTransaction().rollback();
        }
        em.close();
    }
}
```

Fails with the following:
```
java.lang.RuntimeException: java.lang.IllegalArgumentException: Invalid UUID string: 1
	at org.eclipse.persistence.jpa.test.uuid.TestUUIDGenerator.testBasicUUIDIdentity_AUTO_Generator(TestUUIDGenerator.java:329)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
[...]

PR #1655, opened 2022-08-17
[Master] ECL1625: Address incomplete UUID impl with AUTO generator

jgrassel:
For #1625

(also includes a small fix to SessionControlPanel as an errant character in a comment was causing my compiler to fail)


dazey3 (approved):
LGTM

rfelcman (approved):
LGTM

PR #1658, opened 2022-08-18
[3.0] Promotion build script fix - version qualifier and buildType property

rfelcman:
This PR fix some parameters in the _release.sh_ script to set correctly properties `buildRevision` and `buildType` in the _version.properties_ file.


lukasj (approved):
LGTM

Issue #1641, opened 2022-08-09
Function ceiling, floor, and round return wrong types


hantsy:
Glassfish 7.0.0-M7 + Built-in Derby Database
Java 17
Windows 10

I have tried to refresh my [JPA 3.1 examples](https://github.com/hantsy/jakartaee10-sandbox/tree/master/jpa) and test functionalities against the newest Glassfish 7.0.0 M7. 

The `Person` `salary` property is a `BigDecimal`.  But the result is a `Double`, obviously it is not my expected.

```java
cb.ceiling(root.get("salary")),
                    cb.floor(root.get("salary")),
                    cb.round(root.get("salary"), 1),
```

Tomas-Kraus:
This may be caused by many things. DB native function is being invoked in most of the cases, but Derby is a special case here. It does not support some of the functions defined in spec so they are emulated using math expressions. See [DerbyPlatform.java Line 474](https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/DerbyPlatform.java#L474)
This may be the cause of type change.
I can verify whether return type of the expression is set properly for those functions.

Tomas-Kraus:
Also make sure that you are using latest shapshot build of EclipseLink. It was not updated in GF yet.

hantsy:
I was using Glassfish 7.0 M7 built-in EclipseLink.

Tomas-Kraus:
I forced return type to follow number argument in CriteriaBuilderImpl functions definitions. Also added jUnits to verify returned type for ceiling, floor and round functions.

PR #1665, opened 2022-08-19
Issue #1641 - Fixed ceiling, floor and round return types.

Tomas-Kraus:


lukasj (approved):
LGTM

Next time AVOID using hamcrest - which is indirect dependency of junit (and as such seen as a long time junit bug) - on places it is not used

Issue #1668, opened 2022-08-23
PostgreSQL new error being logged on EclipseLink 4.0.0


dazey3:
Running with PostgreSQL + EclipseLink 4.0.0, I am seeing a new log message

```
[8/23/22, 15:20:52:524 CDT] 0000003a eclipselink                                                  I CWWJP9990I: [eclipselink] EclipseLink, version: Eclipse Persistence Services - 4.0.0.v202208221323
[8/23/22, 15:20:52:533 CDT] 0000003a eclipselink                                                  E CWWJP9992E: [eclipselink] Missing PostgreSQL10JsonPlatform from org.eclipse.persistence.pgsql module. JSON extension will not work properly. 
```

I am using bundles
```
org.eclipse.persistence:org.eclipse.persistence.core:4.0.0
org.eclipse.persistence:org.eclipse.persistence.jpa.jpql:4.0.0
org.eclipse.persistence:org.eclipse.persistence.jpa.modelgen.processor:4.0.0
org.eclipse.persistence:org.eclipse.persistence.jpa:4.0.0
```

dazey3:
https://github.com/eclipse-ee4j/eclipselink/blob/b8b009d799389faf74b9c8a5d96f58cf6884d54b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/localization/i18n/LoggingLocalizationResource.java#L369

https://github.com/eclipse-ee4j/eclipselink/blob/b8b009d799389faf74b9c8a5d96f58cf6884d54b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQL10Platform.java#L61-L73

It looks like `org.eclipse.persistence.platform.database.PostgreSQL10Platform` (which is in `org.eclipse.persistence.core`) is requiring `org.eclipse.persistence.pgsql` for PostgreSQL.

Could `org.eclipse.persistence.platform.database.PostgreSQL10Platform` instead check for `org.eclipse.persistence.json`? If `org.eclipse.persistence.json` exists, but it isn't `PostgreSQL10JsonPlatform`, then maybe it is `SessionLog.SEVERE`. But without `org.eclipse.persistence.json`, I think it should be `SessionLog.INFO`

lukasj:
[...]

PR #1670, opened 2022-08-24
Issue #1668 - Lowered log level and made messages less aggressive.

Tomas-Kraus:
Also fixed module info for pgsql and oracle modules.

### DIFF
--- a/bundles/eclipselink/src/main/java/module-info.java
+++ b/bundles/eclipselink/src/main/java/module-info.java
@@ -45,6 +45,8 @@ module eclipselink {
 
     requires static com.sun.xml.bind.core;
 
+    uses org.eclipse.persistence.internal.databaseaccess.spi.JsonPlatformProvider;
+
     exports org.eclipse.persistence.jpa.jpql;
     exports org.eclipse.persistence.jpa.jpql.parser;
     exports org.eclipse.persistence.jpa.jpql.utility.iterable;
--- a/foundation/org.eclipse.persistence.core/src/main/java/module-info.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/module-info.java
@@ -38,6 +38,8 @@ module org.eclipse.persistence.core {
     requires static jakarta.messaging;
     requires static jakarta.resource;
 
+    uses org.eclipse.persistence.internal.databaseaccess.spi.JsonPlatformProvider;
+
     exports org.eclipse.persistence;
     exports org.eclipse.persistence.annotations;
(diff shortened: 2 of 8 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/JsonPlatformManager.java
```diff
@@ -15,42 +15,88 @@
...
+            final Map<Class<? extends DatabasePlatform>, Supplier<DatabaseJsonPlatform>> fallbackMap, final String className) {
+        final Class<JsonPlatformProvider> providerClass = getProvider(className);
+        if (providerClass != null) {
+            try {
+                final JsonPlatformProvider provider  = providerClass.getConstructor().newInstance();
+                final Map<Class<? extends DatabasePlatform>, Supplier<DatabaseJsonPlatform>> platformsMap = provider.platforms();
+                if (platformsMap != null) {
+                    fallbackMap.putAll(platformsMap);
+                }
+            } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException | InstantiationException ex) {
```
lukasj, line 63: what about `ReflectiveOperationException`?

foundation/org.eclipse.persistence.core/src/main/java/module-info.java
```diff
@@ -38,6 +38,8 @@
     requires static jakarta.messaging;
     requires static jakarta.resource;
 
+    uses org.eclipse.persistence.internal.databaseaccess.spi.JsonPlatformProvider;
```
lukasj, line 41: so the core module will find it; what about eclipselink.jar?

Tomas-Kraus: Let me guess ...it won't find it.

bundles/eclipselink/src/main/java/module-info.java
```diff
@@ -45,6 +45,8 @@
 
     requires static com.sun.xml.bind.core;
 
+    uses org.eclipse.persistence.internal.databaseaccess.spi.JsonPlatformProvider;
```
lukasj, line 48: the flow of instructions in other descriptors is `requires`, `exports` and then `uses`. What is the reason for breaking the consistency here?

lukasj: see ie https://dev.java/learn/introduction-to-modules-in-java/

PR #1688, opened 2022-09-02
Oracle21Platform unused field removal to fix JPA remote test

rfelcman:
This PR fixes issue which happens if EntityManager API using a remote EntityManager and `org.eclipse.persistence.platform.database.oracle.Oracle21Platform` is specified.
`org.eclipse.persistence.platform.database.oracle.Oracle21Platform` had `private final OracleJsonFactory factory;` field which wasn't used but is serialized via RMI if remote EntityManager is used.
With this field following exception was thrown:
```
...
Internal Exception: java.rmi.UnmarshalException: error unmarshalling return; nested exception is: 
	java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: oracle.sql.json.OracleJsonFactory
[INFO] 
[ERROR] Tests run: 106, Failures: 1, Errors: 99, Skipped: 0
```
in
`mvn verify -pl :org.eclipse.persistence.jpa.testapps.remote -amd -Pstaging,oracle`

### DIFF
--- a/foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
+++ b/foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
@@ -21,7 +21,6 @@ import java.util.Hashtable;
 import java.util.Map;
 
 import oracle.jdbc.OracleType;
-import oracle.sql.json.OracleJsonFactory;
 import oracle.sql.json.OracleJsonValue;
 import org.eclipse.persistence.internal.databaseaccess.FieldTypeDefinition;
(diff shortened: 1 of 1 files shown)


foundation/org.eclipse.persistence.oracle/src/main/java/org/eclipse/persistence/platform/database/oracle/Oracle21Platform.java
```diff
@@ -34,15 +33,11 @@
  */
 public class Oracle21Platform extends Oracle19Platform {
 
-    // Oracle JSON factory.
-    private final OracleJsonFactory factory;
```
lukasj, line 38: @Tomas-Kraus should know whether this needs to be marked `transient` or not instead

Tomas-Kraus (approved):
LGTM.

PR #1689, opened 2022-09-06
[master] Hamcrest dependency removal

rfelcman:
This is minor test code refactoring to remove code dependency to hamcrest api. These tests with Hamcrest API should work in Maven environments as Hamcrest dependency is transitively loaded by JUnit but there should be problem in Ant and other non-Maven build systems.


lukasj (approved):
LGTM

PR #1702, opened 2022-09-12
Add defensive null check to stop null pointer exception

Mustard:
When joining tables (entities) that have a @Customizer (history policy) eclipselink will throw a NullPointerException

I have a [simple proof of concept project](https://github.com/Mustard/eclipselink-npe) that demonstrates the issue.  Ideally this could be added to the eclipselink project as a test, any help around where to add this test in the project would be appreciated as I couldn't even get the 2.7 branch to build locally and am unfamiliar with the project

This pull request would resolve this issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=526836

A possible other solution / improvement might be to initialise `asOfClause` in `org.eclipse.persistence.internal.expressions.DataExpression` such that it is not null in the first place

```
protected AsOfClause asOfClause = AsOfClause.NO_CLAUSE;
```

I have agreed to the Eclipse Contributor Agreement https://www.eclipse.org/user/mustard


lukasj:
wrt test - take a look at  [jpa/eclipselink.jpa.testapps](https://github.com/eclipse-ee4j/eclipselink/tree/master/jpa/eclipselink.jpa.testapps) and try to find an existing test this could be added to - `jpa.test.advanced` or `jpa.test.jpql` look like good candidates; it does not make much sense to add whole new project for this

Mustard:
Hi @lukasj 

Thanks for your help!

Took me a while to get back to it but I have added tests to the pull request, if there is anything else I can do to move this along let me know

robertpatrick:
@arjantijms   I reproduced the original problem and a new regression using current 5.x code with the PR’s production changes:

| Test | Current master | Full PR change | Two null guards only |
| ---- | ---- | ---- | ---- |
| Prepare the reported join | NPE | Pass | Pass |
| Historical joined query | Past parent + past child | Past parent + current child | Past parent + past child |

The problem is the new `AsOfClause.NO_CLAUSE` initializer. Previously, `null` allowed a child expression to inherit its parent’s historical timestamp. `NO_CLAUSE` prevents that inheritance, silently mixing historical and current data.

What the author should do:

1. Remove that initializer, retaining the two targeted null guards—the narrower variant passed both local tests.
2. Add a historical-join result test and close the test’s EntityManager. The existing preparation test is meaningful and correctly registered.
3. Refresh onto master and obtain clean CI, including the existing tests affected by adding history policies to Room/Door.

PR #1703, opened 2022-09-12
[master] Nightly build fix - aggregated JPA test reports

rfelcman:
This is change to catch up latest changes in JPA tests as tests where divided into multiple modules from origin `org.eclipse.persistence.jpa.test`. There are aggregated JPA test reports (Surefire, Failsafe) from `org.eclipse.persistence.jpa.test`, `org.eclipse.persistence.jpa.testapps` and `org.eclipse.persistence.jpa.jse.test`.
There is new Maven module `org.eclipse.persistence.jpa.test.report` which generate aggregated report output.


lukasj (commented):
will the report be produced even if the build fails in the middle?

rfelcman:
> will the report be produced even if the build fails in the middle?

It's refactored now. POM `org.eclipse.persistence.nightly` from `bundles/nightly/pom.xml` was removed from `org.eclipse.persistence.parent`/`org.eclipse.persistence.bundles` and it's used/called as a standalone module/project to prepare EclipseLink bundles and test results for publishing into nightly builds storage/page during Jenkins build only.
It prepare/aggregate JPA test report from `org.eclipse.persistence.jpa.test`, `org.eclipse.persistence.jpa.testapps` and `org.eclipse.persistence.jpa.jse.test` modules.

lukasj:
then I see no point in having ant wrapped by maven other than removing the need for having ant installed in the environment

lukasj (approved):
I'd rather see this in pure ant but if you think this is easier to use/maintain, I have no objections. LGTM

rfelcman:
I preferred use Maven as rest of the master branch using it, to keep consistency in:

- build qualifier (date, time, git commit id) used in published pages (handle it)
- Maven plugins are specified in `org.eclipse.persistence.parent/org.eclipse.persistence.bundles` used as a project parent POM (still one central place for the upgrades)
- Test report html page generation. To use same technology for JPA in `org.eclipse.persistence.nightly` like other html reports like CORE, SDO...

PR #1704, opened 2022-09-14
[master] Nightly build fix - Jenkins pipeline

rfelcman:
This fix related with #1703 and `etc/jenkins/build.groovy` pipeline. Due security reason "Publish to nightly" stage is executed under different container, than EclipseLink build and tests. There should be Maven, but in different version, than `el-build`container used for a build and tests.


lukasj (approved):
LGTM

PR #1707, opened 2022-09-19
[master] Fix ConcurrentModificationException in BatchFetchPolicy - backport from 2.7

### LINKED ISSUE #1506: BatchFetchPolicy.clone() cause memory leak and ConcurrentModifcationException

rfelcman:
Fixes #1506 . This change will fix the ConcurrentModificationException of the attributeExpression list, when running the same (batch)query with different entity managers.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/BatchFetchPolicy.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/BatchFetchPolicy.java
@@ -77,6 +77,9 @@ public class BatchFetchPolicy implements Serializable, Cloneable {
             dataResults.put(clone, list);
         }
         clone.setDataResults(dataResults);
+        if(this.attributeExpressions != null) {
+            clone.attributeExpressions = new ArrayList<>(this.attributeExpressions);
+        }
         return clone;
     }
 


lukasj (approved):
LGTM

Issue #1711, opened 2022-09-24
ClassCastException updating foreignkeys with identity generation


dazey3:
We are seeing a ClassCastException from EclipseLink 4.0.0. Setting `<property name="eclipselink.target-database-properties" value="supportsReturnGeneratedKeys=false" />` resolves the issue. It appears to be a bug in the GeneratedKeys support https://github.com/eclipse-ee4j/eclipselink/issues/1480 I added.

Entity:
```java
@Entity
public class StepExecutionEntity {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private long id;

    @ManyToOne
    @JoinColumn(name = "FK_ID")
    private StepExecutionEntity childEntity;
```

Test:
```java
    EntityManager em = emf.createEntityManager();
    try {
        StepExecutionEntity e1 = new StepExecutionEntity();
        e1.setChildEntity(e1);

        em.getTransaction().begin();
        em.persist(e1);
        em.getTransaction().commit();
```

Exception:
```
Caused by: java.lang.ClassCastException: org.eclipse.persistence.queries.SQLCall incompatible with java.lang.Integer
	at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.updateObject(DatasourceCallQueryMechanism.java:905)
	at org.eclipse.persistence.internal.queries.StatementQueryMechanism.updateObject(StatementQueryMechanism.java:462)
	at org.eclipse.persistence.internal.queries.CallQueryMechanism.updateForeignKeyFieldAfterInsert(CallQueryMechanism.java:435)
	at org.eclipse.persistence.internal.queries.DatabaseQueryMechanism.updateForeignKeyFieldAfterInsert(DatabaseQueryMechanism.java:795)
	at org.eclipse.persistence.internal.queries.DatabaseQueryMechanism.insertObjectForWrite(DatabaseQueryMechanism.java:471)
	at org.eclipse.persistence.queries.InsertObjectQuery.executeCommit(InsertObjectQuery.java:86)
	at org.eclipse.persistence.queries.InsertObjectQuery.executeCommitWithChangeSet(InsertObjectQuery.java:97)
	at org.eclipse.persistence.internal.queries.DatabaseQueryMechanism.executeWriteWithChangeSet(DatabaseQueryMechanism.java:326)
	at org.eclipse.persistence.queries.WriteObjectQuery.executeDatabaseQuery(WriteObjectQuery.java:61)
[...]

PR #1712, opened 2022-09-24
[master] Issue 1711: ClassCastException while updating foreignkeys with identity generation

dazey3:
for #1711


jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/sequence/TestIdentityGeneration.java
```diff
@@ -95,6 +96,29 @@ public void testPersistWithSecondaryTables() {
         }
     }
 
+    @Test
+    public void testPersistWithSecondaryTables2() {
```
lukasj, line 100: Would it make sense to add a reference to a bug this test is covering either as a comment or as a link here?

dazey3: @lukasj Sure! How is that?

lukasj (approved):
LGTM

PR #1733, opened 2022-10-25
Correctly handle times after noon

omolenkamp:
Some conversions in EclipseLink use `Calendar.HOUR` instead of `Calendar.HOUR_OF_DAY`, causing times after noon to be processed incorrectly. (`Calendar.HOUR` should only be used in conjunction with `Calendar.AM_PM`, in all other cases `Calendar.HOUR_OF_DAY` must be used)

In particular all sorts of conversions to `OffsetTime` and `OffsetDateTime` in `org.eclipse.persistence.internal.helper.ConversionManager` were producing wrong results.

This PR just fixes the bug, but the conversion code in `ConversionManager` could be streamlined greatly by using Java Time directly rather than converting  with a `Calendar`.

Code like this:
```
        } else if (sourceObject instanceof java.util.Date) {
            // handles sql.Time, sql.Date, sql.Timestamp
            Calendar cal = Helper.allocateCalendar();
            cal.setTime((java.util.Date) sourceObject);
            offsetDateTime = java.time.OffsetDateTime.of(
                    cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1, cal.get(Calendar.DAY_OF_MONTH),
                    cal.get(Calendar.HOUR_OF_DAY), cal.get(Calendar.MINUTE), cal.get(Calendar.SECOND), cal.get(Calendar.MILLISECOND) * 1000000,
                    java.time.ZoneOffset.ofTotalSeconds((cal.get(Calendar.ZONE_OFFSET) + cal.get(Calendar.DST_OFFSET)) / 1000));
            Helper.releaseCalendar(cal);
        }
```
could become something like:
```
        } else if (sourceObject instanceof java.util.Date) {
            offsetDateTime = ((java.util.Date) sourceObject).toInstant()
                    .atZone(getDefaultZoneOffset())
                    .toOffsetDateTime();
        }
```
This may change the offset, but it looks like that's the point of `getDefaultZoneOffset()`; if the offset should stay the same `ZoneId.systemDefault()` should be used instead of `getDefaultZoneOffset()`. (I'm not sure why `getDefaultZoneOffset()` is there in the first place, if conversions fail with the system default zone instead of `ZoneOffset.UTC`, the code is broken and should be fixed, rather than changing the zone that's used to work around the bug)

### DIFF
--- a/foundation/eclipselink.core.test/src/it/java/org/eclipse/persistence/testing/tests/identitymaps/cacheinvalidation/CacheInvalidationPolicyCloneTest.java
+++ b/foundation/eclipselink.core.test/src/it/java/org/eclipse/persistence/testing/tests/identitymaps/cacheinvalidation/CacheInvalidationPolicyCloneTest.java
@@ -54,7 +54,7 @@ public class CacheInvalidationPolicyCloneTest extends TestCase {
             assertFalse("Clone should not be the same instance", policy == policyClone);
             assertEquals("Clone should be of the same type", policy.getClass(), policyClone.getClass());
             assertEquals("Clone's expiry time should be the same", policy.getExpiryTime().getTimeInMillis(), policyClone.getExpiryTime().getTimeInMillis());
-            assertEquals("Clone's expiry time should be 01:02:03.004 - hour", 1, policyClone.getExpiryTime().get(Calendar.HOUR));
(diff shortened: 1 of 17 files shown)


lukasj (approved):
LGTM

PR #1760, opened 2022-12-06
[master] Oracle NoSQL cloud connection fix to set compartment

rfelcman:
Without this NoSQL instance (driver) from cloud environment should throw
`Caused by: oracle.nosql.driver.TableNotFoundException: .....` although table exist in the DB.


lukasj (approved):
LGTM

PR #1761, opened 2022-12-06
fix: issue with multiple aliases against same table

ryaneorth:
With the current implementation of `TableExpression#assignAlias` you cannot have multiple aliases against the same table. The code block below prevents the second reference from being added to the table alias list.

```
        if (this.baseExpression.isQueryKeyExpression()){
            QueryKeyExpression qkExpression = ((QueryKeyExpression)this.baseExpression);
            if (qkExpression.getTableAliases() != null && qkExpression.getTableAliases().keyAtValue(table) != null ) {
                return;
            }
        }
```

Removing the override of `assignAlias` in `TableExpression` and deferring to `DataExpression`'s implementation results in the expected behavior.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/TableExpression.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/TableExpression.java
@@ -45,7 +45,7 @@ public class TableExpression extends DataExpression {
     protected void assignAlias(DatabaseTable alias, DatabaseTable table) {
         if (this.baseExpression.isQueryKeyExpression()){
             QueryKeyExpression qkExpression = ((QueryKeyExpression)this.baseExpression);
-            if (qkExpression.getTableAliases() != null && qkExpression.getTableAliases().keyAtValue(table) != null ) {
+            if (!qkExpression.hasQueryKey && qkExpression.getTableAliases() != null && qkExpression.getTableAliases().keyAtValue(table) != null ) {
                 return;
             }
         }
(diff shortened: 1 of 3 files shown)


rfelcman:
Sorry but during first brief check this PR can't pass due a following reasons:

- It breaks current functionality please check details for *EclipseLink PR build / Test on JDK 17 (pull_request)* and *continuous-integration/jenkins/pr-merge* checks.
- For functional fixes/changes like this is good to create some unit test (it's some kind of bug description and avoids regression in the future)

ryaneorth:
Thank you @rfelcman ! I see documentation on how to run the tests [here](https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.testapps/README.md), but is there any guidance on how to run these tests individually in an IDE to debug through them to identify the source of the failure(s)?

rfelcman:
About run tests individually it's described at https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.testapps/README.md to pass to the IDE command like `mvn test -pl :org.eclipse.persistence.jpa.testapps.jpql -Dtest=JUnitJPQLDateTimeTest -Pmysql` like in IDEA IteliJ run profiles.
In this PR its about `org.eclipse.persistence.core.test` module not jpa.
Example command is
`mvn test -pl :org.eclipse.persistence.core.test -Pmysql -Dtest=MapCollectionsTestModel -Ddb.platform=org.eclipse.persistence.platform.database.MySQLPlatform -Ddb.driver=com.mysql.cj.jdbc.Driver -Ddb.url=jdbc:mysql://localhost:3306/ecltests?allowPublicKeyRetrieval=true -Ddb.user=root -Ddb.pwd=root
`
but DB property values depends on Your local environment.

ryaneorth:
Thank you for your guidance and patience with me on this issue, @rfelcman ! I have implemented a solution that solves the original problem while not leading regressions. I have also added a test case for the scenario, which fails without the associated change to TableExpression.

rfelcman (approved):
LGTM

ryaneorth:
Thank you @rfelcman ! Out of curiosity, when will the next patch release be for 4.x.x?

rfelcman:
Plan is end of Januray 2023
https://projects.eclipse.org/projects/ee4j.eclipselink/releases/4.0.1
Maven Snapshot based on Nightly build
https://ci.eclipse.org/eclipselink/job/eclipselink-nightly-master/
at
https://jakarta.oss.sonatype.org/content/repositories/snapshots/org/eclipse/persistence/eclipselink/4.0.1-SNAPSHOT/
during 1 hour if all tests will pass

PR #1764, opened 2022-12-08
[master] Oracle NoSQL Database - Lob type fix

rfelcman:
This is fix when Lob type is loaded from the database. There are some additional tests too to verify various JPA/NoSQL data types.


lukasj (approved):
LGTM

PR #1765, opened 2022-12-15
[master] New WildFly EE platform

rfelcman:
This change add WildFly server into JEE platforms supported by EclipseLink.
It should better reflect changes in WildFly as latest WildFly (27.0.0.Final) support Jakarta EE10 and JBoss (EAP 7.4) Jakarta EE 8.
There are test dependencies update to the latest WildFly 27.0.0.Final.


lukasj (approved):
LGTM, just update cp headers before merge. Thx!

PR #1769, opened 2022-12-23
[3.0] Class read fallback needs to reread the stream from the beginning - build fix

rfelcman:


lukasj (approved):
ups, thanks

PR #1777, opened 2023-01-06
[2.7] Jenkins environment - switch to JDK 11

rfelcman:
This is technical switch into JDK 11. Output will be still compatible with JDK 8.0 due Tycho setting in `.settings/org.eclipse.jdt.core.prefs`.


lukasj (approved):
LGTM

Issue #1771, opened 2023-01-04
Can't use UUID primary key with JPA 3.1 on PostgreSQL


mthmulders:
I am building an application with Jakarta EE 10 and JPA 3.1. The JPA-implementation is provided by OpenLiberty: EclipseLink 4.0.0.v202210051929.

My entity looks like this:

```java
@Entity
@NamedQueries({
    @NamedQuery(name = "Car.findById", query = "select c from Car c where c.id = :id")
})
@Table(name = "car")
public class Car {
    @Id
    @GeneratedValue(strategy = GenerationType.UUID)
    private UUID id;
    private String model;
    private Double price;
    // getters and setters omitted for brevity
}
```

The repository looks like this:

```java
public class CarRepository {
    private EntityManager em;

    public CarRepository(EntityManager em) {
        this.em = em;
    }

    public void store(Car car) {
        em.merge(car);
    }

    public Optional<Car> findById(UUID id) {
        return em.createNamedQuery("Car.findById", Car.class)
                .setParameter("id", id)
                .getResultStream()
                .findFirst();
    }
}
```

And finally, the PostgreSQL table is created using

```sql
create table car
(
    id uuid primary key,
    model varchar not null,
    price real not null
);
```

Now, I want to store a `Car` instance:

```java
var car = new Car();
car.setId(UUID.randomUUID());
car.setModel("Model");
car.setPrice(10_000d);

repository.store(car);
```

But it fails with

```
Internal Exception: org.postgresql.util.PSQLException: ERROR: operator does not exist: uuid = character varying
  Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
  Position: 44
Error Code: 0
Call: SELECT ID, MODEL, PRICE FROM car WHERE (ID = ?)
	bind => [6c7f5f3d-2c71-4f2e-a79b-ca6a804670bc]
Query: ReadObjectQuery(referenceClass=Car )
```

The [attached project](https://github.com/eclipse-ee4j/eclipselink/files/10345227/eclipselink-jpa31-postgresql-uuid.zip) is a simplified version (without OpenLiberty) but illustrates the same problem.

If I understand [the specification](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1.pdf) correctly, it should be possible to declare fields with type `UUID` without having to write a `AttributeConverter` anymore.

[...]

PR #1778, opened 2023-01-07
Fix UUID handling for PostgreSQL

mthmulders:
Fix for #1771. Seems PostgreSQL doesn't like it when a `UUID` value is passed as a `String` - using `setObject` ensures the `UUID` value is passed as a `UUID`.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQLPlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQLPlatform.java
@@ -531,7 +531,7 @@ public class PostgreSQLPlatform extends DatabasePlatform {
         return "; END ; $$ LANGUAGE plpgsql;";
     }
 
-     /**
+    /**
      * INTERNAL: Used for sp calls.  PostGreSQL uses a different method for executing StoredProcedures than other platforms.
      */
     @Override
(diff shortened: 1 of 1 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/PostgreSQLPlatform.java
```diff
@@ -749,4 +753,40 @@ protected void setNullFromDatabaseField(DatabaseField databaseField, PreparedSta
...
+            statement.setObject(index, parameter, Types.OTHER);
+        } else {
+            super.setParameterValueInDatabaseCall(parameter, statement, index, session);
+        }
+    }
+
+    /**
+     * INTERNAL
+     * Set the parameter in the JDBC statement at the given index in case it is a UUID value.
+     */
```
rfelcman, line 781: Please add there for JavaDoc `* @param ....` like in the method above.

rfelcman (commented):
Please update copyright year in file header.

rfelcman (approved):
LGTM

mthmulders:
@rfelcman Please apologise, I see that I didn't push the Javadoc fix. Would you like to receive a separate PR for that?

rfelcman:
Please do it, it should be better.

mthmulders:
Please see https://github.com/eclipse-ee4j/eclipselink/pull/1795.

PR #1810, opened 2023-02-10
Fix injection issue (HK2) for MoxyJsonProvider

senivam:


moxy/org.eclipse.persistence.moxy/src/main/java/module-info.java
```diff
@@ -43,6 +43,7 @@
     exports org.eclipse.persistence.jaxb.rs;
     exports org.eclipse.persistence.jaxb.xmlmodel;
     opens org.eclipse.persistence.jaxb.xmlmodel to jakarta.xml.bind;
+    opens org.eclipse.persistence.jaxb.rs; //required for MOXyJSON provider injection
```
rfelcman, line 46: Just a question. Is chance to specify `to <module name>` there?

senivam: I would rather not. As soon as it is the public API, it shall be open for reflection to anyone. If the ``to`` part is specified, only that framework will be allowed to inject the provider. What about CDI then?

rfelcman (approved):
LGTM

PR #1477, opened 2022-04-08
[master] ASM separation

rfelcman:
Refactor EL to allow usage booth ASM implementations (standard provided by `org.ow2.asm` and internal `org.eclipse.persistence.asm`).

Main reason behind this enhancement is quickly adopt new Java version as Java release cycle is quicker now.
Usage of ASM implementation is controlled by new system property `eclipselink.asm.service` . There are following values possible
- eclipselink - use ASM implementation from  `org.eclipse.persistence.asm` project
-	ow2		  -	use ASM implementation from  `org.ow2.asm` project (DEFAULT)
Various EclipseLink tests should be executed with this system property e.g.
```
mvn verify -pl :org.eclipse.persistence.jpa.jse.test -P mysql -Declipselink.asm.service=ow2
mvn verify -pl :org.eclipse.persistence.jpa.jse.test -P mysql -Declipselink.asm.service=eclipselink

mvn test -pl :org.eclipse.persistence.moxy -P test-moxy-lrg -Declipselink.asm.service=ow2
mvn test -pl :org.eclipse.persistence.moxy -P test-moxy-lrg -Declipselink.asm.service=eclipselink

mvn clean verify -pl :org.eclipse.persistence.jpa.test -P test-jpa-lrg,mysql -Declipselink.asm.service=ow2
mvn clean verify -pl :org.eclipse.persistence.jpa.test -P test-jpa-lrg,mysql -Declipselink.asm.service=eclipselink

mvn verify -pl :org.eclipse.persistence.jpa.testapps -amd -P mysql -Declipselink.asm.service=ow2
mvn verify -pl :org.eclipse.persistence.jpa.testapps -amd -P mysql -Declipselink.asm.service=eclipselink
```
Code in this enhancement should be divided into following parts:
- Adapter/wrapper classes in `org.eclipse.persistence.core` module in `org.eclipse.persistence.asm` package. There are classes which translate calls from EclipseLink into ASM implementation.
Code where is ASM used is pointed to `org.eclipse.persistence.asm` Java package.
- Required code changes in the rest of EclipseLink code base to use adapter/wrapper classes from the previous point. Major part of these changes is 
`org.eclipse.persistence.internal.libraries.asm.Opcodes.ACC_PUBLIC` -> `org.eclipse.persistence.asm.Opcodes.valueInt("ACC_PUBLIC")`
or
[...]

PR #1820, opened 2023-03-01
Correct clobbered message from git rebase for (#1477)

jgrassel:
It looks like merging #1477 clobbered the message num 7357 (`"The \"[{0}]\" URL for \"[{1}]\" resource does not belong to a valid persistence root (as per Jakarta Persistence Specification)"`) in ValidationExceptionResource.java by mistake (there are two entries for the same message number).  Pushing a correction update.


rfelcman (approved):
LGTM

Issue #1815, opened 2023-02-15
Update queries to collections of basic types incorrectly transforms object types


dazey3:
When executing an UpdateQuery, that updates a Collection of basic types (not embeddables), EclipseLink incorrectly transforms the object type. This is particularly bad because the Update query is successfully executed, but when the value is retrieved from the database later, EclipseLink can't properly convert the value back into the expected type.

Entity:
```java
    @Entity
    public class City {
        @Id 
        public String name;
        public Set<Integer> areaCodes;
    }
```

Test:
```java
    Query q = em.createQuery("UPDATE City o SET o.areaCodes=?1 WHERE o.name=?2");
    q.setParameter(1, Set.of(567));
    q.setParameter(2, "Denver");
    q.executeUpdate();
...
    City city = em.find(City.class, "Denver");
```

Exception:
```
Exception Description: Trying to set value [563] for instance variable [areaCodes] of type [java.util.Set] in the object.  The specified object is not an instance of the class or interface declaring the underlying field, or an unwrapping conversion has failed.
Internal Exception: java.lang.IllegalArgumentException: Can not set java.util.Set field test.jakarta.data.jpa.web.City.areaCodes to java.lang.Integer
Mapping: org.eclipse.persistence.mappings.DirectToFieldMapping[areaCodes-->CITY.AREACODES]
Descriptor: RelationalDescriptor(test.jakarta.data.jpa.web.City --> [DatabaseTable(CITY)])
	at org.eclipse.persistence.exceptions.DescriptorException.illegalArgumentWhileSettingValueThruInstanceVariableAccessor(DescriptorException.java:703)
	at org.eclipse.persistence.internal.descriptors.InstanceVariableAttributeAccessor.setAttributeValueInObject(InstanceVariableAttributeAccessor.java:188)
	at org.eclipse.persistence.mappings.DatabaseMapping.setAttributeValueInObject(DatabaseMapping.java:1652)
	at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.buildCloneFromRow(AbstractDirectMapping.java:207)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildAttributesIntoWorkingCopyClone(ObjectBuilder.java:2018)
	at org.eclipse.persistence.internal.descriptors.ObjectBuilder.buildWorkingCopyCloneFromRow(ObjectBuilder.java:2271)
[...]

PR #1842, opened 2023-03-21
[master] Issue 1815: Transforming object types for collections

dazey3:
for #1815

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/CollectionExpression.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/CollectionExpression.java
@@ -38,12 +38,32 @@ public class CollectionExpression extends ConstantExpression {
 
     @Override
     public void printSQL(ExpressionSQLPrinter printer) {
-        Object value = this.value;
+        Object objectValue = this.value;
         if(this.localBase != null) {
-            value = this.localBase.getFieldValue(value, getSession());
+            if(objectValue instanceof Collection) {
+                Collection<?> values = (Collection<?>)objectValue;
+                Vector<Object> fieldValues = new Vector<>(values.size());
+                for (Object value : values) {
+                    Object translated = value;
+                    if (!(value instanceof Expression)){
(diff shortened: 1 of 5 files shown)


arjantijms:
@dazey3 it's been a while, but this PR is still in draft mode and there's merge conflicts now.

What would you like to do with this one? Shall we close it?

robertpatrick:
@dazey3 @arjantijms I reproduced #1815 on the current master, and this PR's change fixes the bulk-update problem. Unfortunately, it also breaks some existing IN queries, so I don't think it's safe to merge yet.

I compared current master with the PR's production changes on H2 and Derby. Both databases produced the same outcomes:

| Local test | Current master |  With this change |
| ---- | ---- | ---- |
| Bulk-update a serialized or converter-backed Set | Fails during update or subsequent read | Correct update and readback |
| Bound IN list of integer IDs, enums, or converter-backed values | Correct results | Conversion exceptions |
| Bound IN list of related entities | Correct results | Entities reach JDBC instead of their IDs; query fails |
| Bound IN list of Sets | Fails | Still fails |
| Existing inline scalar/relationship IN and equality controls | Correct results | Unchanged |

The underlying problem is that ParameterExpression still passes the whole bound collection to QueryKeyExpression. After this change, the scalar converter receives that collection instead of each candidate value. The relationship path also loses entity-ID extraction. Moving the loop into CollectionExpression handles inline lists, but doesn't complete the bound-parameter path.

What we need:

| Action | Reason |
| ---- | ---- |
| Distinguish one collection-valued basic value from an IN list of candidates | Preserve the bulk-update fix while retaining per-candidate conversion and entity-ID extraction. |
| Complete the nested-collection test and change res.get(1) to res.get(0) | The query still fails; the index would also be invalid after asserting one result. Assert the expected city and exact set contents. |
| Add focused regression coverage alongside the existing IN tests |  Cover the scalar, converter-backed, and relationship parameter cases above. The new tests already fit the existing JSE integration-test runner. |
| Rebase, finish the draft, and obtain clean existing CI | The branch conflicts, and no current-head CI results are reported. |

This fixes a real issue and is worth finishing, but a rebase alone won't resolve the merge blockers.

These local results were focused local comparisons.

PR #1852, opened 2023-04-03
[3.0] SDO potential memory leak fix - backport from master

rfelcman:
This change solve potential memory leak if SDO component is used in JEE environment. There is refactor of `SDOHelperContext.helperContexts` map. Instead of current `ConcurrentHashMap` is `CacheIdentityMap` used which supports Least Recently Used (LRU) strategy. This one should be controlled by new SDO system property `eclipselink.sdo.helper.contexts.max.size`.

(cherry picked from commit 3271f25350dc500f52016954e60ccbe588fc0343)


lukasj (approved):
LGTM

PR #1853, opened 2023-04-03
[2.7] SDO potential memory leak fix - backport from master

rfelcman:
This change solve potential memory leak if SDO component is used in JEE environment. There is refactor of `SDOHelperContext.helperContexts` map. Instead of current `ConcurrentHashMap` is `CacheIdentityMap` used which supports Least Recently Used (LRU) strategy. This one should be controlled by new SDO system property `eclipselink.sdo.helper.contexts.max.size`.


sdo/eclipselink.sdo.test/src/org/eclipse/persistence/testing/sdo/helper/SDOHelperContextTest.java
```diff
@@ -17,6 +17,7 @@
 import commonj.sdo.helper.HelperContext;
 import junit.framework.TestCase;
 import org.eclipse.persistence.sdo.helper.SDOHelperContext;
+import static org.eclipse.persistence.sdo.SDOSystemProperties.SDO_STRICT_TYPE_CHECKING_PROPERTY_NAME;
```
lukasj, line 20: my preference is to avoid static imports (just sayin')

lukasj (approved):
LGTM

PR #1844, opened 2023-03-29
[master] SDO potential memory leak fix

rfelcman:
This change solve potential memory leak if SDO component used in JEE environment.
There is refactor of `SDOHelperContext.helperContexts` map. Instead of current `ConcurrentHashMap` is `CacheIdentityMap` used which supports Least Recently Used (LRU) strategy. This one should be controlled by new SDO system property `eclipselink.sdo.helper.contexts.max.size`.

PR #1855, opened 2023-04-04
[master] SDO potential memory leak fix 2.0

rfelcman:
This change solve potential memory leak if SDO component is used in JEE environment.
There is refactor of `SDOHelperContext.helperContexts` map content (nested map). Instead of current `ConcurrentHashMap` is `CacheIdentityMap` used which supports Least Recently Used (LRU) strategy. This one should be controlled by new SDO system property `eclipselink.sdo.helper.contexts.max.size`.

Previous change #1844 will be reverted by this.


lukasj (approved):
LGTM

PR #1880, opened 2023-05-17
[master] JDK 21 build and test support

rfelcman:
There are following changes due JDK21:

- @SuppressWarnings("this-escape") mainly in `jpa.jpql`, `dbws.builder.oracle.test` modules

- `<comp.xlint>-Xlint:all,-this-escape</comp.xlint>` in _pom.xml_ in `core.test.framework` as compiler hangs there without any additional message
- some package renaming in tests in `corba` module as JDK21 doesn't like same packages in different jars in classpath/doesn't find classes (testing jars are on the classpath not module path)
- `SecurityManager` tests are not called in JDK21
- GitHub Action JDK21-EA added


lukasj (approved):
LGTM

PR #1881, opened 2023-05-19
[2.7] JDK 21 build and test support

rfelcman:
There are following changes due JDK21:

- Tycho upgrade into 3.0.4 in core modules like jpa.jpql, core, jpa, sdo ... but only for JDK17+, for JDK11 remains origin 1.3.0 by Maven profile
- `maven.compiler.release` remains at 8
- Antbuild script change to accept JDK range from 1.8 to 21
- Antbuild script sharedlib.xml change to use external JavaScript engine as this one was removed from higher versions of the JDKs
- Weld upgrade to make JPA.JSE test (Dead Lock) functional on JDK 17+
- GitHub Action with JDK 17 and 21-EA (Eclipse PR verification job remains with JDK 11)


foundation/eclipselink.core.test/antbuild.xml
```diff
@@ -562,9 +565,10 @@
 
     <!-- Default test target, runs test.class from test.properties (default is LRG). -->
     <target name="test" depends="init-testing">
+        <echo message="is.less.21.jdk: ${is.less.21.jdk}"/>
         <test include="**/junit/**/*Test.java"/>
         <test include="${test.class}"/>
-        <test include="org.eclipse.persistence.testing.tests.security.SecurityTestModel"/>
+        <test include="org.eclipse.persistence.testing.tests.security.SecurityTestModel" if:true="${is.less.21.jdk}"/>
```
lukasj, line 571: the `if` here can be avoided by having an extra target with `if`/`unless`

rfelcman: OK resolved by recommended way.

lukasj (approved):
LGTM

PR #1892, opened 2023-06-07
[4.0] Maven plugins and dependencies update

rfelcman:
It leads into two code/build changes:
- package refactoring to avoid naming collisions across modules
- jar-test-sources execution removal to avoid duplicity calls


lukasj (approved):
LGTM

PR #1896, opened 2023-06-12
[4.0] maven-surefire-report-plugin title settings change

rfelcman:
This change fixes different location (XPath) in test HTML reports as maven-surefire-report-plugin uses different site skin.


lukasj (approved):
LGTM

PR #1906, opened 2023-07-03
[master] Maven plugins and dependencies update

rfelcman:
Plus:
- fixes different location (XPath) in test HTML reports as maven-surefire-report-plugin uses different site skin
- jar-test-sources unnecessary calls removal
- document generation (asciidoctorj) plugin and dependencies update (latest are not functional)

### DIFF
--- a/bundles/nightly/pom.xml
+++ b/bundles/nightly/pom.xml
@@ -72,7 +72,7 @@
                 <artifactId>maven-antrun-plugin</artifactId>
                 <executions>
                     <execution>
-                        <phase>process-resources</phase>
+                        <phase>generate-resources</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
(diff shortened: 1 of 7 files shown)


bundles/nightly/pom.xml
```diff
@@ -123,7 +123,30 @@
...
+                    <execution>
+                        <id>prepare-nightly-build-dir-test-reports-resources</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <copy todir="${project.build.directory}${nightlyDir}/Eclipse/css/">
+                                    <fileset dir="../../foundation/eclipselink.core.test/target/${project.build.testReports.subdirectory}/css/"/>
```
lukasj, line 144: can the content be taken from test-assembly (or some other build-time artifact) to avoid depending on some, possibly random, paths?

rfelcman: I tried find some solution with origin XML reports (see _target/failsafe-reports_), but there is no any single overall XML report. Just only number of XML reports one per each test suite. I didn't find any tool which should merge them (surefire + failsafe) except `maven-surefire-report-plugin`. But `maven-surefire-report-plugin` can produce only HTML aggregated output, but not XML.

lukasj (approved):

PR #1915, opened 2023-07-24
[master] SDO helper context removal

rfelcman:
This change allows remove specified SDO helper context. It offers to applications or application frameworks like Oracle ADF to control/remove unused helper context from the memory as in case of older JVMs (1.8.0) high amount of Metaspace memory should be occupied.
Application or application framework must recreate helper context if it will be used again.


lukasj (approved):
LGTM

PR #1923, opened 2023-08-14
return user defined type from functionexpression in core, fix handling of XmlEnum in moxy

lukasj:
* fix handling of moxy dynamic classes on JPMS
* avoid instanceof where possible (in core only) 
* small step towards moxy compatibility with android
* fix imports in core module
* Deprecate inefficient Helper.getShortClassName methods and replace their usages

(check this commit by commit, the last one updates imports only)


moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/JAXBContext.java
```diff
@@ -1478,7 +1477,7 @@ private QName getSchemaTypeForTypeMappingInfo(Type type) {
                 if (name == null) {
                     Class<?> theClass = (Class) type;
                     //Change default for byte[] to Base64 (JAXB 2.0 default)
-                    if (type == CoreClassConstants.ABYTE || type == CoreClassConstants.APBYTE || type == Image.class || type == Source.class || theClass.getCanonicalName().equals("jakarta.activation.DataHandler")) {
+                    if (type == CoreClassConstants.ABYTE || type == CoreClassConstants.APBYTE || theClass.getCanonicalName().equals("java.awt.Image") || type == Source.class || theClass.getCanonicalName().equals("jakarta.activation.DataHandler")) {
```
rfelcman, line 1480: Possible NPE `getCanonicalName()` should return `null` in some cases, but I'm not sure if it happens in this context.

moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/javamodel/reflection/JavaClassImpl.java
```diff
@@ -177,9 +177,13 @@ public Collection<JavaField> getDeclaredFields() {
...
         for (Field field : fields) {
-            if (!field.trySetAccessible()) {
-                AbstractSessionLog.getLog().log(SessionLog.FINE, SessionLog.MISC, "set_accessible_in",
-                        "field", field.getName(), jClass.getName());
+            try {
+                if (!field.trySetAccessible()) {
+                    AbstractSessionLog.getLog().log(SessionLog.FINE, SessionLog.MISC, "set_accessible_in",
+                            "field", field.getName(), jClass.getName());
+                }
+            } catch (NoSuchMethodError nsme) {
```
rfelcman, line 185: Just to be curious. In which exact cases this `catch` happens?

moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/JAXBContext.java
lukasj: * type will never be null, see also the next (pre-existing) part of the condition
* they are moxy specific, so oxm is not the right place for this; the consolidation of this sort of constants spread through `JAXBContext`, `MappingsGenerator`, `AnnotationsProcessor` and `SchemaGenerator` deserves separate PR/commit, I believe

moxy/org.eclipse.persistence.moxy/src/main/java/org/eclipse/persistence/jaxb/javamodel/reflection/JavaClassImpl.java
lukasj: dalvik/android does not have the `trySetAccessible()` method from SE 9+

Issue #1885, opened 2023-06-01
Implement new JPQLGrammar for upcoming Jakarta Persistence 3.2


rfelcman:
It includes following tasks:

- [x] https://github.com/jakartaee/persistence/issues/397
       https://github.com/eclipse-ee4j/eclipselink/pull/1903
- [x] https://github.com/jakartaee/persistence/issues/396
       https://github.com/eclipse-ee4j/eclipselink/pull/1924
- [x] https://github.com/jakartaee/persistence/issues/318
       https://github.com/eclipse-ee4j/eclipselink/pull/1933
- [x] https://github.com/jakartaee/persistence/issues/398
       https://github.com/eclipse-ee4j/eclipselink/pull/1937
- [x] https://github.com/jakartaee/persistence/issues/395
       https://github.com/eclipse-ee4j/eclipselink/pull/1937

PR #1924, opened 2023-08-17
[master] Jakarta Persistence 3.2 new feature - JPQL functions REPLACE()

rfelcman:
This change contains following changes:

- New JPQL String function REPLACE() for string content replacement (find and replace)
     Tests are available in the JPQL parser and JPA JSE test module
      EclipseLink issue is described at #1885 and Jakarta Persistence specification request at the https://github.com/jakartaee/persistence/issues/318
      This feature is not functional on Apache Derby Database as this DB doesn't have related implementation in SQL, or anything similar what should by used as a workaround in the `DerbyPLatform` class.


jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQuerySyntaxStringTests.java
```diff
@@ -211,4 +213,129 @@ public void testString_concatOperatorWhere02() {
...
         }
     }
+
+    @Test
+    public void testString_replaceFunctionSelect01() {
+        if (emf == null)
+            return;
+
+        //REPLACE is not implemented in Apache Derby DB
+        if (emf.unwrap(Session.class).getPlatform().isDerby()) {
```
lukasj, line 223: can we get this emulated somehow? (at least file an issue so it can be investaged..)

jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/ReplaceExpression.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2006, 2023 Oracle and/or its affiliates. All rights reserved.
```
lukasj, line 2: new files should have 2023 only...

jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/ReplaceExpression.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The second and third arguments of the <b>REPLACE</b> function denote the pattern to replace and
+ * replacement string. These arguments are strings. The <b>REPLACE</b> function returns a string.
+ * <br>
+ * Jakarta Persistence 3.2:
+ * <div><b>BNF</b> ::= <code>REPLACE(string_expression, string_expression , string_expression)</code></div>
+ *
+ * <div>Example: <b>UPDATE</b> Employee e <b>SET</b> e.firstName = <b>REPLACE</b>('TopLink Workbench', 'Top', 'Eclipse')</div>
+ *
+ * @version 4.1
```
lukasj, line 27: I'd avoid using `@version`; using `@since` is OK

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQuerySyntaxStringTests.java
rfelcman: I tried to find any SQL function in Derby
https://db.apache.org/derby/docs/10.16/ref/
https://db.apache.org/derby/docs/10.16/ref/rrefsqlj29026.html
but I don't see there any String function which should search String within String or something near String replacement.

About emulation in `org.eclipse.persistence.platform.database.DerbyPlatform` like method `derbyPowerOperator()` I'm missing any basic String functions to do it as I mentioned.

lukasj (approved):

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/query/TestQuerySyntaxStringTests.java
lukasj: ok

Issue #836, opened 2020-06-18
No way to execute a query with literal question mark in it


jaroslawr:
? is intepreted by JPQL to bind parameters, but PostgreSQL has operators for JSONB that use a literal question mark (?, ?|, ?&) and there does not seem to be an obvious documented way to pass through a '?' so that EclipseLink does not try to interpret it and passes it through to the database, in particular with native queries and with the sql() function. Documentation in both cases does not mention any way of escaping the question mark to prevent it from being interpreted by eclipselink:

https://www.eclipse.org/eclipselink/documentation/2.4/concepts/queries005.htm
https://www.eclipse.org/eclipselink/documentation/2.4/jpa/extensions/j_sql.htm

PR #1935, opened 2023-08-30
Allow escaping of '?' using '??' in SQL expression

pvarga88:
Using [SQL operator,](https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#SQL) it is not possible to use '?' as a native operator (e.g. on Postgres). This PR is to allow using '??' as an escape mechanism, similar to the escaping mechanism used by [Postgres JDBC](https://jdbc.postgresql.org/documentation/query/#using-the-statement-or-preparedstatement-interface).

This is to fix issue #836

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/expressions/Expression.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/expressions/Expression.java
@@ -1994,6 +1994,14 @@ public abstract class Expression implements Serializable, Cloneable {
             v.add(sql.substring(start, sql.length()));
         }
         anOperator.printsAs(v);
+        //Postgres expects '??' as an escape mechanism for '?' in parameterized queries
+        //https://jdbc.postgresql.org/documentation/query/#using-the-statement-or-preparedstatement-interface
+        //On other platforms, replace ?? with ? in code which is passed as a part of SQL into DB
+        if (getSession() == null || !getSession().getPlatform().isPostgreSQL()) {
+            for (int i = 0; i < anOperator.getDatabaseStrings().length; i++) {
+                anOperator.getDatabaseStrings()[i] = anOperator.getDatabaseStrings()[i].replace("??", "?");
+            }
+        }
         anOperator.bePrefix();
         anOperator.setNodeClass(ClassConstants.FunctionExpression_Class);
 
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/expressions/Expression.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/expressions/Expression.java
@@ -1981,6 +1981,11 @@ public abstract class Expression implements Serializable, Cloneable {
         int start = 0;
         int index = sql.indexOf('?');
         while (index != -1) {
+            // ? can be escaped as ?? to support ? as a native SQL operator (e.g. on Postgres)
+            if (index < sql.length() - 1 && sql.charAt(index + 1) == '?') {
+                index = sql.indexOf('?', index + 2);
+                continue;
+            }
             v.add(sql.substring(start, index));
             start = index + 1;
             index = sql.indexOf('?', start);
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java
@@ -254,4 +254,32 @@ public class TestJson implements JsonTestConverter.ConverterStatus {
         }
     }
 
+    @Test
+    public void testEscapedQuestionMarkInSQLOperator() {
+        EntityManager em = emf.createEntityManager();
+
+        if (emf.unwrap(Session.class).getPlatform().isOracle()) {
+            JsonValue value = Json.createObjectBuilder()
+                    .add("id", "1007")
+                    .build();
+            try {
+                em.getTransaction().begin();
+                JsonEntity e = new JsonEntity(1007, value);
+                em.persist(e);
+                em.flush();
+                em.getTransaction().commit();
+                em.clear();
+
+                JsonEntity dbValue = em.createQuery(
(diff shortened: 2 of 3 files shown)


pvarga88:
Is there anything I need to do to have this looked at by maintainers? or just patience?

rfelcman:
Could You please provide there examples of SQL (jsonb) query where `?` is used as an operator?
Maybe JPA based example should be nice too?
These should be used as starting point for the tests which we want for functional changes.
I'm not sure if this code change related with Postgres is placed to the right area. It should be stored in Postgres platform area.
But I'm not sure if `org.eclipse.persistence.internal.databaseaccess.Platform` offers some API for this.
Some Postgres JSON features were introduced by #1389 .

pvarga88:
The fix is not specific to Postgres, nor JSON, really. I just happen to run into this because Postgres uses '?' as an operator in many of its operators on JSONB:
https://www.postgresql.org/docs/11/functions-json.html

Oracle also uses '?' as an operator within its JSON_EXISTS
https://docs.oracle.com/en/database/oracle/oracle-database/21/adjsn/condition-JSON_EXISTS.html

As for an example, the following JPA query is not possible on Postgres without this fix:
SQL('myentity.myjsonfield->stanza ?? ?', :searchval")

↑ Note that I used two question marks where I want a literal question mark, rather than specifying a parameter in a parameterized query. This works with my proposed fix in this PR.

rfelcman:
Please extend this PR with test. Good candidate where to put it into new method is 
https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java

There is some example query against Oracle
`SELECT je.value FROM jsonentity je WHERE json_exists(je.value, '$?(@.id == 1001)');`

Example how to restrict test/code against some platform
https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/criteria/TestMathFunctions.java#L555
https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/criteria/TestDateTimeFunctions.java#L99

pvarga88:
I'll give it a go but I don't have the full environment to test the tests so it may take a few trial and error commits.

pvarga88:
Done. Test case added.

rfelcman (commented):
Please update copyright year in _TestJson.java_ into 
`....Copyright (c) 2022, 2023....`

jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/json/TestJson.java
```diff
@@ -254,4 +255,32 @@ public void testSelectJsonInWhereCondition() {
...
+            try {
+                em.getTransaction().begin();
+                JsonEntity e = new JsonEntity(1007, value);
+                em.persist(e);
+                em.flush();
+                em.getTransaction().commit();
+                em.clear();
+
+                JsonEntity dbValue = em.createQuery(
+                    "SELECT v.value FROM JsonEntity v WHERE SQL('JSON_EXISTS(?, ''$??(@.id == 1007)'')', v.value)", JsonEntity.class)
```
rfelcman, line 275: I think, that correct query should be `SELECT v FROM JsonEntity.....` not `SELECT v.value FROM JsonEntity.....`.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/expressions/Expression.java
```diff
@@ -1981,6 +1981,11 @@ public Expression sql(String sql, List arguments) {
         int start = 0;
         int index = sql.indexOf('?');
         while (index != -1) {
+            // ? can be escaped as ?? to support ? as a native SQL operator (e.g. on Postgres)
+            if (index < sql.length() - 1 && sql.charAt(index + 1) == '?') {
+                index = sql.indexOf('?', index + 2);
+                continue;
+            }
```
rfelcman, line 1988: It's about line behind 1996 with `anOperator.printsAs(v);`
please add there

```
// replace ?? with ? in code which is passed as a part of SQL into DB to avoid issue like
//Error Message = ORA-40597: JSON path expression syntax error ('$??(@.id == 1007)')
//JZN-00228: Question mark must be followed by parenthetical expression (line 1, position 3)
for (int i = 0; i < anOperator.getDatabaseStrings().length; i++) {
    anOperator.getDatabaseStrings()[i] = anOperator.getDatabaseStrings()[i].replace("??", "?");
}
```

Why? Oracle DB doesn't accept `??` sequence in SQL (Json) queries.
`anOperator.printsAs(v);` convert Expression into Strings which are passed/printed into DB SQL query.
Please let me know if this change suits with Postgres.

pvarga88: I see. Makes sense for Oracle, but this doesn't work for the Postgres JDBC driver, which expects '??' in the prepared statement in order to not confuse the '?' [in the statement with positional parameters.](https://jdbc.postgresql.org/documentation/query/#using-the-statement-or-preparedstatement-interface) Somehow the Oracle JDBC drives knows the difference. Do we need a change in some writer elsewhere that again replaces '?' with '??' for Postgres only? I don't know the code enough to know where to do that.

pvarga88: I gave it an attempt. See below.

rfelcman (approved):
LGTM

pvarga88:
@rfelcman I would like to backport this to 4.0. Is there a policy or process for that or do I just submit a separate PR to the 4.0 branch?

rfelcman:
> @rfelcman I would like to backport this to 4.0. Is there a policy or process for that or do I just submit a separate PR to the 4.0 branch?

Just create new PR against 4.0 branch.

pvarga88:
Thank you. 
Created #2156

PR #1948, opened 2023-09-22
Fix slightly inaccurate QUERY_TIMEOUT hint's javadoc

anestoruk:
Javadoc states that `seconds` is the default unit of time, but the actual default value seem to be `milliseconds`:

https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/descriptors/DescriptorQueryManager.java#L135


rfelcman (approved):
LGTM

PR #1953, opened 2023-10-02
avoid unnecessary reflection calls,

lukasj:
improve codegen to handle Map related mappings


rfelcman (approved):
LGTM (just solve test failures)

PR #1954, opened 2023-10-03
[2.7] Eclipselink.jar - JPMS fixes

rfelcman:
This change contains two fixes:
- opens some packages in module-info.java as there are some resource files needed to read
- eclipselink source bundle to include there correct module-info.java file


lukasj (approved):
LGTM

PR #1955, opened 2023-10-03
[master] JPA Test Advanced fix

rfelcman:
It seems, that after transformation from `jpa.test` to `jpa.testapps` `EntityManagerJUnitTestSuite` from `jpa.testapps.advanced` was excluded from the execution. This change returns this suite back.


lukasj (approved):
LGTM

PR #1956, opened 2023-10-04
[2.7] fix #1950: Duplicate objects in UnitOfWorkImpl.primaryKeyToNewObjects - backport from master

### LINKED ISSUE #1950: Duplicate objects in UnitOfWorkImpl.primaryKeyToNewObjects
When `UnitOfWorkImpl#registerNotRegisteredNewObjectForPersist` is called, the object is potentially added twice to the primaryKeyToNewObjects map.
First when the `assignSequenceNumber` method is called (assuming a sequence is used) and a second time when calling `registerNewObjectClone` (which calls `addNewObjectToPrimaryKeyToNewObjects`).

When the object is later unregistered, it is only removed once from the map (see `removeObjectFromPrimaryKeyToNewObjects`).

Subsequent calls on `EntityManger.find` will still find the remaining object, which should not be the case.

The following code snippet illustrates the behaviour:

```java
  @Test
  public void test() {
    Entity entity = new Entity();
    this.entityManager.persist(entity);
    
    Entity savedEntity = this.entityManager.find(Entity.class, entity.getId());
    this.entityManager.remove(savedEntity);
    
    savedEntity = this.entityManager.find(Entity.class, entity.getId());
    assertThat(savedEntity, nullValue());
  }
```

The assertion fails, because the object is still in UnitOfWorkImpl.primaryKeyToNewObjects as it was added twice but only removed once.

Version affected: 2.7.13
Last version with expected behaviour: 2.7.12
Change introduced with https://github.com/eclipse-ee4j/eclipselink/pull/1843

Zuplyx:
- fixes #1950 

Backport of #1951 to 2.7. For details see PR to master.

### DIFF
--- a/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/unitofwork/UOWPrimaryKeyToNewObjectsDuplicateObjectsTest.java
+++ b/foundation/eclipselink.core.test/src/org/eclipse/persistence/testing/tests/unitofwork/UOWPrimaryKeyToNewObjectsDuplicateObjectsTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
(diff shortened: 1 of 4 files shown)


rfelcman:
Just some note about failing `eclipsefdn/eca` check. It seems that e-mail address in the git **Author** doesn't match with sign-off footer.
```
Author: Patrick Schmitt <patrick_98@t*****.de>
Date:   Tue Oct 3 09:40:29 2023 +0200

......
......    
    Signed-off-by: Patrick Schmitt <Patrick.Schmitt@c****-s*****.com>
```

rfelcman (approved):
LGTM

PR #1958, opened 2023-10-05
[master] fix #1957: Presize HashMap allocations in UnitOfWorkImpl

### LINKED ISSUE #1957: Presize HashMap allocations in UnitOfWorkImpl
I've noticed, that there are several places in `UnitOfWorkImpl` where HashMaps are created without specifying an initial size even though their expected size is known or can be reliably estimated. This can lead to unnecessary resizing and rehashing operations, which will hurt performance, especially for large UnitOfWork.

I will provide PRs to fix this for master and 2.7

Zuplyx:
- fixes #1957 

Add initial size to map creation where expected size is known or can be estimated


rfelcman (approved):
LGTM

PR #1959, opened 2023-10-06
[2.7] fix 1957: Presize HashMap allocations in UnitOfWorkImpl - backport from master

### LINKED ISSUE #1957: Presize HashMap allocations in UnitOfWorkImpl

Zuplyx:
- fixes #1957 

backport of #1958


Zuplyx:
@rfelcman I am unsure how to deal with the test failure.
I cannot reproduce it locally and the jenkins log does not provide me with enough information about the cause.

Issue #968, opened 2020-11-27
Poor performance of InformixPlatform with IDENTITY generated primary keys


sgflt:
Method InformixPlatform#buildSelectQueryForIdentity creates following select:

`SELECT DISTINCT(DBINFO('sqlca.sqlerrd1')) FROM systables`

1) The 'sqlca. sqlerrd1' option returns a single integer that provides the last serial value that is inserted into a table. 
2) DISTINCT does not makes sense on sigle value
3) multiple values are produced by systables which may have hundreds of rows

Proposed fix:

`SELECT FIRST 1 DBINFO('sqlca.sqlerrd1') FROM systables`

Fragnite-hub:
Can I work on this !!

PR #1971, opened 2023-10-14
 adding FIRST 1 DBINFO #968

Fragnite-hub:
Performane Improved by removing SELECT DISTINCT in the query WHICH IS NOT REQUIRED  and adding First !DBINFO.


lukasj (changes requested):
This PR just removes code and adds nothing, so it is unclear what the intent is

rfelcman:
According #1991 this one is pointless. It's going to be closed.

PR #1976, opened 2023-10-20
New IndirectSequencedCollection introduced in Java 21 by java.util.SequencedCollection

rfelcman:
plus unit test and implementation in org.eclipse.persistence.indirection.IndirectList


lukasj (commented):
`IndirectSequencedCollection` should not be needed

PR #1979, opened 2023-10-24
[master] Set JDK 21 as a compiler release switch

rfelcman:
It leads into following minimal changes to pass a build:

- refactor call `new Locale(...)` into `Locale.of(...)`
- resolve `addFirst(...), addLast(...)` collision in `org.eclipse.persistence.internal.helper.linkedlist.ExposedNodeLinkedList` with `java.util.List` [this is incompatible change, but limited to \"internal\" classes
- disable `jpa.test.metamodel.apectj` Maven module as AspectJ 1.9.20.1 is not JDK 21 ready
	waiting for a new relase (see https://github.com/eclipse-aspectj/aspectj/pull/261)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -45,7 +45,7 @@ jobs:
       uses: actions/setup-java@v3
       with:
         distribution: 'zulu'
-        java-version: 17
+        java-version: 21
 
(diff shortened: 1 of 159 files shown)


lukasj (commented):
also bump version to 5.0.0-SNAPSHOT

PR #1996, opened 2023-11-10
Jakarta Persistence 3.2:

Tomas-Kraus:
jakartaee/persistence#457 - add union/intersect/except to JPQL and criteria
Jakarta Persistence dependency updated to 3.2.0-B02


jpa/eclipselink.jpa.testapps/jpa.test.persistence32/pom.xml
```diff
@@ -80,6 +80,34 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Static metamodel is used in tests -->
+            <plugin>
+                <groupId>org.bsc.maven</groupId>
+                <artifactId>maven-processor-plugin</artifactId>
+                <version>4.5</version>
```
lukasj, line 87: why? see ie [inheritance tests](https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.testapps/jpa.test.inheritance/pom.xml#L37-L49) how to use what we already have

Tomas-Kraus: OK, fixed.

lukasj (approved):

PR #2006, opened 2023-11-22
[master] DatabasePlatform.wasFailureCommunicationBased() fix

### LINKED ISSUE #1970: DatabasePlatform.wasFailureCommunicationBased() seems broken when no pingSQL is defined.
When a query execution fails, the system is supposed (with regular settings) to check whether this was due to a communication error, and if this is the case, retry the query with a new connection.
This works fine in most cases, but I stumbled upon a case where it seems to always fail to detect such a communication issue.

If the `pingSQL` property of the `DatabasePlatform` is not set, which happens for instance when using an non-builtin platform without specifying the platform class (e.g., a MariaDB target with no `eclipselink.target-database` defined in `persistence.xml`), then communication failures are never properly detected, and the system throws a `DatabaseException` without retrying anything.

I think the problem lies in the `DatabasePlatform.wasFailureCommunicationBased(SQLException exception, Connection connection, AbstractSession sessionForProfile)` method. The code reads as follows:

```java
(...)
} else if (this.pingSQL == null) {
(...)
    try {
        return connection.isValid(IS_VALID_TIMEOUT);
    } catch (Throwable failed) {
    (...)
```

I'm not super familiar with the `Connection` API, so I'm not sure about the exact semantics of the `isValid()` method, but, shouldn't the return value be the opposite here? That is, if the connection is not valid, then shouldn't `wasFailureCommunicationBased` be `true`, instead of `false`?

It's probably not a big deal as most platforms do provide a default `pingSQL` and are thus unaffected by the issue, but I thought I'd still report this and have someone more familiar with this code double check this.

I've seen this with version 2.7.13, but the code doesn't seem to have changed in 4.0, so I guess this affects both branches (I haven't checked the others though).

Cheers!

---

Reproducing this is fairly straightforward:

- Set up a connection to a MariaDB database without specifying the `eclipselink.target-database`(so that the generic `DatabasePlatform` class is used, and not the `MySQLPlatform`, which does have a `pingSQL` statement set by default).
- Create an `EntityManagerFactory` and run some queries on the DB to check everything is fine.
- Pause.
- Kill the database connection (by restarting the server for instance).
- Try to run some queries again: an exception is thrown to the client code.

A breakpoint inside `DatabasePlatform.wasFailureCommunicationBased()` may help seeing what's going on.

The exact same steps but with `eclipselink.target-database` set to `org.eclipse.persistence.platform.database.MySQLPlatform` for instance will trigger a retry and get the result to the client as expected.

rfelcman:
Fixes #1970 .
This bug should happens is no `pingSQL` DB platform property is specified and query execution fails.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -3224,7 +3224,7 @@ public class DatabasePlatform extends DatasourcePlatform {
              // By default use the JDBC isValid API unless a ping SQL has been set.
              // The ping SQL is set by most platforms, but user could set to null to used optimized JDBC check if desired.
              try {
-                 return connection.isValid(IS_VALID_TIMEOUT);
+                 return !connection.isValid(IS_VALID_TIMEOUT);
              } catch (Throwable failed) {
                  // Catch throwable as old JDBC drivers may not support isValid.
                  return false;


lukasj (approved):

TomerHeber:
Hi,

Is there a reason this fix isn't backported to 3.x and 4.x ? (and just to 2.x).

Thanks!

rfelcman:
> Hi,
> 
> Is there a reason this fix isn't backported to 3.x and 4.x ? (and just to 2.x).
> 
> Thanks!

In progress now see #2228 #2229 .

PR #2010, opened 2023-11-28
[2.7] Close #1806 - Have to the ability to dynamically set the Validation Group during a transaction

### LINKED ISSUE #1806: Proposal: Have to the ability to dynamically set the Validation Group for pre-persist during transaction
One weakness of the JPA Specification is that you cannot (to my knowledge) specify a validation group to evaluate during a persist call.

Example: This does not work:

```
em.setProperty(PersistenceUnitProperties.VALIDATION_GROUP_PRE_PERSIST, SomeValidationGroup.class.toString());
em.persist(entity);
```

Looking at `BeanValidationListener.java` however, this seems like it'd be trivial to implement.

Existing code:
```
    @Override
    public void prePersist (DescriptorEvent event) {
        validateOnCallbackEvent(event, "prePersist", groupPrePersit);
    }
```

Proposal:
```
    @Override
    public void prePersist (DescriptorEvent event) {
        Class<?>[] overrideGroup = event.getSession().getParent().getProperties.get("eclipselink.validation.override.pre-persist");
        if (overrideGroup != null){
            validateOnCallbackEvent(event, "prePersist", overrideGroup);
        } else {
            validateOnCallbackEvent(event, "prePersist", groupPrePersit);
        }
    }
```

If we put together a patch, does anyone have any concerns about this? What we would need to submit for tests, docs, etc? We're still using 2.7.x series, but we'd try to patch all of the active development versions (what are they?)

Thank you!

exabrial:
Hello, this PR Close #1806 for the 2.7 branch. The goal here is to allow a user to specify a validation group that should be executed when persisting. I'm hoping to sort of model this how the schema-per-tenant properties work (Which is very handy btw).

I have a signed ECA, but this is my first PR for EclipseLink so I could use some feedback. If this gets merged, then I'll port the changes to the other branches seeing active development.

Thank you!

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
@@ -323,17 +323,17 @@ public class EntityManagerProperties {
     /**
      * Overrides the Bean Validation Group(s) that will execute during a prePersist event. This should be a class or class[].
      */
-    public static final String VALIDATION_GROUP_PRE_PERSIST = "eclipselink.validation.group.prePersist";
+    public static final String VALIDATION_GROUP_PRE_UPDATET = "eclipselink.beanvalidation.group.pre-persist";
     
     /**
      * Overrides the Bean Validation Group(s) that will execute during a preUpdate event. This should be a class or class[].
      */
-    public static final String VALIDATION_GROUP_PRE_UPDATE = "eclipselink.validation.group.preUpdate";
+    public static final String VALIDATION_GROUP_PRE_UPDATE = "eclipselink.beanvalidation.group.pre-update";
     
     /**
      * Overrides the Bean Validation Group(s) that will execute during a preRemove event. This should be a class or class[].
      */
-    public static final String VALIDATION_GROUP_PRE_REMOVE = "eclipselink.validation.group.preRemove";
+    public static final String VALIDATION_GROUP_PRE_REMOVE = "eclipselink.beanvalidation.group.pre-remove";
 
     private static final Set<String> supportedProperties = new HashSet<String>() {
 
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2025 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -319,6 +319,21 @@ public class EntityManagerProperties {
      * )
      */
     public static final String COMPOSITE_UNIT_PROPERTIES = PersistenceUnitProperties.COMPOSITE_UNIT_PROPERTIES;
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a prePersist event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_PERSIST = "eclipselink.validation.group.prePersist";
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a preUpdate event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_UPDATE = "eclipselink.validation.group.preUpdate";
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a preRemove event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_REMOVE = "eclipselink.validation.group.preRemove";
 
     private static final Set<String> supportedProperties = new HashSet<String>() {
 
@@ -344,6 +359,9 @@ public class EntityManagerProperties {
             add(PERSISTENCE_CONTEXT_COMMIT_ORDER);
             add(FLUSH_CLEAR_CACHE);
             add(COMPOSITE_UNIT_PROPERTIES);
+            add(VALIDATION_GROUP_PRE_PERSIST);
+            add(VALIDATION_GROUP_PRE_UPDATE);
+            add(VALIDATION_GROUP_PRE_REMOVE);
         }
     };
 
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
@@ -323,7 +323,7 @@ public class EntityManagerProperties {
     /**
      * Overrides the Bean Validation Group(s) that will execute during a prePersist event. This should be a class or class[].
      */
-    public static final String VALIDATION_GROUP_PRE_UPDATET = "eclipselink.beanvalidation.group.pre-persist";
+    public static final String VALIDATION_GROUP_PRE_PERSIST = "eclipselink.beanvalidation.group.pre-persist";
     
     /**
      * Overrides the Bean Validation Group(s) that will execute during a preUpdate event. This should be a class or class[].
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/GermanPhone.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/GermanPhone.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2009, 2025 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+
+package org.eclipse.persistence.testing.models.jpa.beanvalidation;
+
+public interface GermanPhone {
+
+}
--- a/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/USPhone.java
+++ b/jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/USPhone.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2009, 2025 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+
+package org.eclipse.persistence.testing.models.jpa.beanvalidation;
+
+public interface USPhone {
+
+}
--- a/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
+++ b/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2025 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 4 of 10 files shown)


rfelcman (changes requested):
1. For functional changes like this we expect some test (in this case integration test). Please check and maybe extend `jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java`
2. Please ensure, that copyright year in the modified files is updated.

exabrial:
Thank you for the direction! I really appreciate it. We'll get to work on these things.

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
```diff
@@ -319,6 +319,21 @@ public class EntityManagerProperties {
      * )
      */
     public static final String COMPOSITE_UNIT_PROPERTIES = PersistenceUnitProperties.COMPOSITE_UNIT_PROPERTIES;
+
```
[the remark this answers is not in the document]
exabrial: I intended them to be Entity Manager Properties, since we could allow them to be changed in a transaction. I believe the PersistenceUnit already has support for this feature, but it can't be changed dynamically, hence the reason for this enhancement. I hope that makes sense! Please let me know if it doesn't or if my reasoning is not correct.

exabrial:
This pul isn't abandoned, just still figuring out how to write tests :)

exabrial:
still haven't abandoned this, I'll get to it.

rfelcman:
Please check and sign [eclipsefdn/eca](https://api.eclipse.org/git/eca/status/gh/eclipse-ee4j/eclipselink/2010)

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/GermanPhone.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.persistence.testing.models.jpa.beanvalidation;
```
exabrial, line 1: missing copyright

jpa/eclipselink.jpa.test/src/org/eclipse/persistence/testing/models/jpa/beanvalidation/USPhone.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.persistence.testing.models.jpa.beanvalidation;
```
exabrial, line 1: missing copyright

exabrial:
> Please check and sign [eclipsefdn/eca](https://api.eclipse.org/git/eca/status/gh/eclipse-ee4j/eclipselink/2010)

It's all good now! there was an errant signature

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
```diff
@@ -319,6 +319,21 @@ public class EntityManagerProperties {
...
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a preUpdate event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_UPDATE = "eclipselink.validation.group.preUpdate";
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a preRemove event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_REMOVE = "eclipselink.validation.group.preRemove";
```
rfelcman, line 336: I believe, that there should be used same naming and reference to `PersistenceUnitProperties` 
e.g.
`public static final String VALIDATION_GROUP_PRE_PERSIST = PersistenceUnitProperties.VALIDATION_GROUP_PRE_PERSIST;`
to keep property names consistent like other properties in this class.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
```diff
@@ -81,19 +84,38 @@ public void prePersist (DescriptorEvent event) {
         //    tx.commit();
         //  "invalid data" would get inserted into database.
         //
-        //  preInsert can be used to work around above issue. Howerver, the JPA spec does not itent it.
+        //  preInsert can be used to work around above issue. However, the JPA spec does not itent it.
         //  This might be corrected in next iteration of spec
-        validateOnCallbackEvent(event, "prePersist", groupPrePersit);
+        Object overrideGroups = event.getSession().getParent().getProperties().get(EntityManagerProperties.VALIDATION_GROUP_PRE_PERSIST);
```
rfelcman, line 89: I don't think, that this code should be there.
For EntityManager properties is location
https://github.com/eclipse-ee4j/eclipselink/blob/2.7/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java#L4009
https://github.com/eclipse-ee4j/eclipselink/blob/2.7/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/EntityManagerSetupImpl.java#L4038
or
https://github.com/eclipse-ee4j/eclipselink/blob/2.7/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/deployment/BeanValidationInitializationHelper.java#L45

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
```diff
@@ -81,19 +84,38 @@ public void prePersist (DescriptorEvent event) {
...
-    @Override
     public void aboutToUpdate(DescriptorEvent event) {
         Object source = event.getSource();
         UnitOfWorkImpl unitOfWork = (UnitOfWorkImpl )event.getSession();
         // preUpdate is also generated for deleted objects that were modified in this UOW.
         // Do not perform preUpdate validation for such objects as preRemove would have already been called.
-        if(!unitOfWork.isObjectDeleted(source)) {
-            validateOnCallbackEvent(event, "preUpdate", groupPreUpdate);
+        if (!unitOfWork.isObjectDeleted(source)) {
+          Object overrideGroups = event.getSession().getParent().getProperties().get(EntityManagerProperties.VALIDATION_GROUP_PRE_UPDATE);
```
rfelcman, line 108: Same as above

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
```diff
@@ -104,9 +126,17 @@ public void preUpdateWithChanges(DescriptorEvent event) {
 
     @Override
     public void preRemove (DescriptorEvent event) {
-        if(groupPreRemove != null) { //No validation performed on preRemove if user has not explicitly specified a validation group
-           validateOnCallbackEvent(event, "preRemove", groupPreRemove);
-        }
+      Object overrideGroups = event.getSession().getParent().getProperties().get(EntityManagerProperties.VALIDATION_GROUP_PRE_REMOVE);
```
rfelcman, line 129: Same as above

rfelcman (changes requested):

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
exabrial: @rfelcman thank you for the feedback! I definitely understand the importance of consistency, especially because once it's released, it's nearly impossible to change!

To that goal, I checked `PersistenceUnitProperties` and found the exact same property names:
* `PersistenceUnitProperties.VALIDATION_GROUP_PRE_PERSIST`
* `PersistenceUnitProperties.VALIDATION_GROUP_PRE_UPDATE`
* `PersistenceUnitProperties.VALIDATION_GROUP_PRE_REMOVE`

So I believe the property names are correct... Did I understand your feedback correctly?

exabrial: OH I see what you're saying, it's the contents of the string, not the actual constant name. Yes. I will fix that. Great catch!

exabrial: OK! So JPA 2.1 contains:

```
public static final String VALIDATION_GROUP_PRE_PERSIST = "javax.persistence.validation.group.pre-persist";
```

There is no official specification that I can find in JPA 2.0-3.1 that I can find that allows overriding it at the EntityManager level, which is why I introduced the property with the `eclipselink.x` prefix rather than 
`jakarta.persistence.validation.x`

This is so in the future if/when this is added to the spec, the EclipseLink property can be aliased to the official one, much like how eclipselink's extended caching options are prefixed with `eclipselink` to avoid namespace clashing.


Looking at EntityManagerProperties however, I See that I should probably be using `eclipselink.beanvalidation.x` as the prefix. I will make that update and push. 

Also, please take a peek at my question about testing, thank you so much!

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
exabrial: Apologies, I don't think I understand what you're saying. 

Perhaps this is the confusion: my patch is designed to allow the user to override these settings during execution. Otherwise the validation groups are fixed in stone and cannot be modified.

rfelcman: I mean, that locations which are mentioned in the links are correct where new EntityManager properties should be evaluated like other properties. Not there. This is not consistent with other EntityManager properties or PersistenceUnit bean validation properties.

rfelcman:
> @rfelcman If I could have a bit of feedback, I added several tests, but I'm not convinced they're actually running, because I never see them in the junit test report html. Whats the ant command to run an individual test class? I've tried several things:
> 
>     * `~/opensource/eclipselink$ ant -f antbuild.xml test-jpa test-core`
> 
>     * `~/opensource/eclipselink$ ant -f antbuild.xml test-jpa -Dtestcase=BeanValidationJunitTest`
> 
> 
> Any help appreciated!

ant -f antbuild.xml test-jpa22 is better

Individual test class in 2.7 (Ant build) is complicated. It's easier start with other higher branches (3.0 and above) which are based on Maven and make backport.

jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/metadata/listeners/BeanValidationListener.java
exabrial: Thank you again for working with me. I read all three of your examples. I'm not trying to say my code is perfect (totally aware of my absolute average skills) however, I don't think I can do use them for this purpose. 

Perhaps a use case would be beneficial? Here is the scenario. We have a form where the user provides 'tidbits' of information  as they progress through the form. We want to save those off to db as we go. However, eventually all of the fields need to be required.

```
import javax.enterprise.context.ApplicationScoped;
import javax.persistence.Column;
import javax.persistence.Entity;
import javax.persistence.EntityManager;
import javax.persistence.PersistenceContext;
import javax.transaction.Transactional;
import javax.transaction.Transactional.TxType;
import javax.validation.constraints.NotNull;
import javax.validation.groups.Default;

import org.eclipse.persistence.config.EntityManagerProperties;

import lombok.Data;

@ApplicationScoped
@Transactional(TxType.MANDATORY)
public class YeeHahDatabaseService {
	@PersistenceContext(unitName = "superbiz-model")
	private EntityManager em;

	public void addContact(final String contactName) {
		final SuperBizContact contact = new SuperBizContact();
		contact.setName(contactName);
		em.persist(contact); // Would throw an error because phone is marked notNull! But I don't have phone yet!
	}

	public void addContactWithPhone(final String contactName, final String phone) {
		final SuperBizContact contact = new SuperBizContact();
		contact.setName(contactName);
		contact.setPhone(phone);
		em.persist(contact); // All good here, both fields set
	}

	public void addContatc2(final String contactName) {
		final SuperBizContact contact = new SuperBizContact();
		contact.setName(contactName);
		em.setProperty(EntityManagerProperties.VALIDATION_GROUP_PRE_PERSIST, NameOnly.class);
		em.persist(contact); // This time the persist works!
	}

	@Data
	@Entity
	public class SuperBizContact {
		@Column
		@NotNull(groups = { Default.class, NameOnly.class })
		private String name;
		@Column
		@NotNull
		private String phone;
	}

	public @interface NameOnly {
	}
}
```


Example 1/2:  `EntityManagerSetupImpl.java` as far as I can tell, this code has already fired by the time the dev sees the injection.

Example 3: `BeanValidationInitializationHelper` same here. This code has fired already when the dev see the injection. 


The purpose of this property is to allow the dev to set the EntityManager for the current `@PersistenceContext` rather than reading it from an unmodifiable configuration XML (which is actually already supported).

foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/config/EntityManagerProperties.java
```diff
@@ -319,6 +319,21 @@ public class EntityManagerProperties {
      * )
      */
     public static final String COMPOSITE_UNIT_PROPERTIES = PersistenceUnitProperties.COMPOSITE_UNIT_PROPERTIES;
+    
+    /**
+     * Overrides the Bean Validation Group(s) that will execute during a prePersist event. This should be a class or class[].
+     */
+    public static final String VALIDATION_GROUP_PRE_UPDATET = "eclipselink.beanvalidation.group.pre-persist";
```
exabrial: probably a typo here

exabrial:
I figured out testing. Take a look! Thank you so much

exabrial:
Hello admins! I think this is possibly ready to go. Thank you!

arjantijms:
The github tests won't run really:

> BUILD FAILED
/home/runner/work/eclipselink/eclipselink/antbuild.xml:460: Maven 3 not found. Build won't run if Maven 3 not installed and M2_HOME properly set!.

@exabrial can you take a look at why they fail?

arjantijms:
@lukasj any opinion on this one?

exabrial:
> The github tests won't run really:
> 
> > BUILD FAILED
> > /home/runner/work/eclipselink/eclipselink/antbuild.xml:460: Maven 3 not found. Build won't run if Maven 3 not installed and M2_HOME properly set!.
> 
> @exabrial can you take a look at why they fail?

I would love to, but I can't log into the CI server.  Any chance I could access?

Issue #2031, opened 2024-01-08
Null value support for UUID


adamsaghy:
**Describe the bug**
When the attribute value of a UUID data type is null, the built-in UUIDConverter is called, but it fails with IllegalArgumentException hence the null is not a valid UUID type.

Also if the UUID attribute has `null` value, the `org.eclipse.persistence.internal.databaseaccess.DatabasePlatform#getJDBCType` method is fall back to VARCHAR type, which is wrong. If it needs to be fallback to something then it should be `OTHER`, so the Postgres db engine would not fail when we are sending varchar `null` to a `uuid` type column. 

**To Reproduce**
Steps/resources to reproduce the behaviour:

- EclipseLink version: 4.0.2
- Code example which leads into bug like:

```

public class TestEntityMaster {
...
@Column("uuid")
private UUID uuid;
...
}
...
EntityManagerFactory entityManagerFactory = Persistence.createEntityManagerFactory("test-jpa-pu");
EntityManager em = entityManagerFactory.createEntityManager();
em.getTransaction().begin();
TestEntityMaster entity = new TestEntityMaster(null);
em.persist(entity);
em.getTransaction().commit();
...
```

**Expected behaviour**

Write `null` into the database.
Read `null` from the database.

adamsaghy:
PR for this: https://github.com/eclipse-ee4j/eclipselink/pull/2032

adamsaghy:
Kindly asking you guys to validate the above change request!

I am interested in whether changing the fallback type from `VARCHAR` to `OTHER` is a valid change (and a good idea).

PR #2032, opened 2024-01-08
Support null value for UUID (#2031)

adamsaghy:
- Add `null` value handling to UUIDConverter
- Change `DatabasePlatform.java#getJDBCType` fallback type from `VARCHAR` to `OTHER` -> Postgres was failing when a `null` value was tried to stored to a UUID type column hence EclipseLink was sent varchar type null.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -1102,7 +1102,7 @@ public class DatabasePlatform extends DatasourcePlatform {
      */
     public int getJDBCType(Class<?> javaType) {
         if (javaType == null) {
-            return Types.VARCHAR;// Best guess, sometimes we cannot determine type from mapping, this may fail on some drivers, other dont care what type it is.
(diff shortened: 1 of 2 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -1102,7 +1102,7 @@ public int getJDBCType(DatabaseField field) {
      */
     public int getJDBCType(Class<?> javaType) {
         if (javaType == null) {
```
rfelcman, line 1104: By bug description it looks like PostgreSQL specific issue. Other DBs like MySQL, Oracle allows store UUID type into VARCHAR columns. I don't think, that this is right place for this change. I'd like recommend `org.eclipse.persistence.platform.database.PostgreSQLPlatform#getJDBCType`.

rfelcman (changes requested):
In general what I missing there:
- Please update year in the copyright header e.g. ` * Copyright (c) 1998, 2024....`
- Missing some unit/integration test. Easiest way is extend https://github.com/eclipse-ee4j/eclipselink/pull/1418/files#diff-e9f994d5b63522f6e937db6d5d0a7cb795fbaa593ea594c8d01f606c4ef343dc

Issue #1577, opened 2022-07-08
StackOverflowError because of cyclic DETACH


mheydem:
After upgrading from
- Java SE 8
- Glassfish 5.1
- Jakarta EE 8
- EclipseLink 2.7.4

to
- Java SE 8
- Glassfish 6.0
- Jakarta EE 9.0
- EclipseLink 3.0.0

the action
- detaching an entity which has a bidirectional oneToMany relation

causes
- cyclic detaching ending up in an StackOverflowError
- DescriptorIterator.iterateIndirectContainerForMapping loops until StackOverflowError

Here an example:
```
public class Group {

	@Column(name = "NAME")
	@NotNull
	private String name;

	@OneToMany(
		cascade = {CascadeType.ALL },
		fetch = FetchType.LAZY,
		mappedBy = "group",
		targetEntity = GroupItem.class)
	@MapKey(name = "name")
	@NotNull
	private Map<String, GroupItem<?>> properties = new HashMap<>();
}
```

```
public class GroupItem {

	@ManyToOne(
		cascade = { CascadeType.ALL },
		fetch = FetchType.EAGER,
		optional = false)
	@JoinColumn(name = "GROUP_ID")
 	@NotNull
 	private Group group;
}
```
The code above is working fine with EclipseLink 2.x, but when running on EclipseLink 3.0.2  and detaching a `Group` is tried a StackOverflowError can be experienced.

Removing the `CascadeType.Detach` from one side of the relation avoids the StackOverflowError, but is not a real workaround as it will not behave exactly the same way:

```
public class GroupItem {
	@ManyToOne(
			cascade = {
					CascadeType.PERSIST,
					CascadeType.MERGE,
					CascadeType.REMOVE,
					CascadeType.REFRESH
					/*CascadeType.DETACH*/ },
			fetch = FetchType.EAGER,
			optional = false)
	@JoinColumn(name = "GROUP_ID")
	@NotNull
	private Group group;
```

Here the stack trace:
```
Caused by: java.lang.StackOverflowError
	at sun.nio.ch.IOUtil.write(IOUtil.java:58)
	at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:470)
	at oracle.net.nt.TimeoutSocketChannel.write(TimeoutSocketChannel.java:215)
	at oracle.net.ns.NIOPacket.writeToSocketChannel(NIOPacket.java:211)
[...]

PR #2037, opened 2024-01-17
[master] detach() fixes

### LINKED ISSUE #2017: Lazy collections are fetched on detach
I've recently migrated from javax to jakarta, from Eclipselink 2.x to 4.0.2.
Something that didn't happen before, but is happening with 4.0.2 is that when calling `em.detach(entity)`, if the entity has an uninitialized lazy collection, it gets fetched in the detach operation.
This is a high impact bug, as it could potentially load fetch a lot of db rows to memory.
In the project I work, we have many mapped `@OneToMany` collections which are private, and are only there to cascade deletes.
And we do a lot of detach to prevent many objects from end up in memory.
With this bug, the result is the exact opposite...

I have created a minimal project to reproduce the issue: https://github.com/luisfpg/eclipselink-detach-bug

I'm copying here parts of the README to make it easier to screen the problem:

This is the code to reproduce the issue:

```
var state = em.find(State.class, "s1");
var cities = (IndirectSet<City>) state.getCities();
System.out.printf("Is collection initialized? %s%n", cities.isInstantiated());
System.out.println("Before detach");
em.detach(state);
System.out.printf("State %s detached%n", state.getCode());
System.out.printf("Is collection initialized? %s%n", cities.isInstantiated());
```

And the output is:

```
[EL Info]: 2023-12-06 08:02:59.36--ServerSession(232307208)--EclipseLink, version: Eclipse Persistence Services - 4.0.2.v202306161219
Loaded state s1
Is collection initialized? false
Before detach
Loaded state s1
Loaded city s1b
Loaded city s1a
State s1 detached
Is collection initialized? true
```

Note that the collection gets initialized by the detach operation, which makes no sense.

rfelcman:
There are two fixes for EntityManager.detach():
- Lazy collections are fetched on detach
  Unnecessary fetch of lazy loaded collection `Indirect...` declared in the detached entity. See #2017
- StackOverflowError because of cyclic DETACH
  It happens if there is cross relation between entities like 
  `@OneToMany(cascade = {CascadeType.ALL }, fetch = FetchType.LAZY,...` from the first entity and
  `@ManyToOne(cascade = { CascadeType.ALL}, fetch = FetchType.EAGER,...` from the second. See #1577
fixes #2017 #1577 
replaces #292

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
@@ -161,7 +161,7 @@ public abstract class DescriptorIterator {
         return visitedObjects;
     }
 
-    public Map getVisitedMappings() {
+    public Map<Integer, DatabaseMapping> getVisitedMappings() {
         return visitedMappings;
     }
 
@@ -712,7 +712,7 @@ public abstract class DescriptorIterator {
         }
 
         public boolean shouldNotCascade(DatabaseMapping mapping){
-            return !(shouldCascadeAllParts() || (shouldCascadePrivateParts() && mapping.isPrivateOwned()));
+            return !shouldCascade(mapping);
         }
     }
 
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
@@ -54,7 +54,7 @@ public abstract class DescriptorIterator {
     public static final int CascadePrivateParts = 2;
     public static final int CascadeAllParts = 3;
     protected Map visitedObjects;
-    protected Map visitedMappings;
+    protected Map<Integer, DatabaseMapping> visitedMappings;
     protected Stack visitedStack;
     protected AbstractSession session;
     protected DatabaseMapping currentMapping;
@@ -94,7 +94,7 @@ public abstract class DescriptorIterator {
     protected DescriptorIterator() {
         // 2612538 - the default size of Map (32) is appropriate
         this.visitedObjects = new IdentityHashMap();
-        this.visitedMappings = new HashMap();
+        this.visitedMappings = new HashMap<>();
         this.visitedStack = new Stack();
         this.cascadeDepth = CascadeAllParts;
         this.shouldIterateOverIndirectionObjects = true;// process the objects contained by ValueHolders...
@@ -455,7 +455,7 @@ public abstract class DescriptorIterator {
             this.currentItem = currentItemOriginal;
         } else {
             for (DatabaseMapping mapping : mappings) {
-                if (!this.cascadeCondition.shouldNotCascade(mapping)) {
+                if (this.cascadeCondition.shouldCascade(mapping)) {
                     mapping.iterate(this);
                 }
             }
@@ -705,6 +705,10 @@ public abstract class DescriptorIterator {
          * Default constructor.
          */
         public CascadeCondition() {
+        }
+
+        public boolean shouldCascade(DatabaseMapping mapping){
+            return !shouldNotCascade(mapping);
         }
 
         public boolean shouldNotCascade(DatabaseMapping mapping){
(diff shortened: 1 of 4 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
```diff
@@ -92,6 +94,7 @@ public abstract class DescriptorIterator {
     protected DescriptorIterator() {
         // 2612538 - the default size of Map (32) is appropriate
         this.visitedObjects = new IdentityHashMap();
+        this.visitedMappings = new HashMap();
```
lukasj, line 97: use generics

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
```diff
@@ -441,7 +455,9 @@ protected void internalIterateReferenceObjects(Object sourceObject) {
             this.currentItem = currentItemOriginal;
         } else {
             for (DatabaseMapping mapping : mappings) {
-                mapping.iterate(this);
+                if (!this.cascadeCondition.shouldNotCascade(mapping)) {
```
lukasj, line 458: I'd probably rather add `shouldCascade` to the cascadeCondition

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/descriptors/DescriptorIterator.java
```diff
@@ -707,6 +707,10 @@ public class CascadeCondition{
         public CascadeCondition() {
         }
 
+        public boolean shouldCascade(DatabaseMapping mapping){
+            return !shouldNotCascade(mapping);
```
lukasj, line 711: the whole point of my comment was to avoid negation of negated value..., the idea is for this to be:
```
return shouldCascadeAllParts() || (shouldCascadePrivateParts() && mapping.isPrivateOwned())
```

and `shouldNotCascade` to be:
```
return !shouldCascade(...);
```

so in the end one can avoid having a call stack evaluated to:
`(!(!(shouldCascadeAllParts() || (shouldCascadePrivateParts() && mapping.isPrivateOwned())))`

rfelcman: Uff fixed. Needed update `UnitOfWorkImpl` too.

lukasj (approved):

chicko12345:
Hello Community,

as i understand this fix wasn't back ported for 4.x. Am i right?

Kind Regards

rfelcman:
> Hello Community,
> 
> as i understand this fix wasn't back ported for 4.x. Am i right?
> 
> Kind Regards

See
https://github.com/eclipse-ee4j/eclipselink/pull/2225

chicko12345:
> > Hello Community,
> > as i understand this fix wasn't back ported for 4.x. Am i right?
> > Kind Regards
> 
> See #2225

Thank you very much!

rfelcman:
> > > Hello Community,
> > > as i understand this fix wasn't back ported for 4.x. Am i right?
> > > Kind Regards
> > 
> > 
> > See #2225
> 
> Thank you very much!

Hello,
it's available in the latest snapshots e.g. https://jakarta.oss.sonatype.org/content/groups/staging/org/eclipse/persistence/eclipselink/4.0.5-SNAPSHOT/ but it's not included in the latest release 4.0.4 https://github.com/eclipse-ee4j/eclipselink/releases/tag/4.0.4 . Will be available in 4.0.5 release (autumn 2024).

chicko12345:
> > > > Hello Community,
> > > > as i understand this fix wasn't back ported for 4.x. Am i right?
> > > > Kind Regards
> > > 
> > > 
> > > See #2225
> > 
> > 
> > Thank you very much!
> 
> Hello, it's available in the latest snapshots e.g. https://jakarta.oss.sonatype.org/content/groups/staging/org/eclipse/persistence/eclipselink/4.0.5-SNAPSHOT/ but it's not included in the latest release 4.0.4 https://github.com/eclipse-ee4j/eclipselink/releases/tag/4.0.4 . Will be available in 4.0.5 release (autumn 2024).

Tnx for the info. This one was a show stopper. So if it'll be available soon (autumn is absolutely ok) - very nice.

chicko12345:
Hi Community,

since the fix is now an official part of glassfish 7.0.22 we've decided to upgrade. For the small datasets it did the trick. However, during testing with bigger datasets (realistic cases in production environments) we run into the same stack overflow issue. Should i create a new issue?

rfelcman:
> Hi Community,
> 
> since the fix is now an official part of glassfish 7.0.22 we've decided to upgrade. For the small datasets it did the trick. However, during testing with bigger datasets (realistic cases in production environments) we run into the same stack overflow issue. Should i create a new issue?

Yes please, but we need what is mentioned behind "bigger datasets". The best could be some test case which leads into issue.
Or is it possible to simulate it by https://github.com/eclipse-ee4j/eclipselink/pull/2037/files#diff-016938172968e33273de6874cb7b991a500d4682b7bdf1c226651ab92b57a8a1
but with more DB rows/entities?
Or there is different entity model/tree?

PR #2045, opened 2024-01-24
Removed most Vector usages from ClassDescriptor and Project

lukasj:
* clean up in tests
* updated code to use pattern matching, text blocks
* removed no-warning compiler requirement from various places to allow building on different SEs without hassles


rfelcman (approved):
LGTM

PR #2047, opened 2024-01-25
[4.0] Startup Time Regression in EclipseLink 4.0.1

anija-anil:
4-6% Startup Time Regression in EclipseLink 4.0.1
for #2048

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023,2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -14,6 +14,7 @@
 //     Oracle - initial API and implementation
 package org.eclipse.persistence.asm;
 
+import org.eclipse.persistence.asm.internal.Util;
 import org.eclipse.persistence.config.SystemProperties;
 import org.eclipse.persistence.exceptions.ValidationException;
 import org.eclipse.persistence.internal.security.PrivilegedAccessHelper;
@@ -26,6 +27,7 @@ import java.lang.reflect.Field;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -33,16 +35,15 @@ import java.util.regex.Pattern;
 
 public class ASMFactory {
 
-    private final static SessionLog LOG = AbstractSessionLog.getLog();
-
+    // This block must be first - begin
     public final static String ASM_SERVICE_ECLIPSELINK = "eclipselink";
     public final static String ASM_SERVICE_OW2 = "ow2";
     private final static String ASM_OW2_CLASS_VISITOR = "org.objectweb.asm.ClassVisitor";
     private final static String ASM_ECLIPSELINK_CLASS_VISITOR = "org.eclipse.persistence.internal.libraries.asm.ClassVisitor";
+    private final static SessionLog LOG = AbstractSessionLog.getLog();
+    // This block must be first - end
 
-    //Should be changed in case of ASM upgrade
-    public final static int ASM_API_SELECTED = Opcodes.valueInt("ASM9");
-    public final static int JAVA_CLASS_VERSION = Opcodes.valueInt("V1_8");
+    // Should be changed in case of ASM upgrade
     public final static int JAVA_CLASS_LATEST_VERSION = ASMFactory.getLatestOPCodeVersion();
 
     public static AnnotationVisitor createAnnotationVisitor(final int api) {
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/Opcodes.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/Opcodes.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023,2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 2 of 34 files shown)


rfelcman:
Just to be curious. Do You have some performance metrics?

tkburroughs:
@rfelcman I've opened issue #2048 to provide further details about the regression. The Open Liberty performance team reported a 4-6% regression in startup when moving from EclipseLink 4.0.0. to 4.0.1.  The changes proposed here have eliminated that regression in Open Liberty testing.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java
```diff
@@ -41,8 +41,8 @@ public class ASMFactory {
     private final static String ASM_ECLIPSELINK_CLASS_VISITOR = "org.eclipse.persistence.internal.libraries.asm.ClassVisitor";
 
     //Should be changed in case of ASM upgrade
-    public final static int ASM_API_SELECTED = Opcodes.valueInt("ASM9");
```
tkburroughs, line 44: @anija-anil The test failures are caused by a static initialization ordering problem if the Opcodes class is initialized before ASMFactory.  When ASMFactory is initialized first there is no issue; when Opcodes is initialized first, it has a reference to ASMFactory which triggers ASMFactory to then initialize before Opcodes completes initialization and ASM_API_SELECTED ends up set to 0. To avoid the circular initialization error, I'd recommend moving lines 38 and 39 above to Opcodes, and then replacing them here with:

    public final static String ASM_SERVICE_ECLIPSELINK = Opcodes.ASM_SERVICE_ECLIPSELINK;
    public final static String ASM_SERVICE_OW2 = Opcodes.ASM_SERVICE_OW2;

foundation/org.eclipse.persistence.core/src/main/java/module-info.java
```diff
@@ -215,6 +215,7 @@
             org.eclipse.persistence.json,
             org.eclipse.persistence.pgsql,
             org.eclipse.persistence.oracle;
+    exports org.eclipse.persistence.asm.internal;
```
tkburroughs, line 218: This change is not needed; please remove.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/Opcodes.java
```diff
@@ -22,6 +22,7 @@
 
 public class Opcodes {
 
+    //This block must be first - begin
```
tkburroughs, line 25: To avoid a static initialization order problem with ASMFactory, add the following two lines here:

    public final static String ASM_SERVICE_ECLIPSELINK = "eclipselink";
    public final static String ASM_SERVICE_OW2 = "ow2";
    
    Then change lines 32 and 33 below to use the above variables:
    
        ASM_OPCCODES_MAP.put(ASM_SERVICE_OW2, ASM_OPCCODES_OW2);
        ASM_OPCCODES_MAP.put(ASM_SERVICE_ECLIPSELINK, ASM_OPCCODES_ECLIPSELINK);

rfelcman:
There is alternative fix #2050 but when if initialization will be solved there this one should be preferred.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/Opcodes.java
tkburroughs: @anija-anil My prior recommendations do not resolve the issue because Opcodes needs to reference ASMFactory in order to initialize all the static fields.  Looks like the only way to resolve  the circular initialization is to change ASMFactory to not reference any static fields in Opcodes.... so disregard the prior suggestion and instead do something like this in ASMFactory:

```
    private final static String ASM_ECLIPSELINK_CLASS_VISITOR = "org.eclipse.persistence.internal.libraries.asm.ClassVisitor";

    // Note: Do not reference static fields in Opcodes to avoid circular static initialization

    //Should be changed in case of ASM upgrade
    public final static int ASM_API_SELECTED = valueInt("ASM9");
    public final static int JAVA_CLASS_VERSION = valueInt("V1_8");

...
    private static int valueInt(String fieldName) {
        return ((int) Util.getFieldValue(ASM_OPCCODES_MAP, fieldName, Integer.TYPE));
    }
```
This approach eliminates all dependencies that ASMFactory has on Opcodes static fields.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java
tkburroughs: Resolved

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/Opcodes.java
tkburroughs: Resoloved

foundation/org.eclipse.persistence.core/src/main/java/module-info.java
tkburroughs: resolved

tkburroughs (approved):
Latest changes to remove circular initialization look good.

tkburroughs:
@rfelcman The circular static initialization issue between ASMFactory and Opcodes has been resolved; tests that were encountering this are passing.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java
```diff
@@ -26,25 +27,45 @@
...
     private final static String ASM_OW2_CLASS_VISITOR = "org.objectweb.asm.ClassVisitor";
     private final static String ASM_ECLIPSELINK_CLASS_VISITOR = "org.eclipse.persistence.internal.libraries.asm.ClassVisitor";
 
-    //Should be changed in case of ASM upgrade
-    public final static int ASM_API_SELECTED = Opcodes.valueInt("ASM9");
-    public final static int JAVA_CLASS_VERSION = Opcodes.valueInt("V1_8");
+    private final static String ASM_OPCCODES_OW2 = "org.objectweb.asm.Opcodes";
+    private final static String ASM_OPCCODES_ECLIPSELINK = "org.eclipse.persistence.internal.libraries.asm.Opcodes";
+
+    private final static Map<String, String> ASM_OPCCODES_MAP = new HashMap<>();
```
rfelcman, line 48: Duplicit with `org.eclipse.persistence.asm.Opcodes#ASM_OPCCODES_MAP`. Maybe change scope and use this one only.

rfelcman (approved):
LGTM

PR #2052, opened 2024-01-29
Apply new language features

lukasj:
* enhanced switch
* Objects.equals, String.isEmpty, String.contains etc

also fixes duplicated case for W2 in Mongo (added missing W3), and missing optional requires for asm in eclipselink module descriptor and [ReportQueryResult.equals()](https://github.com/eclipse-ee4j/eclipselink/blob/4.0/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/ReportQueryResult.java#L510)


rfelcman (approved):
LGTM

PR #2064, opened 2024-02-14
EL Annotations refresh

lukasj:
* javadoc improvements (snippet tag is used on few places)
* applied Repeatable in OXM/Oracle ann packages


rfelcman (approved):
LGTM

PR #2078, opened 2024-02-27
[master] Jakarta Persistence 3.2: java.time.Instant and java.time.Year support as a basic types

rfelcman:
This change adds support for java.time.Instant and java.time.Year classes according spec 
2.6. Basic Types, 11.1.6. Basic Annotation
https://jakartaee.github.io/persistence/latest-nightly/nightly.html#a486 https://jakartaee.github.io/persistence/latest-nightly/nightly.html#a14205 or see Specification appendix A.1. Jakarta Persistence 3.2 https://jakartaee.github.io/persistence/latest-nightly/nightly.html#jakarta-persistence-3-2

PR #2081, opened 2024-02-29
Jakarta Persistence 3.2: java.time.Instant and java.time.Year - minor fixes

rfelcman:
Fixes for comments in #2078

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
@@ -101,6 +101,7 @@ import java.sql.SQLXML;
 import java.sql.Statement;
 import java.sql.Struct;
 import java.sql.Types;
+import java.time.Year;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/databaseaccess/DatabasePlatform.java
```diff
@@ -2684,7 +2685,7 @@ public int appendParameterInternal(Call call, Writer writer, Object parameter) {
                 java.sql.Timestamp ts = java.sql.Timestamp.valueOf(java.time.LocalDateTime.of(java.time.LocalDate.ofEpochDay(0), ot.toLocalTime()));
                 appendTimestamp(ts, writer);
             } else if (dbValue instanceof java.time.Year){
-                appendNumber((Number)dbValue, writer);
+                appendNumber(((java.time.Year) dbValue).getValue(), writer);
```
pzygielo, line 2688: With the Year imported - perhaps FQCN is not needed here?

rfelcman: I want keep there same pattern and rest of code in the `DatabasePlatform` class and to be sure (avoid any conflict in the future), that "right" `Year` class is used.

lukasj (approved):

PR #2130, opened 2024-05-07
[master] Performance tests improvement - use EmulatedDB

rfelcman:
This change in EclipseLink JPA performance tests replace physical database by emulated one.
Emulation is based on `org.eclipse.persistence.testing.tests.performance.emulateddb.*` classes (JDBC driver).
There are some fixes like new required `EmulatedDatabaseMetaData` class and change in `EmulatedStatement` to distinguish passed SQL queries with parameters.

### DIFF
--- a/foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedConnection.java
+++ b/foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedConnection.java
@@ -45,6 +45,7 @@ import java.util.concurrent.Executor;
 public class EmulatedConnection implements Connection {
     protected Map<String, List<DatabaseRecord>> rows;
     protected Connection connection;
+    protected DatabaseMetaData databaseMetaData;
 
     public EmulatedConnection() {
         this.rows = new HashMap<>();
@@ -327,10 +328,10 @@ public class EmulatedConnection implements Connection {
      */
     @Override
     public DatabaseMetaData getMetaData() throws SQLException {
-        if (connection != null) {
-            return connection.getMetaData();
+        if (this.databaseMetaData == null) {
+            this.databaseMetaData = new EmulatedDatabaseMetaData();
         }
-        return null;
+        return this.databaseMetaData;
     }
 
     /**
--- a/foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedConnection.java
+++ b/foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedConnection.java
@@ -46,9 +46,17 @@ public class EmulatedConnection implements Connection {
     protected Map<String, List<DatabaseRecord>> rows;
     protected Connection connection;
     protected DatabaseMetaData databaseMetaData;
+    private String url;
+    private Properties info;
 
     public EmulatedConnection() {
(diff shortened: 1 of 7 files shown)


foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedConnection.java
```diff
@@ -327,10 +328,10 @@ public boolean isClosed() throws SQLException {
      */
     @Override
     public DatabaseMetaData getMetaData() throws SQLException {
-        if (connection != null) {
-            return connection.getMetaData();
```
lukasj, line 331: if there is an existing connection, metadata should be taken from there; in the other words, this piece should have stayed

performance/eclipselink.perf.test/src/test/java/org/eclipse/persistence/testing/perf/jpa/tests/basic/JPAReadAbstract.java
```diff
@@ -72,25 +92,22 @@ public void testReadEntity() {
...
-            em.getTransaction().commit();
-        } catch (Throwable e) {
-            throw new RuntimeException(e);
-        } finally {
-            if (em.getTransaction().isActive()) {
-                em.getTransaction().rollback();
+    private synchronized void prepareData(EntityManager em) {
+        AbstractSession session = em.unwrap(AbstractSession.class);
+        EmulatedConnection conn = (EmulatedConnection)session.getAccessor().getConnection();
+        Vector<DatabaseField> masterFields = new Vector<>(session.getDescriptor(MasterEntity.class).getAllFields());
```
lukasj, line 98: No need to use Vector here

rfelcman: Due a `org.eclipse.persistence.internal.sessions.ArrayRecord#ArrayRecord(java.util.Vector, org.eclipse.persistence.internal.helper.DatabaseField[], java.lang.Object[])` is `java.util.Vector` needed.

foundation/org.eclipse.persistence.core.test.framework/src/main/java/org/eclipse/persistence/testing/tests/performance/emulateddb/EmulatedDatabaseMetaData.java
```diff
@@ -0,0 +1,913 @@
...
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.RowIdLifetime;
+import java.sql.SQLException;
+
+public class EmulatedDatabaseMetaData implements DatabaseMetaData {
+
+    private EmulatedConnection connection;
+
+    public EmulatedDatabaseMetaData(EmulatedConnection emulatedConnection) {
```
lukasj, line 27: this should probably be a `Connection` instead and where the cast is needed, one should call `unwrap`, WDYT?

rfelcman: I'm not sure if it helps with anything except little bit complicated code. This is our implementation of `DatabaseMetaData` and almost any access to `this.connection` will need cast/unwrap call as specific properties are fetched there.

lukasj: ...and forcing us to use the right API calls in tests... but I'll leave it up to you ;-)

PR #2137, opened 2024-05-09
fixes #2136 - NPE in ExpressionOperator.printCollection 

### LINKED ISSUE #2136: NPE in NullPointerException in ExpressionOperator.printCollection since 2.7.11+, 3.0.0+
**Describe the bug**

Since switching to EL 2.7.14 we are getting a mysterious NPE at ExpressionOperator.printCollection().
The exception happens only occasionally from time to time (like once a week) on different queries.

It also happend to other people #1867 and #1717

**To Reproduce**

I don't know how to reproduce it, but it is clearly a bug in the source code. Read further.

*UPDATE 10.06.2024*
I created a sample project that reproduces the issue: https://github.com/igormukhin/eclipselink-issue-1717-test
The problem is when a queries are running in parallel threads and use COALESCE(...) function or CASE WHEN statement.

**Expected behavior**

Same as before v2.7.11.

*UPDATE 10.06.2024* The following problem description and the solution can be ignored as it is invalid.

**The Problem**

In the class `ExpressionOperator` there is a field `argumentIndices`

Up to 2.7.10 the field was used in the method `printCollection` like this:

```java
    public void printCollection(List<Expression> items, ExpressionSQLPrinter printer) {
        // ... skipped some code ...

        if (this.argumentIndices == null) {
            this.argumentIndices = new int[items.size()];
            for (int i = 0; i < this.argumentIndices.length; i++){
                this.argumentIndices[i] = i;
            }
        }

        // ... skipped some code ...
        for (final int index : argumentIndices) {
            Expression item = items.get(index);
            // ... skipped some code ...
                item.printSQL(printer);
           // ... skipped some code ...
        }
    }

```

In 2.7.11 the line `for (final int index : argumentIndices)` was changed to `for (int i = 0; i < this.argumentIndices.length; i++)`
That is causing the exception.

The field `argumentIndices` can be changed while inside the loop because the loop calls `item.printSQL(printer)` which in turn can call `ExpressionOperator.copyTo` where the field `argumetIndices` on the already used operator instance will be set to `null`.

I can't explain why it does work most of the time and only occasionally fails but it is clearly the reason for the NPE.

**Solution**

While it is not the ultimate solution for the underlying problem (the field can be changed), we still can fix it with just by returning back to caching the value of `argumentIndices` while looping which is already done in 3.1.0-M1 and 4.0.0 but not in 2.7.x, 3.0.x

igormukhin:
The bug is described in detail in #2136

We are looping over an array which is located in the instance field which can be set to `null` while lopping which is causing NPE. To fix it we are caching the value of `argumentIndices` by looping over an iterator of the array.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/ExpressionOperator.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/ExpressionOperator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018, 2022 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
--- a/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/ExpressionOperator.java
+++ b/foundation/org.eclipse.persistence.core/src/org/eclipse/persistence/expressions/ExpressionOperator.java
@@ -2388,8 +2388,7 @@ public class ExpressionOperator implements Serializable {
         }
 
         String[] dbStrings = getDatabaseStrings(items.size());
-        for (int i = 0; i < this.argumentIndices.length; i++) {
-            final int index = this.argumentIndices[i];
+        for (final int index : this.argumentIndices) {
             Expression item = items.get(index);
             if ((this.selector == Ref) || ((this.selector == Deref) && (item.isObjectExpression()))) {
                 DatabaseTable alias = ((ObjectExpression)item).aliasForTable(((ObjectExpression)item).getDescriptor().getTables().firstElement());


igormukhin:
If this PR will be approved, the same thing should be done on the 3.0.x branch

rfelcman:
I'm asking itself how this change helps
`for (int i = 0; i < this.argumentIndices.length; i++) {` -> `for (final int index : this.argumentIndices) {`
with NPE because in case of `int[] argumentIndices = null;` it throws `java.lang.NullPointerException: Cannot read the array length because "<local1>" is null`

igormukhin:
@rfelcman yeah, this is very hard to spot.
 `for (final int index : this.argumentIndices) {` internally creates an iterator that reference the array. So, no matter what happens with the assignment of `this.argumentIndices` the iterator still references the array.

Example:
```java
class Sample {
    private int[] fields;

    public void willWork() {
        fields = new int[] {1,2,3};
        for (int n : this.fields) {
            System.out.println(n);
            this.fields = null;
        }
    }

    public void willFail() {
        fields = new int[] {1,2,3};
        for (int i = 0; i < this.fields.length; i++) {
            int n = this.fields[i];
            System.out.println(n);
            this.fields = null;
        }
    }

    public static void main(String[] args) {
        System.out.println("Will work:");
        new Sample().willWork();

        System.out.println("Will fail:");
        new Sample().willFail();
    }
}
```

rfelcman:
As others mentioned NPE under stress it looks like concurrency issue not simple NPE.

igormukhin:
For our project the issue does not happen under stress. It just happens sometimes.

My first take on the issue was also concurrency, but we also have logs where the issue happens surely when there is only one thread.

Also the best indication for the issue is that it was introduced in 2.7.11 where the line `for (final int index : this.argumentIndices) {` was replaced with `for (int i = 0; i < this.argumentIndices.length; i++) {`

rfelcman (approved):
LGTM just please update copyright year in the class header

PR #2138, opened 2024-05-11
Remove duplicated declaration

quaff:
`EntityManager` and `EntityManagerFactory` already extends `AutoCloseable`.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/JpaEntityManager.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/JpaEntityManager.java
@@ -36,7 +36,7 @@ import org.eclipse.persistence.sessions.server.ServerSession;
  * @author Gordon Yorke
  */
 
-public interface JpaEntityManager extends jakarta.persistence.EntityManager, AutoCloseable {
+public interface JpaEntityManager extends jakarta.persistence.EntityManager {
 
     /**
      * This method returns the current session to the requester.  The current session
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/JpaEntityManagerFactory.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/JpaEntityManagerFactory.java
@@ -31,7 +31,7 @@ import org.eclipse.persistence.sessions.server.ServerSession;
  * </p>
  * @see jakarta.persistence.EntityManagerFactory
  */
-public interface JpaEntityManagerFactory extends EntityManagerFactory, AutoCloseable {
+public interface JpaEntityManagerFactory extends EntityManagerFactory {
 
     /**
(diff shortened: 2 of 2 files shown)


pzygielo:
> Please update copyright year in modified files at the line 2 into `..., 2025 Oracle and/or its affiliates....`

1. Why it should be Oracle in that copyright?
2. What about ECA?

rfelcman:
> > Please update copyright year in modified files at the line 2 into `..., 2025 Oracle and/or its affiliates....`
> 
>     1. Why it should be Oracle in that copyright?
Because EclipseLink started as source closed project Oracle TopLink. Donation to Eclipse.org doesn't mean, that copyright is changed.
> 
>     2. What about ECA?
See https://www.eclipse.org/legal/eca/

pzygielo:
> 
> > ```
> > 2. What about ECA?
> > ```
> 
> See https://www.eclipse.org/legal/eca/

What about ECA of the Author?

pzygielo:
> Because EclipseLink started as source closed project Oracle TopLink. Donation to Eclipse.org doesn't mean, that copyright is changed.
> 

https://en.wikipedia.org/wiki/Oracle_TopLink:

> Toplink was originally developed by The Object People in Smalltalk. It was ported to Java in 1996-1998 and called "TopLink for Java". In 2002, TopLink was acquired by Oracle Corporation and was developed under the Oracle Fusion Middleware product.

rfelcman:
> Hello @waynebeaton - could you share your opinion, please? I wonder if it's really ok to update copyright year for the current one, since the code was donated to EF, even if the contributor is in no relation with original copyright holder? Handbook:
> 
> > Alternatively, additional copyright holders may be added by including multiple copyright statements.
> 
> While this case if of cosmetic, non-essential nature - what about more serious contributions? Should it still be Oracle referenced in header? Unclear.


I'm Oracle employee, so following applies:
Copyright holders are legal entities: either individuals or companies. Note that employment contracts very often stipulate that the intellectual property developed by an employee is the property of the company (in this case, the company should be listed as the copyright holder, not the individual; the individual should still be listed as an author).
From FAQs - https://www.eclipse.org/projects/handbook/#legaldoc-faq - see 4: How does the copyright header change when we accept a change to an existing file?
we prefer to either
you can add a separate line for the new copyright holder (with the year of the new contribution).
or
just add a 2025 year to the first line
in any case, doing nothing is wrong.

pzygielo:
> From FAQs - https://www.eclipse.org/projects/handbook/#legaldoc-faq - see 4: How does the copyright header change when we accept a change to an existing file? 

> we prefer ... just add a 2025 year to the first line

There is no such option there.
I see
> 1. if the current copyright statement indicates copyright of a single entity, and the contribution comes from a different entity, then you can generally append "and others" to the existing copyright statement.

"and others" which is missing here and "and/or its affiliates" is not the same.

> 2. Or, you can add a separate line for the new copyright holder (with the year of the new contribution).

pzygielo:
> What about ECA of the Author? 

Revalidated PR manually and now it's reported in "All checks have passed" box.

PR #2156, opened 2024-05-29
Allow escaping of '?' using '??' in SQL expression (#1935)

pvarga88:
Backport of #1935 to 4.0
Allow escaping of '?' using '??' in SQL expression and testEscapedQuestionMarkInSQLOperator to test ?? in SQL operator


rfelcman (approved):
LGTM

PR #2168, opened 2024-06-12
fixes coalesce and case operators in multithreaded environments (bug #2136)

igormukhin:
The bug is described in the issue #2136

ExpressionOperator.java - printCollection method is modified so it does not modify the shared argumentIndexes field.
ArgumentListFunctionExpression.java - previous "workarounds" are removed


igormukhin:
@rfelcman Radek, sorry for tagging you but I would like move ahead with this PR. This change fixes a severe bug for coalesce function which was reported by multiple people. Can we get it merged somehow? If so, I would also prepare fixes for other versions.

rfelcman (approved):
LGTM

rfelcman:
master is EL 5.0, please make a backport to 4.0 and 3.0 branches

PR #2175, opened 2024-06-19
[master] Virtual Threads: performance impact with Helidon integration

rfelcman:


rfelcman:
This one will be replaced by PR into 4.0 branch due Jakarta EE 10 dependencies used in Helidon.

Issue #2183, opened 2024-07-02
Assignment to computed value misinterpreted in JDQL UPDATE when entity identification variable is ommitted


Riva-Tholoor-Philip:
**Describe the bug**

When the optional entity identification variable is omitted (for consistency with JDQL) of a JPQL query such as the following,

`UPDATE Coordinate SET x = :newX, y = y / :yDivisor WHERE id = :id`

EcilpseLink complains that "The left expression is not an arithmetic expression" instead of interpreting the assignment operation correctly.

```
Caused by: java.lang.IllegalArgumentException: An exception occurred while creating a query in EntityManager: 
Exception Description: Syntax error parsing [UPDATE Coordinate SET x = :newX, y = y / :yDivisor WHERE id = :id]. 
[37, 38] The left expression is not an arithmetic expression.
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1848)
	at io.openliberty.data.internal.persistence.RepositoryImpl.invoke(RepositoryImpl.java:1079)
	... 40 more
Caused by: Exception [EclipseLink-0] (Eclipse Persistence Services - 5.0.0-B02.v202404111748): org.eclipse.persistence.exceptions.JPQLException
Exception Description: Syntax error parsing [UPDATE Coordinate SET x = :newX, y = y / :yDivisor WHERE id = :id]. 
[37, 38] The left expression is not an arithmetic expression.
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:169)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.validate(HermesParser.java:345)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.populateQueryImp(HermesParser.java:292)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildQuery(HermesParser.java:174)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.buildEJBQLDatabaseQuery(EJBQueryImpl.java:144)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.buildEJBQLDatabaseQuery(EJBQueryImpl.java:120)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.<init>(EJBQueryImpl.java:107)
[...]

PR #2218, opened 2024-07-23
Fix for JPQL queries where identifier clashes with function

OndroMih:
Provides test cases and fix for https://github.com/eclipse-ee4j/eclipselink/issues/2187

Adds some more test cases and fixes for implicit variable `this` in JPQL queries.

PR #2222, opened 2024-08-02
Fix for Jakarta data issue #2183

ajaypaul-ibm:
This PR contains the workaround for #2183.(Jakarta Data issue )
This will work if update query is structured like :

- _`UPDATE Coordinate SET x = :newX, y = y / :yDivisor WHERE id = :id`_  (Here we have 2 parameters x and y in SET clause and also we have a "id" in the WHERE clause) 

-  _`UPDATE Coordinate SET x = :newX, y = y / :yDivisor`_ (This solution will work even if WHERE clause is not there)

-  _`UPDATE Coordinate SET x = :newX,`_ (This solution will work for simple update queries )

Please note : This is not a general solution for all types of update queries. This will work only for the type of Jakarta update queries which are mentioned above.


OndroMih:
I believe that #2183 is already fixed in master branch by https://github.com/eclipse-ee4j/eclipselink/pull/2218. Please try it.

Issue #2182, opened 2024-07-02
JPQL SELECT not allowing entity attribute without optional entity identification variable


Riva-Tholoor-Philip:
EcilpseLink is rejecting JPQL SELECT clauses where the entity attribute omits the optional entity identification variable.
For example for, SELECT hexadecimal FROM AsciiCharacter WHERE hexadecimal IS NOT NULL AND thisCharacter = ?1, EclipseLink raises an error: "The identification variable 'hexadecimal' is not defined in the FROM clause". It shouldn't need to be defined the from clause because this JPQL query is intentionally written without the entity identification variable to be consistent with JDQL.

Refer the logs:

```
Caused by: java.lang.IllegalArgumentException: An exception occurred while creating a query in EntityManager: 
Exception Description: Problem compiling [SELECT hexadecimal FROM AsciiCharacter WHERE hexadecimal IS NOT NULL AND thisCharacter = ?1]. 
[7, 18] The identification variable 'hexadecimal' is not defined in the FROM clause.
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1848)
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1869)
	at io.openliberty.data.internal.persistence.RepositoryImpl.invoke(RepositoryImpl.java:871)
	... 40 more
Caused by: Exception [EclipseLink-0] (Eclipse Persistence Services - 5.0.0-B02.v202404111748): org.eclipse.persistence.exceptions.JPQLException
Exception Description: Problem compiling [SELECT hexadecimal FROM AsciiCharacter WHERE hexadecimal IS NOT NULL AND thisCharacter = ?1]. 
[7, 18] The identification variable 'hexadecimal' is not defined in the FROM clause.
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildException(HermesParser.java:169)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.validate(HermesParser.java:358)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.populateQueryImp(HermesParser.java:292)
	at org.eclipse.persistence.internal.jpa.jpql.HermesParser.buildQuery(HermesParser.java:174)
	at org.eclipse.persistence.internal.jpa.EJBQueryImpl.buildEJBQLDatabaseQuery(EJBQueryImpl.java:144)
[...]

PR #1937, opened 2023-09-01
[master] Jakarta Persistence 3.2 new feature - JPQL UNION, INTERSECT, EXCEPT, CAST

rfelcman:
This change doesn't create anything new as specified operators/functions were already specified as a part of EclipseLink JPQL extensions from 2.4 version.
There is refactoring/movement of affected code related with `org.eclipse.persistence.jpa.jpql.parser.CastExpression`, `org.eclipse.persistence.jpa.jpql.parser.DatabaseType`, `org.eclipse.persistence.jpa.jpql.parser.UnionClause` from `EclipseLink...` classes to default JPQL classes.
Due this movement all specified operators/functions are available from Jakarta Persistence/JPQL Grammar 3.2.

EclipseLink issue is described at https://github.com/eclipse-ee4j/eclipselink/issues/1885 and Jakarta Persistence specification request at the https://github.com/jakartaee/persistence/issues/395 (cast) and https://github.com/jakartaee/persistence/issues/398 (union, intersect, and except)

PR #1903, opened 2023-06-20
New JPQL 3.2 and `||` as a String concat operator

rfelcman:
This change contains following changes:

- New JPQL Grammar 3.2 (version matches upcoming Jakarta Persistence 3.2) and EclipseLink JPQL Grammar 4.1
- New JPQL String expression category with related classes (like arithmetic expression)
- New JPQL String operator `||` for string concatenation like JPQL `CONCAT(....)` function
     Tests are available in JPQL parser and JPA JSE test module
      EclipseLink issue is described at #1885 and JPA specification request at the https://github.com/jakartaee/persistence/issues/397
`||` operator is handled at JPQL level, but translated into SQL function `CONCAT(...)` as not every database supports `||` or is mentioned as OR logical operator (MySQL).

PR #2223, opened 2024-08-02
Fix for Jakarta data issue #2182

anija-anil:
This PR includes the fix for https://github.com/eclipse-ee4j/eclipselink/issues/2182.  (Jakarta Data issue ).
This will work with a query like below comes from Jakarta Data 

`[SELECT hexadecimal FROM AsciiCharacter WHERE hexadecimal IS NOT NULL AND thisCharacter = ?1]. `

Please note : This is not a generic solution for all SELECT queries. This will only work for the above-mentioned Jakarta SELECT queries and other straight SELECT queries without an Identification variable.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/JPQLExpression.java
+++ b/jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/JPQLExpression.java
@@ -18,6 +18,9 @@ package org.eclipse.persistence.jpa.jpql.parser;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rfelcman:
Hello,

there are JPQL promised PRs which You can analyze and take as a starting point.

Jakarta Persistence 3.2 new feature - JPQL functions ID(), VERSION() #2108
Jakarta Persistence 3.2: JPQL generate missing Entity default this alias in simple SELECT queries #2102
Jakarta Persistence 3.2 new feature - JPQL UNION, INTERSECT, EXCEPT, CAST #1937
Jakarta Persistence 3.2 new feature - JPQL functions LEFT(), RIGHT() #1933
Jakarta Persistence 3.2 new feature - JPQL functions REPLACE() #1924
New JPQL 3.2 and || as a String concat operator #1903

PR #2259, opened 2024-09-11
Fixes coalesce and case operators in multithreaded environments with different number of arguments

igormukhin:
The bug is described in the issue #2136. The bug was in the 2.7 version, and it was somewhat fixed in later versions but still failed when statements had different number of parameters.

ExpressionOperator.java - fixes getArgumentIndices by disabling the caching of the dynamically created argument indexes.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/ArgumentListFunctionExpression.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/ArgumentListFunctionExpression.java
@@ -97,11 +97,10 @@ public class ArgumentListFunctionExpression extends FunctionExpression {
      */
     @Override
     public void printSQL(ExpressionSQLPrinter printer) {
-        ListExpressionOperator realOperator;
-        realOperator = (ListExpressionOperator)getPlatformOperator(printer.getPlatform());
-        operator.copyTo(realOperator);
-        realOperator.setIsComplete(true);
-        realOperator.printCollection(this.children, printer);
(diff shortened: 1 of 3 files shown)


lukasj (approved):

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/expressions/ArgumentListFunctionExpression.java
```diff
@@ -97,11 +97,10 @@ public void setOperator(ExpressionOperator theOperator) {
      */
     @Override
     public void printSQL(ExpressionSQLPrinter printer) {
-        ListExpressionOperator realOperator;
-        realOperator = (ListExpressionOperator)getPlatformOperator(printer.getPlatform());
```
rfelcman, line 101: Sorry but this removal is problematic as there should be some differences between DB platforms.
Try e.g. `mvn clean verify -pl :org.eclipse.persistence.jpa.jse.test -P derby` against before modification and after to see what happens. `-P derby` means that Implicit Apache Derby DB is used (it is managed started/stopped by EclipseLink Maven build scripts).

igormukhin: Ok, I see what you mean.
In that case the problem is that getPlatformOperator() returns the same cached instance of the operator on each call which leads to errors with the operators with _variable number_ of arguments like "coalesce" and "case when".
I think I have a solution in mind. I will push it later.

rfelcman (approved):
LGTM

PR #2263, opened 2024-09-12
Fixes coalesce and case operators in multithreaded environments with different number of arguments (backport to 3.0)

igormukhin:
The bug is described in the issue #2136.

ExpressionOperator.java - fixes getArgumentIndices by disabling the caching of the dynamically created argument indexes.


rfelcman (commented):
LGTM

PR #2265, opened 2024-09-13
Don't validate if the only change is to a @Version attribute (e.g. em.find() with PESSIMISTIC_FORCE_INCREMENT)

jgallimore:
I wonder if I could propose this change. I'm very happy to incorporate any feedback. Thank you in advance for your reviews and consideration.

A call like:

`Task task = em.find(Task.class, 900, LockModeType.PESSIMISTIC_FORCE_INCREMENT);`

will increment the `@Version` attribute of the entity, and this will be persisted when the transaction completes. The entity will also be validated with Bean Validation at this point, even though there are no other changes to the object. If the data read from the database happens to be invalid according to the bean validation constraints, the validation will fail, and the transaction will be rolled back.

I can't see a reference in the specification as to whether the entity should be validated in this instance, but the behaviour of Eclipselink appears to differ from Hibernate and OpenJPA, which will only attempt to validate the object if a change has has been made.

### DIFF
--- a/jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/main/java/org/eclipse/persistence/testing/models/jpa/beanvalidation/BeanValidationTableCreator.java
+++ b/jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/main/java/org/eclipse/persistence/testing/models/jpa/beanvalidation/BeanValidationTableCreator.java
@@ -22,6 +22,7 @@ public class BeanValidationTableCreator extends TableCreator {
         setName("BeanValidationEmployeeProject");
 
         addTableDefinition(buildProjectTable());
+        addTableDefinition(buildTaskTable());
         addTableDefinition(buildEmployeeTable());
         addTableDefinition(buildEmployeeProjectTable());
     }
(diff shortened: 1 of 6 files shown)


jgallimore:
Hi - just wondering if I can nudge this? If there's anything that this needs to get it over the line, please do let me know, I'm very happy to receive any feedback. Equally, if its a "no", let me know, I'll understand :-)

jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/main/java/org/eclipse/persistence/testing/models/jpa/beanvalidation/BeanValidationTableCreator.java
```diff
@@ -22,6 +22,7 @@ public BeanValidationTableCreator() {
         setName("BeanValidationEmployeeProject");
```
rfelcman, line 22: Please update year in the copyright header into `Copyright (c) 2009, 2024 Oracle.....`.
It's about all updated files _*.java, persistence.xml_

jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/main/java/org/eclipse/persistence/testing/models/jpa/beanvalidation/Task.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2009, 2022 Oracle and/or its affiliates. All rights reserved.
```
rfelcman, line 2: For the new files correct one is `Copyright (c) 2024 Oracle and/or.....`

jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/main/java/org/eclipse/persistence/testing/models/jpa/beanvalidation/Task.java
```diff
@@ -0,0 +1,62 @@
...
+import jakarta.persistence.Id;
+import jakarta.persistence.Column;
+import jakarta.persistence.Version;
+import jakarta.validation.constraints.NotNull;
+
+@Entity(name="CMP3_BV_TASK")
+public class Task {
+
+    @Id
+    private int        id;
```
rfelcman, line 27: Formating

jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/test/java/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -390,6 +399,163 @@ public void testValidateChangedData() {
...
+        EntityManagerFactory factory = getEntityManagerFactory(props);
+        EntityManager em = factory.createEntityManager();
+        try {
+            beginTransaction(em);
+            
+            Task task = em.find(Task.class, 900, LockModeType.PESSIMISTIC_FORCE_INCREMENT);
+            
+            commitTransaction(em);
+            
+            Vector resultSet = getDatabaseSession().executeSQL("select * from CMP3_BV_TASK where ID=900");
```
rfelcman, line 430: What about `Vector<DatabaseRecord>`? We prefer this style instead of generic collections.

jpa/eclipselink.jpa.testapps/jpa.test.beanvalidation/src/test/java/org/eclipse/persistence/testing/tests/jpa/beanvalidation/BeanValidationJunitTest.java
```diff
@@ -390,6 +399,163 @@ public void testValidateChangedData() {
...
+        EntityManagerFactory factory = getEntityManagerFactory(props);
+        EntityManager em = factory.createEntityManager();
+        try {
+            beginTransaction(em);
+            
+            Task task = em.find(Task.class, 901, LockModeType.PESSIMISTIC_FORCE_INCREMENT);
+            task.setPriority(2);
+            
+            commitTransaction(em);
+        } catch (ConstraintViolationException e) {
```
rfelcman, line 477: Multiple `catch` expressions? I think, that this one should split into two separated tests. Each one should cover mentioned exception. In this form is unclear what is exactly tested.

rfelcman (commented):
This change is targeted into `master` branch (upcoming 5.x). As there is behavior change I'm not sure about backports into 4.x and older versions/branches.

jgallimore:
Hi @rfelcman , thank you so much for the review and feedback. I've pushed another commit, which hopefully addresses the feedback. Does this need to be squashed into a single commit? I'm very happy to do that.

On the behavior change - I completely understand. From my own perspective, it would cool to backport it - would a feature flag be something that could enable that? No problem if not, I'm just thinking out loud.

Thanks again for the review. If there's other feedback I'm happy to rework things further.

jgallimore:
Hi - just wondered if there was any further feedback on this? Thanks!

rfelcman:
If You need this change in 4.x version please make backport/PT to 4.0 branch.

jgallimore:
Thank you so much! A backport to 4.x would be amazing - I have created the PR here: https://github.com/eclipse-ee4j/eclipselink/pull/2371

PR #2280, opened 2024-10-10
Fixed persistence provider check in createEntityManagerFactory(PersistenceConfiguration)

Tomas-Kraus:
Added `jakarta.persistence.provider` to `PersistenceUnitProperties`
Added provider class check to `PersistenceProvider#createEntityManagerFactory(PersistenceConfiguration)`
Optimized provider check code.

### DIFF
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
@@ -192,6 +192,10 @@ public class PersistenceProvider implements jakarta.persistence.spi.PersistenceP
     // Never call this method from this class because of stack frames removal.
     @Override
     public EntityManagerFactory createEntityManagerFactory(PersistenceConfiguration configuration) {
+        // Check whether persistence provider is set and matches EclipseLink known classes
+        if (!isProviderEclipseLink(configuration)) {
+            return null;
+        }
         JPAInitializer initializer = getInitializer(configuration.name(), configuration.properties());
         // Root URL from method caller
         StackWalker stackWalker = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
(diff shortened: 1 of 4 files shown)


jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
```diff
@@ -316,29 +320,59 @@ public JPAInitializer getInitializer(String emName, Map m){
...
+     * <p> * Its implementation class has been specified in the provider element for that persistence unit in the persistence.xml
+     * file and has not been overridden by a different jakarta.persistence.provider property value included in the Map passed
+     * to the createEntityManagerFactory method.
+     * <p> * The jakarta.persistence.provider property was included in the Map passed to createEntityManagerFactory and the value
+     * of the property is the provider’s implementation class.
+     * <p> * No provider was specified for the persistence unit in either the persistence.xml or the property map.
+     *
+     * @since Jakarta Persistence 3.2
      */
-    public boolean checkForProviderProperty(Map properties){
```
rfelcman, line 322: Hmm `public` due a JUnit test only. But it means too, that these methods will be as part of public EclipseLink API as `PersistenceProvider` is not in the internal package.

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
```diff
@@ -316,29 +320,59 @@ public JPAInitializer getInitializer(String emName, Map m){
...
-        Object provider = properties.get("jakarta.persistence.provider");
```
rfelcman, line 323: Hard coded String `"jakarta.persistence.provider"` instead of reference `PersistenceUnitProperties.PROVIDER`.

jpa/org.eclipse.persistence.jpa/src/test/java/org/eclipse/persistence/jpa/PersistenceProviderTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+public class PersistenceProviderTest {
+
+    @Test
+    void testCheckForProviderPropertyWithoutProperty() {
+        Map<String, Object> properties = Map.of();
+        boolean result = PersistenceProvider.checkForProviderProperty(properties);
```
rfelcman, line 32: What about tests based on `Persistence.createEntityManagerFactory(` to avoid `public` scope for `checkForProviderProperty` and others. Something like `org.eclipse.persistence.testing.tests.jpa.advanced.EntityManagerJUnitTestSuite#testProviderPropertySetting`.

rfelcman (changes requested):

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
```diff
@@ -316,29 +320,59 @@ public JPAInitializer getInitializer(String emName, Map m){
...
-            try{
-                if (!(EntityManagerFactoryProvider.class.getName().equals(provider) || PersistenceProvider.class.getName().equals(provider))){
-                    return false;
-                    //user has requested another provider so lets ignore this request.
-                }
-            }catch(ClassCastException e){
-                return false;
-                // not a recognized provider property value so must be another provider.
-            }
+    static boolean isProviderEclipseLink(PersistenceConfiguration configuration) {
```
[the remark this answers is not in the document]
Tomas-Kraus: This method uses nothing from the instance so there is no need to be non static.
Also this check is quite common feature so it may be reused elsewhere when needed in the future, just visibiliti may be raised.

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
Tomas-Kraus: No, there is no connection with jUnit, where pkg only is enough. See `isProviderEclipseLink` which is used in jUnit as well.
As you can see, this method was public before my change and that's the only reason why it's public now. I don't want to change public API of this class because you never know if anyone already used checkForProviderProperty int their own code.
I did change it to static, because:
1. it's not related to any content of the class instance
2. calling static metohod trough instance, e.g. `this.` works as well
3. it may be used elsewhere if needed without copy-paste

jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/jpa/PersistenceProvider.java
Tomas-Kraus: Exactly. Red marked code is the one being removed. And there was no `PersistenceUnitProperties.PROVIDER` before this changeset.
That's why I  added it into `PersistenceUnitProperties` and changed hardcoded valut to the static constant reference. See the green part of the diff.

jpa/org.eclipse.persistence.jpa/src/test/java/org/eclipse/persistence/jpa/PersistenceProviderTest.java
Tomas-Kraus: `checkForProviderProperty` WAS public before my changes. There is no relation between this test and method visibility.

rfelcman (approved):
LGTM

PR #2284, opened 2024-10-17
[master] UUID field type mapping - PostgreSQLPlatform

### LINKED ISSUE #2190: UUIDs fail with EclipseLink and Postgres
When creating a query with a UUID, the parameter is set as a VARCHAR instead of a UUID, and is rejected by Postgres.

Here is an example of the failing query:
```
UUID uuid = p1.id;
List<?> result = em.createQuery("SELECT p FROM PurchaseOrder p WHERE p.id=?1")
                   .setParameter(1, uuid).getResultList();
```

rfelcman:
Add support for UUID DB type in PostgreSQL for DDL generation.
This type is supported by PostgreSQL from 2008 see
https://www.postgresql.org/docs/9.3/release-8-3.html
This is why target class is `PostgreSQLPlatform`.
Fixes #2190


Tomas-Kraus (approved):
Looks good

PR #2294, opened 2024-11-06
Allow conversion from OffsetDateTime to java.sql.Timestamp

TheChronicler:


rfelcman (changes requested):
Hello,
   I'm not sure about this change. What about time zone part? It's removed during `.toLocalDateTime()`. Method `convertObjectToTimestamp` is about "Zoneless" date/time types.
I think, that solution in user code based on `@Convert` and `@Converter` is better.

arjantijms:
Let's close this, as Radek wasn't happy and no feedback to counter it was received since.

Issue #2282, opened 2024-10-15
Db2: Eclipselink creates table with Db2 data type DOUBLE instead of REAL for a Java float


mswatosh:
For an entity with Java float attributes, using `jakarta.persistence.schema-generation.database.action` results in columns being created as `Double` (64-bit precision) instead of `Real` (32-bit precision). This is due to Eclipselink attempting to create the column as `Float` instead of `Real`:
`CREATE TABLE Packages (ID INTEGER NOT NULL, DESCRIPTION VARCHAR(255), HEIGHT FLOAT, LENGTH FLOAT, WIDTH FLOAT, PRIMARY KEY (ID))`

This results in the following:

```
                                Data type                     Column
Column name                     schema    Data type name      Length     Scale Nulls
------------------------------- --------- ------------------- ---------- ----- ------
ID                              SYSIBM    INTEGER                      4     0 No    
DESCRIPTION                     SYSIBM    VARCHAR                    255     0 Yes   
HEIGHT                          SYSIBM    DOUBLE                       8     0 Yes   
LENGTH                          SYSIBM    DOUBLE                       8     0 Yes   
WIDTH                           SYSIBM    DOUBLE                       8     0 Yes   
```

`persistence.xml`
``` xml
<property name="jakarta.persistence.schema-generation.database.action" value="drop-and-create"/>
```

`Package.java`
``` java
@Entity
@Table(name = "Packages")
public class Package {

    public String description;

    @Id
    public int id;

    public float height;

    public float length;

    public float width;
```

The end result of this is the first time an entity is persisted on a db2 connection, the float values can be slightly off due to the precision change from 32 bit to 64 bit. This can result in incorrect sorting like this issue: https://github.com/eclipse-ee4j/eclipselink/issues/2194

mswatosh:
I'm not sure if this is the only place, but essentially:

https://github.com/eclipse-ee4j/eclipselink/blob/ec0c9cdda67c6e88813a78ce4b2b3e3968fef15c/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/platform/database/DB2Platform.java#L314C10-L314C84
``` java
        fieldTypeMapping.put(Float.class, new FieldTypeDefinition("FLOAT", false));
        fieldTypeMapping.put(Double.class, new FieldTypeDefinition("FLOAT", false));
```
should be
``` java
        fieldTypeMapping.put(Float.class, new FieldTypeDefinition("REAL", false));
[...]

PR #2296, opened 2024-11-08
Fixes : Db2: Eclipselink creates table with Db2 data type DOUBLE instead of REAL for a Java float

ajaypaul-ibm:
This PR contains the fix for issue #2282


rfelcman (approved):
LGTM

Issue #2011, opened 2023-11-29
BatchFetch IN fails to fetch eager reference when selecting non-root entity


andilem:
EclipseLink version 4.0.2

When selecting a non-root entity (e.g. `query.select(root.get(child))`) and child has a `@ManyToOne` reference with `@BatchFetch(value=IN)`, the references of the children are not fetched. They are just `null`, even with `fetchType=EAGER`.

There is a simple test case attached with
- three entity types with unidirectional (eager) many-to-one references (Record -> User -> Company)
- User -> Company reference has `@BatchFetch(value=IN)`
- `query.select(query.from(Record.class).get("user"))`
- `em.createQuery(query).getResultList()` returns users with `company = null`
- when removing `@BatchFetch(value=IN)`, everything works as expected

[eclipselink-batchfetch-selectnonroot.zip](https://github.com/eclipse-ee4j/eclipselink/files/13499248/eclipselink-batchfetch-selectnonroot.zip)

impacsl:
We have similar issues with OneToOne-references with three entities using BatchFetchType IN/JOIN/EXISTS:

Record - OneToMany / LAZY - Transaction - OneToOne / BatchFetch - Communication

In some cases the List of Transactions contains Communications that are null.

When removing the @BatchFetch it works as well as expected.

EclipseLink 4.0.2, but also other versions (4.0.1, 2.7.14)

Sheikah45:
This issue can probably be closed now with #2303 being merged

rfelcman:
Fix into 4.0 is #2310 . I'm closing it.

PR #2303, opened 2024-11-18
Fix for ReportQuery not fetching relationships with BatchFetch.IN

Sheikah45:
This PR includes the fix for #2011.

A ReportQuery was not populating the DataResults for it's batchFetchPolicy. This prevented the mapping from knowing what keys to use for the in query and always returned null instead. 

For now it is just targeted to master, but I believe it can be backported as you see fit.

### DIFF
--- a/jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
+++ b/jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
@@ -1,0 +1,141 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>org.eclipse.persistence.jpa.testapps</artifactId>
+        <groupId>org.eclipse.persistence</groupId>
+        <version>5.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>org.eclipse.persistence.jpa.testapps.batchfetch</artifactId>
+
+    <name>Test - batchfetch</name>
+
+    <properties>
+        <argLine/>
+    </properties>
+
+    <build>
+        <plugins>
+            <!-- generate canonical metamodel classes -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <executions>
+                    <!-- turn-off default compilation -->
+                    <execution>
+                        <id>default-compile</id>
+                        <phase>none</phase>
+                    </execution>
+                    <!-- turn-on compilation with canonical model processor -->
+                    <execution>
+                        <id>compile-with-processor</id>
+                        <phase>compile</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <!--Resolve dependencies into Maven properties like ${org.eclipse.persistence:org.eclipse.persistence.jpa:jar} for JPA module-->
+                    <execution>
+                        <id>get-test-classpath-to-properties</id>
+                        <phase>process-test-classes</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- start/stop in-memory derby before/after the test -->
+            <plugin>
+                <groupId>org.carlspring.maven</groupId>
+                <artifactId>derby-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>start-derby</id>
+                        <phase>process-test-classes</phase>
+                    </execution>
+                    <execution>
+                        <id>stop-derby</id>
+                        <phase>prepare-package</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <!-- turn on dynamic weaving -->
+                            <argLine>-javaagent:${org.eclipse.persistence:org.eclipse.persistence.jpa:jar} @{argLine}</argLine>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>server-test</id>
+                        <configuration>
+                            <!-- exclude some tests from server-side run -->
+                            <excludes>
+                                <exclude>**/EntityManagerImplTest</exclude>
+                                <exclude>**/EntityManagerFactoryImplTest</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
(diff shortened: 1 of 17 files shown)


Sheikah45:
After some further testing I ran into the issue where the relationship was not populated when it was selected as not the first element in the Report Query and where the number of rows returned by the original query was greater than the batch size. This caused the foreignReferenceMapping to attempt to do an indexOf with the parent rows to find the starting index. However, the parent rows were the full row returned by the query but ReportQueryResult trimmed the database row that was passed to the ForeignReferenceMapping. Due to this the trimmed row didn't exist in the parent rows and so null was returned instead. For this reason I made it so that ReportQueryResult populates the trimmed rows as the parent row for the mappings specific to that item. But it could very well be the case that this is not the best approach and any input is welcome.

It may also be a possibility that the an untrimmed row could be used as it looked like fields were used directly by ReportQueryResult to get the necessary data out of the DatabaseRecord but I was unsure if this was always the case and it wasn't clear why the trimmed row was being used in the first place.

Sheikah45:
Additionally if there is a better place for the tests rather than their own app, let me know and I can move them to the appropriate location.

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    terms of the Eclipse Public License v. 2.0 which is available at
+    http://www.eclipse.org/legal/epl-2.0,
+    or the Eclipse Distribution License v. 1.0 which is available at
+    http://www.eclipse.org/org/documents/edl-v10.php.
+
+    SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
rfelcman, line 15: In this file I'd like ask You for a cleanup see as an example https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/eclipselink.jpa.testapps/jpa.test.jpql/pom.xml

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+        <version>5.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>org.eclipse.persistence.jpa.testapps.batchfetch</artifactId>
+
+    <name>Test - batchfetch</name>
+
+    <properties>
```
rfelcman, line 29: Needed?

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+        <argLine/>
+    </properties>
+
+    <build>
+        <plugins>
+            <!-- generate canonical metamodel classes -->
+            <plugin>
```
rfelcman, line 36: Needed?

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+                            <excludes>
+                                <exclude>**/EntityManagerImplTest</exclude>
+                                <exclude>**/EntityManagerFactoryImplTest</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
rfelcman, line 102: Needed? There are no dependencies to this artifact.

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+                    <execution>
+                        <id>package-se-tests</id>
+                        <phase>package</phase>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <profiles>
```
rfelcman, line 120: Needed?

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/src/main/java/org/eclipse/persistence/jpa/testapps/batchfetch/BatchFetchTableCreator.java
```diff
@@ -0,0 +1,105 @@
...
+        setName("BatchFetchProject");
+
+        addTableDefinition(buildCompanyTable());
+        addTableDefinition(buildEmployeeTable());
+        addTableDefinition(buildRecordTable());
+    }
+
+    public TableDefinition buildRecordTable() {
+        TableDefinition table = new TableDefinition();
+        table.setName("RECORD");
```
rfelcman, line 32: I'd like avoid this too generic name to avoid collision with other tests. Please apply there prefix `BATCH_IN_` -> `BATCH_IN_RECORD`.
This comment can be applied to all new DB tables `COMPANY`, `EMPLOYEE`.

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/src/main/java/org/eclipse/persistence/jpa/testapps/batchfetch/Count.java
```diff
@@ -0,0 +1,3 @@
+package org.eclipse.persistence.jpa.testapps.batchfetch;
```
rfelcman, line 1: Missing copyright

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/src/main/java/org/eclipse/persistence/jpa/testapps/batchfetch/BatchFetchTableCreator.java
Sheikah45: Prefix added

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/src/main/java/org/eclipse/persistence/jpa/testapps/batchfetch/Count.java
Sheikah45: Added

jpa/eclipselink.jpa.testapps/jpa.test.batchfetch/src/main/java/org/eclipse/persistence/jpa/testapps/batchfetch/Record.java
```diff
@@ -0,0 +1,60 @@
...
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+
+package org.eclipse.persistence.jpa.testapps.batchfetch;
```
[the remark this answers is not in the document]
Sheikah45: Packages have been renamed

rfelcman (approved):
LGTM

rfelcman:
Thank You
It will be available in the next 5.0.0-SNAPSHOT when nightly build pass from https://jakarta.oss.sonatype.org/content/groups/staging/org/eclipse/persistence/eclipselink/5.0.0-SNAPSHOT/ .

Sheikah45:
Thanks, and just for information do you think this will be back ported or just part of 5.0.0?

rfelcman:
> Thanks, and just for information do you think this should be back ported? If so I can open PRe against the other branches.

Please open PR against other branches.

Issue #2301, opened 2024-11-13
Sorting not honored with JPA and Oracle 


ajaypaul-ibm:
When running a Jakarta Data test against Oracle there is an issue with sorting not always being honored. 
Note: Similar issue was found in DB2 database which is solved by the PR : https://github.com/eclipse-ee4j/eclipselink/pull/2296(Root cause was problem in insertion of values in DB2)
Recreation steps below :
Test case :
 ```
 @Test
    // @SkipIfSysProp({DB_Oracle }) 
    // Reference issue: https://github.com/OpenLiberty/open-liberty/issues/28289
    public void testOLGH28289() throws Exception {
        deleteAllEntities(Package.class);

        Package p1 = Package.of(70071, 17.0f, 17.1f, 7.7f, "testOLGH28289#70071"); // tallest and smallest length
        Package p2 = Package.of(70077, 77.0f, 17.7f, 7.7f, "testOLGH28289#70077"); // tallest and largest length
        Package p3 = Package.of(70007, 70.0f, 10.7f, 0.7f, "testOLGH28289#70007");

        List<Package> tallToShort;

        tx.begin();
        em.persist(p1);
        em.persist(p2);
        em.persist(p3);
        tx.commit();

        tx.begin();
        try {
            tallToShort = em
                    .createQuery("SELECT o FROM Package o WHERE (o.height<?1) ORDER BY o.height DESC, o.length",
                            Package.class)
                    .setParameter(1, 8.0)
                    .setLockMode(LockModeType.PESSIMISTIC_WRITE) // Cause of issue
                    .setMaxResults(2)
                    .getResultList();
        } catch (Exception e) {
            tx.rollback();
            throw e;
        }

        System.out.println(tallToShort);

        assertEquals(2, tallToShort.size());

        /**
         * Recreate
         *
         * <pre>
         * SELECT * FROM (
         *   SELECT * FROM (
         *     SELECT EL_TEMP.*, ROWNUMBER() OVER() AS EL_ROWNM
         *     FROM (
         *       SELECT ID AS a1, DESCRIPTION AS a2, HEIGHT AS a3, LENGTH AS a4, WIDTH AS a5
         *       FROM PACKAGE
         *       WHERE (HEIGHT < ?)
         *       ORDER BY HEIGHT DESC, LENGTH
         *     ) AS EL_TEMP
         *   ) AS EL_TEMP2 WHERE EL_ROWNM <= ?
[...]

PR #2314, opened 2024-11-22
[3.0] Issue #2301 - Sorting not honored with JPA and Oracle

Tomas-Kraus:
Test was not backported, because 3.0 contains old tests where new patch does not fit.


rfelcman (commented):
Approved

PR #2315, opened 2024-11-22
[2.7] Issue #2301 - Sorting not honored with JPA and Oracle

Tomas-Kraus:
Backport was done manually, original patch does not fit.
Test was not backported, because 3.0 contains old tests where new patch does not fit.


rfelcman (approved):
LGTM

PR #2318, opened 2024-12-03
Fix #2075 Slow startup because of low entropy for PRNG

### LINKED ISSUE #2075: Slow startup because of low entropy for PRNG
**Describe the bug**

b3607a78aae5e1a4c8824baeaa6e0e348965bffd changed JCEEncryptor to use `SecureRandom.getInstanceStrong()`. In OpenJDK on Unix systems this is `NativePRNG.Blocking` which will stall if the kernel does not provide enough entropy.

We've noticed this behaviour on our CI systems where EclipseLink 2.714 startup would stall the tests.

```
   java.lang.Thread.State: RUNNABLE
        at java.io.FileInputStream.readBytes(java.base@11.0.22/Native Method)
        at java.io.FileInputStream.read(java.base@11.0.22/FileInputStream.java:279)
        at java.io.FilterInputStream.read(java.base@11.0.22/FilterInputStream.java:133)
        at sun.security.provider.NativePRNG$RandomIO.readFully(java.base@11.0.22/NativePRNG.java:424)
        at sun.security.provider.NativePRNG$RandomIO.ensureBufferValid(java.base@11.0.22/NativePRNG.java:526)
        at sun.security.provider.NativePRNG$RandomIO.implNextBytes(java.base@11.0.22/NativePRNG.java:545)
        - locked <0x00000000fcca7318> (a java.lang.Object)
        at sun.security.provider.NativePRNG$Blocking.engineNextBytes(java.base@11.0.22/NativePRNG.java:268)
        at java.security.SecureRandom.nextBytes(java.base@11.0.22/SecureRandom.java:751)
        at org.eclipse.persistence.internal.security.JCEEncryptor$Synergizer.getIvGCM(JCEEncryptor.java:243)
        at org.eclipse.persistence.internal.security.JCEEncryptor$Synergizer.access$4(JCEEncryptor.java:235)
        at org.eclipse.persistence.internal.security.JCEEncryptor.encryptPassword(JCEEncryptor.java:102)
        - locked <0x00000000ff7ea7e8> (a org.eclipse.persistence.internal.security.JCEEncryptor)
        at org.eclipse.persistence.sessions.DatasourceLogin.setPassword(DatasourceLogin.java:416)
        at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl.updateLogins(EntityManagerSetupImpl.java:2483)
        at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl.updateSession(EntityManagerSetupImpl.java:2806)
        at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl.deploy(EntityManagerSetupImpl.java:751)
        at org.eclipse.persistence.internal.jpa.EntityManagerFactoryDelegate.getAbstractSession(EntityManagerFactoryDelegate.java:222)
        - locked <0x00000000fcd757a8> (a org.eclipse.persistence.internal.jpa.EntityManagerFactoryDelegate)
        at org.eclipse.persistence.internal.jpa.EntityManagerFactoryDelegate.createEntityManagerImpl(EntityManagerFactoryDelegate.java:330)
        at org.eclipse.persistence.internal.jpa.EntityManagerFactoryImpl.createEntityManagerImpl(EntityManagerFactoryImpl.java:350)
        at org.eclipse.persistence.internal.jpa.EntityManagerFactoryImpl.createEntityManager(EntityManagerFactoryImpl.java:313)
```

**To Reproduce**
Steps/resources to reproduce the behavior:

- EclipseLink version 2.7.14
- Java/JDK version JDK8/11/17/21

**Expected behavior**
The codepath should use the nonblocking PRNG to avoid a blocked startup because of low entropy.

tnagao7:
Fixes #2075.

To address hang-up due to blocking PRNG returned by  `SecureRandom.getInstanceStrong()`, this pull request introduces  a system property (`eclipselink.security.encryptor.use.strong.random.number.generator`) to switch  `SecureRandom.getInstanceStrong()` and `new SecureRandom()`.
`new SecureRandom()` returns the default PRNG of Java platform [1].

Since `SecureRandom.getInstanceStrong()` may return a blocking PRNG and is not recommended in server-side code (see [2]),  we prefer to use `new SecureRandom()` as the default setting.
Users can change the behavior by configuring the system property if they want.

[1] https://docs.oracle.com/en/java/javase/21/security/java-cryptography-architecture-jca-reference-guide.html#GUID-AEB77CD8-D28F-4BBE-B9E5-160B5DC35D36
[2] https://www.blackduck.com/blog/proper-use-of-javas-securerandom.html

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
@@ -128,14 +128,16 @@ public final class JCEEncryptor implements org.eclipse.persistence.security.Secu
             byte[] ivGCM = new byte[IV_GCM_LENGTH];
             SecureRandom random = null;
             String useStrongRNG = PrivilegedAccessHelper.getSystemProperty(SystemProperties.SECURITY_ENCRYPTOR_USE_STRONG_RANDOM_NUMBER_GENERATOR);
-            if (Boolean.parseBoolean(useStrongRNG)) {
+            if (useStrongRNG == null || useStrongRNG.equalsIgnoreCase("false")) {
+                random = new SecureRandom();
+            } else if (useStrongRNG.equalsIgnoreCase("true")) {
                 try {
                     random = SecureRandom.getInstanceStrong();
                 } catch (NoSuchAlgorithmException e) {
                     throw new RuntimeException(e);
                 }
             } else {
-                random = new SecureRandom();
+                throw ValidationException.invalidBooleanValueForProperty(useStrongRNG, SystemProperties.SECURITY_ENCRYPTOR_USE_STRONG_RANDOM_NUMBER_GENERATOR);
             }
             random.nextBytes(ivGCM);
             return ivGCM;
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -14,6 +15,7 @@
 //     Oracle - initial API and implementation from Oracle TopLink
 package org.eclipse.persistence.internal.security;
 
+import org.eclipse.persistence.config.SystemProperties;
 import org.eclipse.persistence.exceptions.ConversionException;
 import org.eclipse.persistence.exceptions.ValidationException;
 import org.eclipse.persistence.internal.helper.Helper;
@@ -125,10 +127,15 @@ public final class JCEEncryptor implements org.eclipse.persistence.security.Secu
         private static byte[] getIvGCM() {
             byte[] ivGCM = new byte[IV_GCM_LENGTH];
             SecureRandom random = null;
-            try {
-                random = SecureRandom.getInstanceStrong();
-            } catch (NoSuchAlgorithmException e) {
-                throw new RuntimeException(e);
+            String useStrongRNG = PrivilegedAccessHelper.getSystemProperty(SystemProperties.SECURITY_ENCRYPTOR_USE_STRONG_RANDOM_NUMBER_GENERATOR);
+            if (Boolean.parseBoolean(useStrongRNG)) {
+                try {
+                    random = SecureRandom.getInstanceStrong();
+                } catch (NoSuchAlgorithmException e) {
+                    throw new RuntimeException(e);
+                }
+            } else {
+                random = new SecureRandom();
             }
             random.nextBytes(ivGCM);
             return ivGCM;
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/SystemProperties.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -270,6 +271,18 @@ public final class SystemProperties {
      * </ul>
      */
     public static final String ASM_SERVICE = "eclipselink.asm.service";
+
+    /**
+     * <p>
+     * This property control the random number generator (RNG) used for password encryption.
+     * <p>
+     * <b>Allowed Values</b> (case sensitive String)<b>:</b>
+     * <ul>
+     * <li>"{@code false}" (DEFAULT) - use default RNG of Java platform
+     * <li>"{@code true}" - use RNG indicated by the securerandom.strongAlgorithms security property of Java platform
+     * </ul>
+     */
+    public static final String SECURITY_ENCRYPTOR_USE_STRONG_RANDOM_NUMBER_GENERATOR = "eclipselink.security.encryptor.use.strong.random.number.generator";
 
     private SystemProperties() {
         // no instance please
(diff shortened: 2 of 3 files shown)


tnagao7:
Can anyone review this patch?

dmatej (approved):
I have just one question - does it need to be new SecureRandom instance every time, with a new seed? Maybe yes as it is used to encrypt database passwords, but I could not resist to ask ;-)

alwinmark:
Hello @rfelcman, 

as you were the last one touching that file, could you please review this Pullrequest? It would be awesome if we could fix this soon as it affects our CI-Builds and also may affect Cloud Applications.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
```diff
@@ -125,10 +127,15 @@ private static SecretKey getAESGCMMultitasker() throws Exception {
         private static byte[] getIvGCM() {
             byte[] ivGCM = new byte[IV_GCM_LENGTH];
             SecureRandom random = null;
-            try {
-                random = SecureRandom.getInstanceStrong();
-            } catch (NoSuchAlgorithmException e) {
-                throw new RuntimeException(e);
+            String useStrongRNG = PrivilegedAccessHelper.getSystemProperty(SystemProperties.SECURITY_ENCRYPTOR_USE_STRONG_RANDOM_NUMBER_GENERATOR);
+            if (Boolean.parseBoolean(useStrongRNG)) {
```
rfelcman, line 131: It will change default behavior into less secure mode by default.
But I'd like rephrase it and change condition content. If users needs it e.g. due performance reasons, they could enable it by system property like
`SECURITY_ENCRYPTOR_USE_NON_BLOCKING_RANDOM_NUMBER_GENERATOR = "eclipselink.security.encryptor.non.blocking.random.number.generator";` -> `random = new SecureRandom();`
For me in this case security  is higher, than performance so I'd like keep as a default `random = SecureRandom.getInstanceStrong();` and conditionally `random = new SecureRandom();`

rfelcman, line 131: I'd like see there validation for input see e.g.
https://github.com/eclipse-ee4j/eclipselink/blob/master/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/asm/ASMFactory.java#L268
and throw `throw ValidationException.incorrect.....` to point user what is wrong.

rfelcman (changes requested):

tnagao7:
> I have just one question - does it need to be new SecureRandom instance every time, with a new seed? Maybe yes as it is used to encrypt database passwords, but I could not resist to ask ;-)

I'm not sure the reason why the original implementation is creating new SecureRandom instance (i.e. calling `SecureRandom.getInstanceStrong()`) every time...

In my recognition, there is at least no security problem with creating new SecureRandom instance every time because of non-deterministic nature of SecureRandom.
If there are other concerns such as performance issues, we can address it with another pull request.

foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/security/JCEEncryptor.java
tnagao7: I understand your concerns.
But IMHO, this pull request does not immediately lower the security level, because it does not change the cipher (AES GCM) for encryption, and using `new SecureRandom()` as random source is still secure.

From the viewpoint of availability, the impact of stall (almost hang in some environments) due to blocking algorithms cannot be ignored.
If a blocking algorithm is used by default, users might encounter the stall in a production environment, which can cause business losses.
There are several related discussions discouraging use of `SecureRandom.getInstanceStrong()` in server-side applications:

- https://www.blackduck.com/blog/proper-use-of-javas-securerandom.html
- https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/general-recommendations-for-secure-coding/#java-specific-recommendations_13
- https://github.com/akka/akka/issues/20908

Should we still use `SecureRandom.getInstanceStrong()`?

tnagao7: Thank you for your advice.
I tried to find examples of boolean value validation to meet your expectation, but I could not find a good example...

- https://github.com/eclipse-ee4j/eclipselink/blob/5.0.0-B04/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConversionManager.java#L115
- https://github.com/eclipse-ee4j/eclipselink/blob/5.0.0-B04/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/Helper.java#L1133
- https://github.com/eclipse-ee4j/eclipselink/blob/5.0.0-B04/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/UnidirectionalOneToManyMapping.java#L401
- https://github.com/eclipse-ee4j/eclipselink/blob/5.0.0-B04/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/helper/ConcurrencyUtil.java#L68-L73

For example, `java.lang.Boolean.parseBoolean()` can be used to parse boolean values, but the method cannot be used for validation since it does not throw exceptions.
Validation could be achieved as in `if (!"true".equalsIgnoreCase(value) && !"false".equalsIgnoreCase(value)) { throw ... }`, but such code is not pretty.
Could you suggest a good way to validate boolean values if you have?

rfelcman: Sorry I don't have anything better. Similar pattern `"true".equals....` is used sometimes inside EclipseLink e.g.
https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/EntityManagerImpl.java#L275 
Just use this "not pretty" pattern and throw `org.eclipse.persistence.exceptions.ValidationException#invalidBooleanValueForProperty`.

rfelcman: OK I accepted Your arguments. I don't like blocked production systems too. Lets use `new SecureRandom` as a default.

tnagao7: I have reflected your comment now. Please check it.
If an invalid value is specified, an exception like below is thrown.
```
Exception [EclipseLink-7278] (Eclipse Persistence Services - 5.0.0.v202412160411-a5af5837377bcad2adc674a0682717d24a436f69): org.eclipse.persistence.exceptions.ValidationException
Exception Description: The specified boolean value [aaa] for the persistence property [eclipselink.security.encryptor.use.strong.random.number.generator] is invalid, the value must either be "true" or "false".
```

rfelcman (approved):
LTGM

rfelcman:
It's going to master branch (5.0.x ) EclipseLink version. If You need it in older versions please make a backports.

PR #2327, opened 2024-12-16
[master] JDK 24 build fixes

rfelcman:
There are just some javadoc related issues which leads into errors and block EclipseLink build on JDK 24.
There are not fixed any compiler warning messages.


Tomas-Kraus (approved):
LGTM

PR #2330, opened 2024-12-17
[4.0] NPE in QueryBasedValueHolder due to #2181 - bugfix

### LINKED ISSUE #2317: NPE in QueryBasedValueHolder due to #2181
### Description
When populating the second-level cache, eclipselink 4.0.4 sometimes throws

Exception in thread "Thread-2" java.lang.NullPointerException: Cannot invoke "org.eclipse.persistence.queries.ReadQuery.isReadObjectQuery()" because "this.query" is null
	at org.eclipse.persistence.internal.indirection.QueryBasedValueHolder.getValue(QueryBasedValueHolder.java:102)
	at org.eclipse.persistence.internal.indirection.UnitOfWorkValueHolder.instantiateImpl(UnitOfWorkValueHolder.java:176)
	at org.eclipse.persistence.internal.indirection.UnitOfWorkValueHolder.instantiate(UnitOfWorkValueHolder.java:252)
	at org.eclipse.persistence.internal.indirection.DatabaseValueHolder.getValue(DatabaseValueHolder.java:109)
	at concurrency.Employee._persistence_get_department(Employee.java)
	at concurrency.Employee.getDepartment(Employee.java:42)
	at concurrency.Main.lambda$main$0(Main.java:23)
	at java.base/java.lang.Thread.run(Thread.java:840)

There is no exception with version 4.0.3. The change that added the bug is #2181.

### To Reproduce
1. Clone the repo https://github.com/Julien-Hanin/eclipselink-query-is-null 
2. Add a breakpoint (suspending only the current thread) on line 102 of org.eclipse.persistence.internal.indirection::QueryBasedValueHolder. 
3. Start debugging concurrency.Main with the java agent enabling weaving
4. Resume program twice when breakpoint is reached

This results in the exception.

### Explanation
The problem is UnitOfWorkValueHolder uses its "wrappedValueHolderLock" to lock its wrapped queryBasedValueHolder instead of using the queryBasedValueHolder's lock.

I've provided a fix (an ugly one) to show this is the issue. You can use it by removing the "_2" at the end of the provided files DatabaseValueHolder_2 and UnitOfWorkValueHolder_2.

### Potential additional issue
The value holders are clonable and their clone method only makes a shallow-copy of the fields. So, after a clone, two cloned value holders have the same lock.
On the other hand, "synchronized(this)" in 4.0.3 synchronizes on two different objects (the two cloned objects). I didn't investigate further though.

rfelcman:
Fixes #2317
There is regression bug introduced in #2181 . Scope of new `ReentrantLock` variable `instanceLock` was in case of new object instances created by `clone()` method across all cloned instances instead of one `instanceLock` per each instance.
In case of `UnitOfWorkValueHolder` incorrect lock was used `wrappedValueHolderLock` vs `wrappedValueHolder.getInstanceLock()`

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/indirection/IndirectList.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/indirection/IndirectList.java
@@ -100,7 +100,7 @@ public class IndirectList<E> extends Vector<E> implements CollectionChangeTracke
      */
     private boolean useLazyInstantiation = true;
 
-    private final Lock instanceLock  = new ReentrantLock();
+    private Lock instanceLock  = new ReentrantLock();
 
     /**
      * PUBLIC:
@@ -355,16 +355,19 @@ public class IndirectList<E> extends Vector<E> implements CollectionChangeTracke
     */
     @Override
     public Object clone() {
-        instanceLock.lock();
+        //Keep origin pointer to lock in local variable as instance variable is updated inside
+        Lock lock = instanceLock;
+        lock.lock();
         try {
             IndirectList<E> result = (IndirectList<E>)super.clone();
             result.delegate = (Vector<E>)this.getDelegate().clone();
             result.valueHolder = new ValueHolder<>(result.delegate);
+            result.instanceLock = new ReentrantLock();
             result.attributeName = null;
             result.changeListener = null;
             return result;
         } finally {
-            instanceLock.unlock();
+            lock.unlock();
         }
     }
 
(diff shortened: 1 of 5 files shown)


Tomas-Kraus (approved):
Looks good.

rfelcman:
First release candidate of EclipseLink 4.0.5 (4.0.5-RC1) is available.
You can download it from usual sources like:
Maven Central
Milestone builds: https://download.eclipse.org/rt/eclipselink/milestones/4.0.5/RC1/
List of changes against 4.0.4 is available at: https://github.com/eclipse-ee4j/eclipselink/releases/tag/4.0.5-RC1

PR #2367, opened 2025-02-17
[master] Clone batchQuery to allow for proper nested lazy instantiation

### LINKED ISSUE #1998: BatchFetch IN corrupts cache by overwriting BatchFetchPolicy dataResults
EclipseLink version 4.0.2

When loading more referenced entities by `@BatchFetch(value = BatchFetchType.IN)` than the BatchFetch size (seems to be fixed 500, the annotation value is not used), multiple queries are issued to retrieve all results. But the `IN (...)` arguments for the last issued query are stored in the cache for every retrieved entity as "source" query arguments, which means that all entities which were not fetched with the last query have corrupted `@OneToMany` references (when loading them, an empty collection is returned because the query uses the wrong IDs).
The cause is BatchFetchPolicy.setDataResults which overwrites the previous rows. I implemented a dirty workaround by adding the results in executeObjectLevelReadQuery() instead of overwriting them, but this could cause bad performance because the number of rows can grow larger than the BatchFetch size, resulting in multiple queries when querying again.

This is relatively complicated to describe but very easy to reproduce. There is a simple test case attached with
- two entity types (`Parent` and `Child`)
- bidirectional `@OneToMany`/`@ManyToOne` with `@BatchFetch(value = BatchFetchType.IN)`
- persist 501 parents with 1 child each
- query all parents -> OK
- query all children
- observe that all parents apart from the 501st one seem to have no children anymore. FINE logging to see the wrong query.

[eclipselink-batchfetch.zip](https://github.com/eclipse-ee4j/eclipselink/files/13349211/eclipselink-batchfetch.zip)

Sheikah45:
Closes #1998 

This PR changes the execution of the batchQuery to utilize a clone for actual execution of the batch fetching. Without the clone it was possible that further batches would overwrite the DataResults of the original batchQuery. These DataResults were then used by the various BatchValueHolders to instantiate the values lazily. However when overwritten the referenced rows had the possibility of not existing in the policy as only the most recently executed batch was present. This could lead to incorrect results being returned. 

Once this is reviewed I will back port it.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/ForeignReferenceMapping.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/mappings/ForeignReferenceMapping.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2025 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2020 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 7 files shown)


rfelcman (approved):
LGTM

Issue #2414, opened 2025-05-15
Aggregated objects cannot be written/deleted/queried independently from their owners


pardhivkrishna:
The JPQL command `SELECT o FROM ShippingAddress o WHERE (o.streetAddress.recipientInfo IS NOT EMPTY)` gets the error: `Aggregated objects cannot be written/deleted/queried independently from their owners.`

Entity and Embeddable classes:

```java
@Entity
public class ShippingAddress {

    @Id
    public Long id;

    public String city;

    public String state;

    @Embedded
    public StreetAddress streetAddress;

    public int zipCode;
}
```

```java
@Embeddable
public class StreetAddress {
    @ElementCollection
    public ArrayList<String> recipientInfo = new ArrayList<String>();

    public int houseNumber;

    public String streetName;

    public StreetAddress() {
    }
}
```

Tables are generated by EclipseLink:
```
[5/14/25, 15:25:34:703 CDT] 00000038 id=00000000 eclipselink.ps.sql                                           3 CREATE TABLE ShippingAddress (ID BIGINT NOT NULL, DTYPE VARCHAR(31), CITY VARCHAR(255), STATE VARCHAR(255), ZIPCODE INTEGER, HOUSENUMBER INTEGER, STREETNAME VARCHAR(255), FLOORNUMBER INTEGER, OFFICE VARCHAR(255), PRIMARY KEY (ID))

[5/14/25, 15:25:34:800 CDT] 00000038 id=00000000 eclipselink.ps.sql                                           3 CREATE TABLE ShippingAddress_RECIPIENTINFO (ShippingAddress_ID BIGINT  /*  */ , RECIPIENTINFO VARCHAR(255))

[5/14/25, 15:25:34:816 CDT] 00000038 id=00000000 eclipselink.ps.sql                                           3 ALTER TABLE ShippingAddress_RECIPIENTINFO ADD CONSTRAINT ShppngShppngddrssD FOREIGN KEY (ShippingAddress_ID) REFERENCES ShippingAddress (ID)
```

Our test case (DataJPATestServlet.testEmbeddable) inserts the following entity:
```java
ShippingAddress a1 = new ShippingAddress();
a1.id = 1001L;
a1.city = "Rochester";
a1.state = "Minnesota";
a1.streetAddress = new StreetAddress(2800, "37th St NW", List.of("Receiving Dock", "Building 040-1"));
a1.zipCode = 55901;
```

[...]

PR #2419, opened 2025-05-29
Issue #2414: Fixed mapping resolution for @Embeddable in ObjectExpression

Tomas-Kraus:
Attmepts to resolve mapping using before last item in derivedExpressions, which should contain proper descriptor of `@Embeddable` class.

`descriptor.isAggregateDescriptor()` check was deleted in `ReportQuery#prepareSubSelect(AbstractSession,  AbstractRecord, Map<Expression, Expression>)` but I'm not sure whether it's OK to turn it off completely.
Anyway, the exception says _"Aggregate object cannot be deleted or written"_, but this is sub SELECT preparation code, not DML.


Tomas-Kraus:
Fixes #2414.

rfelcman (approved):
LGTM

PR #2457, opened 2025-06-20
Issue #2451 - Fixed table generator code to pass values from @Column(secondPrecision = <precision>)

Tomas-Kraus:
Fixed table generator to use secondPrecision value from` @Column` annotation when present.
Fixed in
- MySQL (all versions with fractional seconds precision support)
- Oracle DB (21c, 23ai)
- PostgreSQL (all versions)
- MS SQL (all versions)
- DB2 (all versions)

Fixes #2451

PR #2479, opened 2025-07-29
Target database persistence property values updated.

Tomas-Kraus:
Removed from PR #2457 to add as separate PR.


rfelcman (approved):
LGTM

Issue #2245, opened 2024-08-22
Query on entity with embeddable record fails with "A reference class must be provided


Riva-Tholoor-Philip:
After the fix for https://github.com/eclipse-ee4j/eclipselink/issues/2214 went in, entities with an embeddable record fail further along when trying to query for the entity. EclipseLink fails with A reference class must be provided/

Entity:
```
 <entity class="test.jakarta.data.web.Participant">
  <table name="WLPParticipant"/>
  <attributes>
   <id name="id">
   </id>
   <embedded name="name">
   </embedded>
  </attributes>
 </entity>
```

Class for entity and embeddable record:
```
public class Participant {

    public Integer id;

    public Name name;

    public static record Name(String first, String last) {
    }

    public static Participant of(String firstName, String lastName, int id) {
        Participant p = new Participant();
        p.id = id;
        p.name = new Name(firstName, lastName);
        return p;
    }
}
```

JPQL:
```
SELECT o FROM Participant o WHERE (o.name.last=?1) ORDER BY o.name.first, o.id
```

```
[8/21/24, 13:16:26:979 CDT] 00000070 id=00000000 eclipselink.ps.sql                                           3 SELECT ID, NAMEFIRST, NAMELAST FROM WLPParticipant WHERE (NAMELAST = ?) ORDER BY NAMEFIRST, ID
	bind => [1 parameter bound]
[8/21/24, 13:16:26:980 CDT] 00000070 id=00000000 eclipselink.ps.connection                                    3 Connection released to connection pool [read].
[8/21/24, 13:16:26:983 CDT] 00000070 id=00000000 eclipselink.ps                                               3 CWWKD0291W: Exception [EclipseLink-6029] (Eclipse Persistence Services - 5.0.0.v202408200932-f556522e743c87b8097f78cefe2b8f24f68cf79e): org.eclipse.persistence.exceptions.QueryException
Exception Description: A reference class must be provided.
Query: ReadObjectQuery()
[8/21/24, 13:16:26:983 CDT] 00000070 id=00000000 eclipselink.ps                                               3 throwable 
                                                                                                               Local Exception Stack: 
Exception [EclipseLink-6029] (Eclipse Persistence Services - 5.0.0.v202408200932-f556522e743c87b8097f78cefe2b8f24f68cf79e): org.eclipse.persistence.exceptions.QueryException
Exception Description: A reference class must be provided.
Query: ReadObjectQuery()
[...]

PR #2527, opened 2025-09-19
Issue #2245 - test with @Embeddable record

Tomas-Kraus:
Issue #2245 test with Segment entity and Point embeddable record.
Used `FROM Segment WHERE this.pointB.y < :yExclusiveMax ORDER BY this.pointB.y ASC, this.id ASC` JPQL statement provided there.


rfelcman (approved):
LGTM

PR #2537, opened 2025-09-24
[master] JPQL EXTRACT(TIME FROM X) generates invalid SQL on DB2 - bugfix

pardhiv-krishna:
This PR implements EXTRACT(TIME FROM X) clause in DB2.
See #2536


rfelcman (approved):
LGTM

PR #2604, opened 2025-12-09
[master] Bugfix - Query on entity with embeddable record fails with "A reference class must be provided"

pardhiv-krishna:
Resolves `A reference class must be provided` error happening on embeddable records.
For #2245


anija-anil (approved):
LGTM

PR #2611, opened 2025-12-23
[master] Support UUID through the more efficient BINARY(16) or equivalent

rfelcman:


pvarga88:
@rfelcman  Just FYI. I updated from Eclipselink 4 to 5 and noticed my UUIDs are messed up on Microsoft SQL server. I suspect this change is the cause. The Microsoft SQL server JDBC driver automatically handles conversion between UUID and UNIQUEIDENTIFIER columns. This is not handled in SQLServerPlatform.java. The conversion being done here ends up with a wrong value in the DB because Microsoft treats the data as a GUID, which has different endianness than UUIDs.

PR #2653, opened 2026-02-24
Make sure build passes with Postgres

lukasj:
bugs:
* disabled auto-applied converter should not be applied
* missing field type in DDL for Postgres

tests:
* test.advanced.embeddable model should use numeric/Integer ID instead of string
* test.advanced should use db specific syntax for update
* test.advanced2 should catch generic SQLException
* disabled failing tests in test.jpa.jse, test.jpql and dbws - most of them were not designed for Postgres

other:
* adopt DDLPlatform in Postgres platforms


rfelcman (commented):
LGTM

Issue #2714, opened 2026-04-08
Clarification on ASM New Version Release Process and Timeline


ajaypaul-ibm:
First of all we are not sure it will be appropriate to create an issue for this.
Could you please clarify how new ASM versions are released and the expected timeline?
We’d like to understand the release process  and if any new version is planned. This will help us plan accordingly.

ajaypaul-ibm:
@lukasj @rfelcman  , Could you please check

anija-anil:
@lukasj @rfelcman  We have a requirement to release the latest version of ASM, and we’re currently a bit unsure about the changes. It would be helpful if you could share any relevant information.

lukasj:
ASM as such is 3rd party project, do you mean [eclipselink-asm](https://github.com/eclipse-ee4j/eclipselink-asm) instead?

ajaypaul-ibm:
Yes , we are talking about [eclipselink-asm](https://github.com/eclipse-ee4j/eclipselink-asm) itself

rfelcman:
In EclipseLink 5.0 is dependency on OW2 ASM only due #2008 .
In EclipseLink 4.x is possible use OW2 ASM or https://github.com/eclipse-ee4j/eclipselink-asm . In this case You have to upgrade OW2 ASM dependency in eclipselink-asm like https://github.com/eclipse-ee4j/eclipselink-asm/pull/35/changes#diff-995f5685f8f080d3768f7d19d57601d98c77fe63e868c33042a5c90e90416e03R89 (just dependency should be enough) and release new version see https://github.com/eclipse-ee4j/eclipselink-asm/releases and after release upgrade EclipseLink project dependencies.

anija-anil:
@lukasj @rfelcman 

Our requirement is to update the ASM version and release it to
https://repo1.maven.org/maven2/org/eclipse/persistence/org.eclipse.persistence.asm
.

We have a requirement for Java 27. Initially, I thought upgrading to 9.9.1 would be sufficient and that we could release it to
https://repo1.maven.org/maven2/org/eclipse/persistence/org.eclipse.persistence.asm
 and use that JAR in our product.

However, 9.9.1 is not compatible with Java 27. While checking https://gitlab.ow2.org/asm/asm/ 
, I noticed that the required version has not yet been released. I’m a bit confused about the process and struggling to fully understand the next steps.

Could you please advise on how to proceed? Specifically, whom should we approach for the upcoming release, and what the process would be?

[...]

PR #2715, opened 2026-04-08
[Master] ASM Upgrade 9.9.1

anija-anil:
ASM Upgrade 9.9.1

For #2714


anija-anil:
@lukasj @rfelcman Could you please look into it.

rfelcman (approved):
LGTM

anija-anil:
@rfelcman If my understanding is correct, do I need to make this change in other versions as well?

lukasj (commented):
What do you want to achieve? This won't update ASM used by eclipselink, an update here is about eclipselink features for eclipse platform which is untested.

rfelcman:
See https://github.com/eclipse-ee4j/eclipselink/blob/master/boms/dependencies/pom.xml#L58

anija-anil:
Our requirement is to update the ASM binaries and release the ASM JAR. We have already raised an issue.https://github.com/eclipse-ee4j/eclipselink/issues/2714

lukasj:
then follow comments there in that issue

PR #2760, opened 2026-06-06
Fix SchemaManager.truncate() to drop FK constraints before truncation

### LINKED ISSUE #2759: SchemaManager.truncate() fails on tables with foreign-key constraints (PostgreSQL)
# SchemaManager.truncate() fails on tables with foreign-key constraints (PostgreSQL)

## EclipseLink version
5.0.0 (GA) — also present on `master`

## Database
PostgreSQL (any FK-enforcing DB that rejects `TRUNCATE` of a referenced table)

## Description

`EntityManagerFactory.getSchemaManager().truncate()` (Jakarta Persistence 3.2
`SchemaManager` API) fails when the schema contains foreign-key relationships:

```
org.postgresql.util.PSQLException: ERROR: cannot truncate a table referenced in a foreign key constraint
  Detail: Table "cargos" references "handling_events".
  Hint: Truncate table "cargos" at the same time, or use TRUNCATE ... CASCADE.
Call: TRUNCATE TABLE handling_events
Query: DataModifyQuery(sql="TRUNCATE TABLE handling_events")
```

## Root cause

The truncate path disables foreign-key-constraint awareness, so the
constraint-drop step that is supposed to wrap the truncation is a no-op:

1. `org.eclipse.persistence.internal.jpa.SchemaManagerImpl.truncate()`
   calls `schemaManager.truncateDefaultTables(false)` — passing
   `generateFKConstraints = false`.

2. `SchemaManager.truncateDefaultTables(false)` calls
   `getDefaultTableCreator(false)`, which builds the `TableCreator`/
   `TableDefinition` model **without** any `ForeignKeyConstraint` objects.

3. `TableCreator.truncateTables(session, schemaManager, false)`:
   ```java
   dropConstraints(session, schemaManager, false);   // (a)
   for (TableDefinition table : tables) {
       ... writeTruncateTable(...) -> TRUNCATE TABLE <name>   // (b)
   }
   createConstraints(tables, session, schemaManager, false); // (c)
   ```

4. At (a), `TableCreator.dropConstraints(..., build=false)` iterates the table
   definitions and drops each table's FK constraints — but because the model
   was built with FK = false (step 2), there are **no** constraint objects to
   drop. Nothing is dropped.

5. The real FK constraints created at DDL-generation/deployment time (that path
   runs with FK = true) are still live in the database, so the `TRUNCATE TABLE`
   in (b) is rejected by PostgreSQL.

For comparison, `SchemaManager.dropDefaultTables()` intentionally calls
`getDefaultTableCreator(true)`, which is why dropping tables works while
truncating does not.

Note also that `getDefaultTableCreator` memoizes `defaultTableCreator` on first
call regardless of the boolean argument, so a fix must ensure the FK-aware
creator is the instance actually used for truncation.

## Suggested fix

Have `SchemaManagerImpl.truncate()` use an FK-aware table model so the
constraints are genuinely dropped before `TRUNCATE` and recreated afterward
(e.g. `truncateDefaultTables(true)`), and/or make `getDefaultTableCreator`
rebuild when a different `generateFKConstraints` value is requested.

## Minimal reproducer

Two entities in a one-to-many / many-to-one relationship (FK from child to
parent), on PostgreSQL:

```java
@Entity
public class Parent {
    @Id Long id;
    @OneToMany(mappedBy = "parent")
    List<Child> children;
}

@Entity
public class Child {
    @Id Long id;
    @ManyToOne Parent parent;   // FK column -> PARENT.id
}
```

Steps:
1. Bootstrap an EMF against PostgreSQL with DDL generation creating the tables
   and FK constraint (`CHILD.parent_id -> PARENT.id`).
2. Insert a few rows.
3. Call `emf.getSchemaManager().truncate()`.

Expected: all tables truncated successfully.
Actual: `PSQLException: cannot truncate a table referenced in a foreign key
constraint` — the FK constraints are never dropped before `TRUNCATE`.

## References
- GlassFish issue: https://github.com/eclipse-ee4j/glassfish/issues/25668
- Cargotracker reproducer: https://github.com/hantsy/cargotracker/pull/454

renatsaf:
## Problem

`EntityManagerFactory.getSchemaManager().truncate()` (Jakarta Persistence 3.2 `SchemaManager` API) fails when the schema contains foreign-key relationships, on databases that reject truncating a referenced table:

```
org.postgresql.util.PSQLException: ERROR: cannot truncate a table referenced in a foreign key constraint
  Detail: Table "cargos" references "handling_events".
  Hint: Truncate table "cargos" at the same time, or use TRUNCATE ... CASCADE.
Call: TRUNCATE TABLE handling_events
```

## Root cause

`SchemaManagerImpl.truncate()` called `schemaManager.truncateDefaultTables(false)`. The `false` (`generateFKConstraints`) makes `getDefaultTableCreator` build the `TableDefinition` model **without** any `ForeignKeyConstraint` objects. As a result the `dropConstraints(...)` step that wraps the truncation inside `TableCreator.truncateTables(...)` iterates over tables that carry no constraint metadata and drops nothing. The real FK constraints created at DDL/deploy time stay live, so the subsequent `TRUNCATE TABLE` is rejected by FK-enforcing databases (e.g. PostgreSQL).

This is also inconsistent with the sibling operations in the same class: `create()` uses `createDefaultTables(true)`, `drop()` uses the FK-aware `dropDefaultTables()`, and `validate()` passes `true` — only `truncate()` passed `false`.

## Fix

Pass `true` so the FK-aware table model is used: the constraints are genuinely dropped before each `TRUNCATE` and recreated afterwards. Because `create()`/`drop()`/`validate()` already use the FK-aware model, this also keeps the cached `defaultTableCreator` consistent across the `SchemaManager` lifecycle.

## Test

`SchemaManagerTruncateOnExistingTest` already exercised the truncate path on a model with a `Trainer -> Team` foreign key, but it could not fail:

- its assertions read `Query.getFirstResult()`, which returns the paging offset (always 0) rather than the count, so all three of them passed unconditionally;
- `getDefaultTableCreator()` sets `setIgnoreDatabaseException(true)`, so the `DatabaseException` raised by the rejected `TRUNCATE` is swallowed and the table simply keeps its rows.

The test now counts with `getSingleResult()` on a `Long` typed query, and additionally asserts that the foreign keys reported by `DatabaseMetaData.getImportedKeys` are identical before and after `truncate()`, i.e. that the constraints dropped to allow the truncation are restored.

`PERSISTENCE32_TEAM` is a referenced table, so the corrected test fails without the fix on any database that rejects truncating such a table. Verified against H2:

- without the fix: `Team table shall be empty after truncate() expected:<0> but was:<1>`
- with the fix: passes

The CI database (MySQL/InnoDB) enforces the same rule as PostgreSQL, so the corrected test covers the fix on CI as well.

## Notes

- Fixes #2759 (originally reported downstream as eclipse-ee4j/glassfish#25668).

### DIFF
--- a/jpa/eclipselink.jpa.testapps/jpa.test.persistence32/src/test/java/org/eclipse/persistence/testing/tests/jpa/persistence32/SchemaManagerTruncateOnExistingTest.java
+++ b/jpa/eclipselink.jpa.testapps/jpa.test.persistence32/src/test/java/org/eclipse/persistence/testing/tests/jpa/persistence32/SchemaManagerTruncateOnExistingTest.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation. All rights reserved.
  * Copyright (c) 2023, 2025 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -10,6 +11,14 @@
  * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
  */
 package org.eclipse.persistence.testing.tests.jpa.persistence32;
+
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Locale;
+import java.util.Set;
+import java.util.TreeSet;
 
 import jakarta.persistence.SchemaManager;
 import junit.framework.Test;
@@ -23,6 +32,9 @@ import static org.eclipse.persistence.testing.tests.jpa.persistence32.AbstractPo
  * Test {@link SchemaManager#truncate()} method on database with already existing schema and data.
  */
 public class SchemaManagerTruncateOnExistingTest extends AbstractSchemaManager {
+
+    // Common prefix of all the tables of the persistence unit under test
+    private static final String TABLE_NAME_PREFIX = "PERSISTENCE32_";
 
     public static Test suite() {
         return suite(
@@ -53,22 +65,73 @@ public class SchemaManagerTruncateOnExistingTest extends AbstractSchemaManager {
                 em.persist(TYPES[i]);
             }
         });
+        // Foreign key constraints of the schema before the truncation
+        Set<String> foreignKeysBeforeTruncate = foreignKeys();
         // Truncate the schema
         SchemaManager schemaManager = emf.getSchemaManager();
         schemaManager.truncate();
         // Verify that tables still exist but are empty
-        // - Team count shall be 0
-        int teamCount = emf.callInTransaction(
-                em -> em.createQuery("SELECT count(t) FROM Team t", Integer.class).getFirstResult());
-        assertEquals(teamCount, 0);
-        // - Trainer count shall be 0
-        int trainerCount = emf.callInTransaction(
-                em -> em.createQuery("SELECT count(t) FROM Trainer t", Integer.class).getFirstResult());
-        assertEquals(trainerCount, 0);
-        // - Type count shall be 0
-        int typeCount = emf.callInTransaction(
-                em -> em.createQuery("SELECT count(t) FROM Type t", Integer.class).getFirstResult());
-        assertEquals(typeCount, 0);
+        // Team is referenced by a foreign key from Trainer, so truncating it requires
+        // the constraint to be dropped first on databases that enforce this
+        assertEquals("Team table shall be empty after truncate()", 0L, count("Team"));
+        assertEquals("Trainer table shall be empty after truncate()", 0L, count("Trainer"));
+        assertEquals("Type table shall be empty after truncate()", 0L, count("Type"));
+        // Verify that the schema was left as it was found: constraints dropped to allow
+        // the truncation must be restored
+        assertEquals("Foreign key constraints shall not be modified by truncate()",
+                     foreignKeysBeforeTruncate,
+                     foreignKeys());
+    }
+
+    // Count of the entity instances stored in the database
+    private long count(String entityName) {
+        return emf.callInTransaction(
+                em -> em.createQuery(String.format("SELECT count(e) FROM %s e", entityName), Long.class)
+                        .getSingleResult());
+    }
+
+    // Foreign key constraints of the tables of the persistence unit under test as
+    // "FK_TABLE.FK_COLUMN -> PK_TABLE.PK_COLUMN" values. Generated constraint names are
+    // not part of the values, only the relationships that the constraints define.
+    private Set<String> foreignKeys() {
+        return emf.callInTransaction(em -> {
+            Connection connection = em.unwrap(Connection.class);
+            assertNotNull("Could not access the database connection", connection);
+            try {
+                DatabaseMetaData metaData = connection.getMetaData();
+                String catalog = connection.getCatalog();
+                Set<String> foreignKeys = new TreeSet<>();
+                for (String table : tables(metaData, catalog)) {
+                    try (ResultSet keys = metaData.getImportedKeys(catalog, null, table)) {
+                        while (keys.next()) {
+                            foreignKeys.add(String.format("%s.%s -> %s.%s",
+                                                          keys.getString("FKTABLE_NAME"),
+                                                          keys.getString("FKCOLUMN_NAME"),
+                                                          keys.getString("PKTABLE_NAME"),
+                                                          keys.getString("PKCOLUMN_NAME")));
+                        }
+                    }
+                }
+                return foreignKeys;
+            } catch (SQLException e) {
+                throw new RuntimeException("Could not read the foreign key constraints", e);
+            }
+        });
+    }
+
+    // Names of the tables of the persistence unit under test. Names are read from the database
+    // metadata to match the identifier case used by the database.
+    private static Set<String> tables(DatabaseMetaData metaData, String catalog) throws SQLException {
+        Set<String> tables = new TreeSet<>();
+        try (ResultSet result = metaData.getTables(catalog, null, "%", new String[] {"TABLE"})) {
+            while (result.next()) {
+                String table = result.getString("TABLE_NAME");
+                if (table != null && table.toUpperCase(Locale.ROOT).startsWith(TABLE_NAME_PREFIX)) {
+                    tables.add(table);
+                }
+            }
+        }
+        return tables;
     }
 
 }
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/SchemaManagerImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/SchemaManagerImpl.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 1998, 2026 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation. All rights reserved.
+ * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2024 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
--- a/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/SchemaManagerImpl.java
+++ b/jpa/org.eclipse.persistence.jpa/src/main/java/org/eclipse/persistence/internal/jpa/SchemaManagerImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2026 Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 1998, 2024 IBM Corporation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
@@ -78,7 +78,15 @@ class SchemaManagerImpl implements SchemaManager {
     @Override
     public void truncate() {
         try {
-            schemaManager.truncateDefaultTables(false);
+            // Generate the foreign key constraint metadata (true) so that the
+            // constraints are dropped before each table is truncated and
+            // recreated afterwards. Without it the constraint drop/recreate that
+            // wraps the truncation is a no-op and TRUNCATE fails on databases
+            // that reject truncating a table referenced by a foreign key
+            // (e.g. PostgreSQL). This also keeps truncate() consistent with
+            // create(), drop() and validate(), which all operate with the
+            // foreign-key-aware table model.
+            schemaManager.truncateDefaultTables(true);
         } catch (EclipseLinkException ex) {
             throw new PersistenceException(ex.getMessage(), ex);
         }


hantsy:
I think it is better to add a test for verify it.

hantsy:
And the `truncate` should just clear the data, not modify the database schema(includes constraints).

renatsaf:
Test added in 3e278cd.

The claim in the PR description that CI cannot catch this was wrong. CI runs MySQL (`-Pstaging,oss-release,test-lrg,mysql`), and InnoDB rejects `TRUNCATE TABLE` on a table referenced by a foreign key, the same as PostgreSQL. The bug was invisible for two other reasons:

1. `getDefaultTableCreator()` sets `setIgnoreDatabaseException(true)`, so the per-table `DatabaseException` thrown inside `TableCreator.truncateTables()` is swallowed and the table just keeps its rows.
2. The assertions in `SchemaManagerTruncateOnExistingTest` read `Query.getFirstResult()`, which returns the paging offset (always 0), not the count, so all three passed unconditionally.

The test now uses `getSingleResult()` on a `Long` typed count query. `Trainer` has a `@ManyToOne Team`, which makes `PERSISTENCE32_TEAM` a referenced table. Verified against H2, which rejects truncating a referenced table the same way:

- without the fix: `Team table shall be empty after truncate() expected:<0> but was:<1>`
- with the fix: passes

On the second comment: `truncate()` already drops and recreates constraints on master. `TableCreator.truncateTables()` calls `dropConstraints(...)` before the loop and `createConstraints(...)` after it, and `generateFKConstraints` only decides whether the `TableDefinition`s carry `ForeignKeyConstraint` objects - with `false` both calls iterate over empty lists. Since `getDefaultTableCreator()` caches the creator on the first call and ignores the flag afterwards, while `SchemaManagerImpl` keeps a single core `SchemaManager` for the lifetime of the factory, a `create()` or `validate()` before `truncate()` already leaves an FK-aware creator cached, and the drop/recreate does happen today depending on call order.

To keep that visible, the test also asserts that the foreign keys reported by `DatabaseMetaData.getImportedKeys` are identical before and after `truncate()`.

The remaining risk in this approach is real: `createConstraints()` builds constraints from the derived model, so against a schema created without foreign keys (external DDL, or generation with `generateFKConstraints=false`) `truncate()` would add constraints that were never there. Jakarta Persistence 3.2 says nothing about constraints for `truncate()`, so this is a design decision rather than a compliance one.

If truncation should not touch the schema at all, the alternatives are:

- delete in foreign key dependency order with `DELETE FROM` - portable and transactional, but does not reset identity columns;
- suspend constraint checking for the session through a new `DatabasePlatform` hook (`SET FOREIGN_KEY_CHECKS=0` on MySQL, `SET session_replication_role = replica` on PostgreSQL, `ALTER TABLE ... DISABLE CONSTRAINT` on Oracle, no-op elsewhere).

Both belong in `TableCreator`/`DatabasePlatform` rather than in `SchemaManagerImpl`. Tell me which direction you prefer and I will rework the PR.

hantsy:
From the API docs, https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/schemamanager

```
Truncate the database tables mapped by entities belonging to the persistence
 unit, and then re-import initial data from any configured SQL scripts
 for data loading.

If a SQL operation fails, the behavior is undefined. A provider may throw 
an exception, or it may ignore the problem and continue.
```

After truncate the tables, then initialize the data from SQL script. I am not sure if it will execute the `jakarta.persistence.sql-load-script-source` settings in the `persistence.xml`.

renatsaf:
Confirmed, and it is not implemented. `SchemaManagerImpl.truncate()` only calls `truncateDefaultTables(...)`; the sole caller of `writeSourceScriptToDatabase(...)` in `org.eclipse.persistence.jpa` is `EntityManagerSetupImpl.writeDDL(...)`, i.e. deployment-time schema generation. After `getSchemaManager().truncate()` the tables are empty and `jakarta.persistence.sql-load-script-source` is never replayed. Filed as #2816, since it is orthogonal to the foreign key defect this PR fixes.

On the rest of the javadoc: the operative clause for this PR is "truncate the database tables mapped by entities". The following sentence makes the behaviour undefined when *a SQL operation fails* - it governs error handling, not the obligation to perform the truncation. On master the `TRUNCATE` is rejected by the database and the exception is swallowed by `setIgnoreDatabaseException(true)`, so the call reports success and the rows are still there. That is the defect, whatever the resolution of the constraint question below.

That question is still open, and it decides the shape of the fix:

1. Keep the drop/recreate of the foreign keys, i.e. this PR. Note this is not a new behaviour introduced here - `TableCreator.truncateTables()` already brackets the loop with `dropConstraints(...)`/`createConstraints(...)`, and because `getDefaultTableCreator()` caches the creator on first use, a `create()` or `validate()` before `truncate()` already produces an FK-aware creator and the drop/recreate already happens on master. The one-line change makes it deterministic rather than dependent on call order. Its real drawback is that against a schema created without foreign keys (external DDL, or `generateFKConstraints=false`) the recreate step would add constraints that never existed.
2. Delete in foreign key dependency order with `DELETE FROM`. Portable and transactional, never touches the schema, but does not reset identity columns and is not a truncation.
3. Suspend constraint checking for the session through a new `DatabasePlatform` hook (`SET FOREIGN_KEY_CHECKS=0` on MySQL, `SET session_replication_role = replica` on PostgreSQL, `ALTER TABLE ... DISABLE CONSTRAINT` on Oracle, no-op where unsupported). Schema-neutral and keeps real truncation semantics, at the cost of a new platform API and per-platform implementations.

2 and 3 belong in `TableCreator`/`DatabasePlatform` rather than in `SchemaManagerImpl`. Tell me which you prefer and I will rework the PR accordingly; the corrected test in 3e278cd applies to all three.

Unrelated to this PR but found while reading the class: `create(boolean createSchemas)` and `drop(boolean dropSchemas)` treat the flag as "perform the operation", while the specification defines it as "also create/drop the schemas" ("otherwise, assume the schemas already exist"). `create(false)` and `drop(false)` currently do nothing at all. Filing that separately.

robertpatrick:
@renatsaf I compared the baseline and PR against H2 using a fresh schema manager and found some potentially undesirable behavior changes:

| Starting schema | Baseline | With this PR |
| ---- | ---- | ---- |
| FK matches the generated definition | Parent table retains rows | Fixed: tables emptied; FK preserved |
| Mapped relationship has no database FK | Tables emptied; no FK | New FK introduced |
| Existing FK has ON DELETE CASCADE | CASCADE preserved; parent retains rows | Tables emptied, but CASCADE becomes RESTRICT |
| Existing FK has a different name |  Parent retains rows; one FK | Parent still retains rows; second FK added |

While the drop/recreate is pre-existing machinery, the PR exposes these behaviors on the fresh-manager path.  It would be ideal if you could:

- Preserve actual existing constraints rather than introducing or replacing them from the mapping defaults.
- Add focused tests for absent FKs, customized rules, and external constraint names.

renatsaf:
@robertpatrick Confirmed, all four rows follow from how the drop/recreate works:

- the drop is `ALTER TABLE <table> DROP CONSTRAINT <generated FK_ name>` with the failure ignored (`TableDefinition.dropForeignConstraintsOnDatabase`), so a constraint under another name, or no constraint at all, is left as it is;
- the recreate is built from the mapping, and `ForeignKeyConstraint` can only add `ON DELETE CASCADE`, and only when the mapping carries `@CascadeOnDelete`; `ON UPDATE`, `SET NULL`, `SET DEFAULT` and deferrability are lost;
- `getDefaultTableCreator()` sets `setIgnoreDatabaseException(true)`, so neither a rejected `TRUNCATE` nor a failed recreate is reported.

The same mechanism has two more effects. Master already drops and recreates unique constraints on every `truncate()`: they are built regardless of `generateFKConstraints`, and `requiresUniqueConstraintCreationOnTableCreate()` is `false` on every platform except HANA and Symfoware. And a foreign key from a table outside the persistence unit is lost for good when its recreate fails on the orphaned rows. The test in this PR could not show any of this, because it compared foreign keys without names and rules.

So `truncate()` should not issue constraint DDL at all, which is also hantsy's earlier point. To choose the replacement, I ran the candidate statement sequences with plain JDBC on a parent/child pair: a generated FK name; a custom name with `ON DELETE CASCADE`; an additional self-reference on the child; `DEFERRABLE INITIALLY DEFERRED` where supported. After each sequence the probe checks that both tables are empty, that `getImportedKeys` (name, rules, deferrability) is unchanged, and that inserting an orphan child row is still rejected.

| Database | `TRUNCATE` of a table referenced by another table | `TRUNCATE` of a self-referencing table | `DELETE` of a whole self-referencing table | Works without constraint DDL | Does not work |
|---|---|---|---|---|---|
| H2 2.4.240 | rejected (90106) | rejected (90106) | accepted | `ALTER TABLE ... SET REFERENTIAL_INTEGRITY FALSE` on every referenced table, the self-referenced one included | the same flag on the referencing table only |
| Derby 10.16.1.1 | rejected (XCL48) | accepted | accepted | – | |
| MySQL 8.0.46, 8.4.11 | rejected (1701) | accepted | rejected (1451) | `SET FOREIGN_KEY_CHECKS = 0` | |
| PostgreSQL 17.11 | rejected (0A000) | accepted | accepted | one `TRUNCATE` listing all the tables | `SET session_replication_role = replica` |
| SQL Server 2022 | rejected (4712) | accepted | accepted | – | `NOCHECK CONSTRAINT` |
| Oracle Database 21c XE 21.3, Oracle AI Database 26ai Free 23.26.3 | rejected (ORA-02266) while the referencing table has rows, even with the FK column NULL; accepted once it is empty | accepted | accepted | emptying the referencing table first | |

Everywhere except Oracle the rejection does not depend on data: emptying the referencing table first does not help. On both Oracle versions it does, contrary to the Oracle documentation ("You cannot truncate the parent table of an enabled foreign key constraint"). The `session_replication_role` row also corrects option 3 of my earlier comment.

The bypasses need a guard. When a table outside the truncated set references a truncated table, `FOREIGN_KEY_CHECKS = 0` on MySQL and `REFERENTIAL_INTEGRITY FALSE` on H2 truncate it without error and leave orphaned rows. On Oracle, `DISABLE CONSTRAINT` + `TRUNCATE` + `ENABLE CONSTRAINT` fails in `ENABLE` (ORA-02298) and leaves the constraint disabled, while `getImportedKeys` still reports it unchanged. PostgreSQL's single `TRUNCATE` rejects that case by itself.

No single statement order works on all databases: H2 rejects `TRUNCATE` of a self-referencing table, and MySQL rejects `DELETE` of one. Proposal:

1. `TableCreator.truncateTables()` stops calling `dropConstraints()`/`createConstraints()`, and `SchemaManagerImpl.truncate()` stays on the model without foreign keys. Unique constraints are no longer touched either.
2. References between tables are read from the database (`DatabaseMetaData.getExportedKeys`), not from the mapping.
3. Truncating the set of tables moves behind a `DatabasePlatform` method:
   - default (Oracle, SQL Server, Derby, others): `TRUNCATE` the tables that no other table in the set references (a self-reference does not count), then `DELETE` the referenced ones child-first; a table referenced from outside the set makes its `DELETE` fail on the constraint;
   - MySQL: `SET FOREIGN_KEY_CHECKS = 0` around the `TRUNCATE` statements, restored in `finally`, only when every referencing table is in the set;
   - H2: `ALTER TABLE ... SET REFERENTIAL_INTEGRITY FALSE` on each referenced table in the set and `TRUE` afterwards, under the same condition;
   - PostgreSQL: one `TRUNCATE` listing all the tables.
4. Tests with the schema created by native DDL: a mapped relationship without a database FK, `ON DELETE CASCADE`, a non-generated FK name, a self-reference, and a referencing table outside the unit. Assertions: tables empty (or the failure reported for the outside reference), `getImportedKeys` snapshot including `FK_NAME`, `UPDATE_RULE`, `DELETE_RULE` and `DEFERRABILITY` unchanged, and an orphan insert still rejected, since metadata alone does not show a disabled constraint.

Open question before the rework: should a failed truncation keep being swallowed by `setIgnoreDatabaseException(true)`, or should `truncate()` throw? The specification allows both, but the swallowed failure is what hid #2759.

robertpatrick:
@renatsaf Thanks for digging into this so thoroughly. The database comparisons are helpful, and checking that constraints are still enforced—not just present in the metadata—is a good addition. I agree with moving away from dropping and recreating constraints  and handling the requested set of tables through the database platform.

On your open question, my personal preference would be to prefer the public truncate() call to throw if it cannot complete successfully, after attempting any necessary cleanup. The specification gives us that choice, and silently leaving some tables populated makes failures unnecessarily difficult to diagnose.  If I were doing this, I would try to keep that change specific to truncation rather than changing the error-handling policy for create/drop. 

There is a catch in SchemaManager.truncateDefaultTables() that logs and swallows the exception, so changing setIgnoreDatabaseException alone won’t be enough. We should also be clear that throwing doesn’t mean earlier truncations were rolled back.

Not sure what @arjantijms and other committers think.

I have independently confirmed your Oracle observation on Oracle 21c Enterprise Edition 21.3:

| Child-table state | Parent TRUNCATE |
| ---- | ---- |
| Never populated | Succeeds |
| Contains a referencing row | Rejected |
| Contains a row with a NULL FK | Rejected |
| Emptied by DELETE, then committed | Succeeds |
| Emptied by TRUNCATE | Succeeds |

This held for both ordinary and initially deferred foreign keys. The constraints remained enabled, validated and enforced. The child emptied by DELETE still had an allocated segment, so that result wasn’t just deferred segment creation.

There are a few adjustments I would like to suggest:

- Check outside references before every strategy, including DELETE. We verified on H2, Derby and Oracle that ON DELETE CASCADE can delete outside rows, and SET NULL can change them. The metadata comparison and orphan-insert check still pass  afterward, so the tests should also assert that outside data is unchanged.
- Handle multi-table cycles explicitly. There isn’t always a child-first ordering. If a platform cannot handle a particular cycle safely, let’s reject it before changing other tables.
- Test failure partway through the checks-disabled sequence. For example in MySQL, the whole sequence needs the same physical connection. Restore the original settings, report restoration failures, and don’t assume that turning checks back on validates any rows left behind.

What do you think?

renatsaf:
@robertpatrick Agreed on all three adjustments, and thanks for the independent Oracle run.

**Throwing.** I will take that direction. Two layers swallow the failure today, so both have to change, and only on
the truncate path: `TableCreator.truncateTables()` runs each table under `shouldIgnoreDatabaseException()`, which
`getDefaultTableCreator()` sets to `true`, and `SchemaManager.truncateDefaultTables()` catches `DatabaseException`
and logs it at `FINEST`. `SchemaManagerImpl.truncate()` already wraps an `EclipseLinkException` into a
`PersistenceException`, so once the two swallows are gone the public call reports the failure without further
changes, and `create`/`drop` keep their current policy.

The javadoc will say that a failure does not undo what was already emptied. It cannot: `TRUNCATE` commits
implicitly on MySQL and Oracle, and on H2 the referential integrity switch commits the open transaction as well.
To keep that caveat as small as possible, the checks below run before the first statement, so a rejected request
normally leaves every table untouched.

**Outside references before every strategy.** You are right that the default path is not exempt. I will read
`DELETE_RULE` rather than only the existence of an exported key: `CASCADE`, `SET NULL` and `SET DEFAULT` from a
table outside the requested set are rejected up front, since they change outside rows without any error and
without any change to the metadata; `NO ACTION` and `RESTRICT` are left to the database, which rejects the
`DELETE` itself once the outside table has rows, and that failure is now reported. The tests will snapshot the
contents of the outside table, not only its constraints.

**Cycles.** Agreed, and the check is the same graph: the strongly connected components of the exported keys
restricted to the requested set. On the default path anything larger than a self reference is rejected before the
first statement. The platforms that disable the checks or truncate in one statement handle a cycle natively, so
the rejection is specific to the default path.

**Failure inside a checks-disabled sequence.** All three points apply. The statements go through
`AbstractSession.priviledgedExecuteNonSelectingCall()`, that is through whatever accessor the session hands out,
so on a `ServerSession` with an external pool the `SET` and the `TRUNCATE` can land on different connections: the
sequence has to hold one connection for its whole length, and a platform that cannot get one falls back to the
default path. The restore runs in `finally` and a failing restore is reported rather than logged, because a
connection returned to the pool with the checks off silently disables enforcement for the application. And
re-enabling validates nothing: I measured that `FOREIGN_KEY_CHECKS = 1` on MySQL and `SET REFERENTIAL_INTEGRITY
TRUE` on H2 leave existing orphans in place while the metadata reports the constraint enforced, so after a partial
failure the remaining tables are emptied if possible and the call throws either way.

**Firebird.** One more row for the table, measured with the same probe (Firebird 5.0 LI-V5.0.0.1306, Jaybird
5.0.10):

| Database | `TRUNCATE` of a table referenced by another table | `TRUNCATE` of a self-referencing table | `DELETE` of a whole self-referencing table | Works without constraint DDL | Does not work |
|---|---|---|---|---|---|
| Firebird 5.0 | no `TRUNCATE` statement at all (-104, token unknown) | no `TRUNCATE` statement at all (-104) | rejected (isc 335544466, "Foreign key references are present for the record") | `DELETE` child first; a self-referencing table only when its column is nullable and is set to `NULL` first | `TRUNCATE` in any form; there is no session or table switch for the referential integrity checks |

Three things follow from it.

Firebird has no `TRUNCATE` statement, and no platform overrides `DatabasePlatform.writeTruncateTable()`, so
`truncate()` emits `TRUNCATE TABLE <name>` there as well and fails on the first table with -104, foreign keys or
not. Both swallows hide that, so the current behaviour on Firebird is an empty log line and fully populated
tables. That is the strongest argument for your preference.

Firebird enforces per row on `DELETE`, like MySQL, so a whole-table `DELETE` of a self-referencing table is
rejected; unlike MySQL it has no switch to suspend the checks, only `DROP CONSTRAINT`. Setting the self
referencing column to `NULL` before the `DELETE` works, so that is what its platform will do.

When that column is `NOT NULL`, the table cannot be emptied by any statement: on a three row hierarchy whose root
has the highest id, the plain `DELETE`, `DELETE ... ORDER BY id DESC` and `ORDER BY parent_id DESC` are all
rejected, and the `UPDATE` to `NULL` is rejected by the column. (The ordering only worked in an earlier run where
the ids happened to follow the hierarchy, so it is data dependent and not a candidate.) It is an instance of your
second point, a case the platform has to reject before touching anything else.

@arjantijms one question for the committers before I rework this: `truncate()` never throws today, so an
application whose schema the new code refuses, or whose truncation fails, sees a `PersistenceException` where it
previously saw a silent no-op. Is that acceptable for 5.0, or should the failure be reported some other way?

arjantijms:
@renatsaf I encountered a very similar sounding issue in the 6.0 branch. Maybe a good idea to compare it to there.

renatsaf:
@arjantijms Thanks, found it: `34d9d1c964` on `6.0` switches `SchemaManagerImpl.truncate()` to `truncateDefaultTables(true)` and adds a `DELETE` fallback in `TableCreator.truncateTables()`, and #2842 adds the `checkTableExists` skip. That is the first version of this PR plus the fallback, so the findings above apply to `6.0` as it stands: with the drop/recreate bracket active, a relationship without a database FK gets one, `ON DELETE CASCADE` comes back as `RESTRICT`, and an FK with a non-generated name is left in place next to a newly added one. The fallback adds one more case. When a table outside the persistence unit references a truncated table with `ON DELETE CASCADE` or `SET NULL`, the `DELETE` changes that table's rows without an error. Both failures are still swallowed.

Two parts of it carry over into the rework: the `checkTableExists` skip, and `DELETE` as the default path, which also covers Firebird (it has no `TRUNCATE` statement). The comment's premise needs a correction. Oracle 21c and 26ai accept `TRUNCATE` of the parent once the child is empty, and MySQL and Firebird reject a whole-table `DELETE` of a self-referencing table.

Since `truncateTables` already differs between the branches: should the rework target `master` and then replace the `6.0` code in a forward port, or go to `6.0` first? And the question above still stands: is a `PersistenceException` from `truncate()`, where there was a silent no-op before, acceptable?

PR #2761, opened 2026-06-06
Resolve enum constants in UPDATE/DELETE queries without an identification variable

### LINKED ISSUE #2758: JPQL bulk UPDATE drops fully-qualified enum literal in SET clause when target field is unqualified (no alias)
### Summary

In a JPQL bulk `UPDATE`, a **fully-qualified enum literal** used as the new value of a `SET` item is **silently dropped from the generated SQL** when the target state field is referenced **without an identification variable (alias)**. The resulting SQL is malformed (`SET <COL> = WHERE ...`) and fails at the database.

The same query works correctly if the entity is given an alias and the field is qualified with it. Non-enum literals (e.g. strings) work fine even without an alias. So the failure is specific to the combination **unqualified SET target field + fully-qualified enum literal**.

This was originally reported against GlassFish / Jakarta Data, where alias-less queries (JDQL style) are idiomatic and are forwarded verbatim to `EntityManager.createQuery(...)`: eclipse-ee4j/glassfish#26015

### Environment

- EclipseLink `5.0.0` (build `5.0.0.v202603230926`)
- JDK 21
- Database: H2 2.3.230 (DB-independent — failure is at JPQL→SQL generation, before execution)

### Entity

```java
package com.example.domain;

import jakarta.persistence.*;

@Entity
public class Todo {
    @Id
    Long id;
    String title;
    @Enumerated(EnumType.STRING)
    Status status;
    // getters/setters omitted
}

public enum Status { PENDING, COMPLETED }
```

### Reproduction (results)

Running each statement via `em.createQuery(jpql).setParameter("id", 1L).executeUpdate()`:

| JPQL | Result |
|------|--------|
| `update Todo set status=com.example.domain.Status.PENDING where id=:id` | ❌ **FAIL** |
| `update Todo t set t.status=com.example.domain.Status.PENDING where t.id=:id` | ✅ OK (`status` set to `PENDING`) |
| `update Todo set title='done' where id=:id` | ✅ OK |

### Expected

The first statement should resolve `com.example.domain.Status.PENDING` to the enum literal and update the row, identical to the aliased form.

### Actual

EclipseLink generates SQL with an **empty new value**:

```
UPDATE TODO SET STATUS =  WHERE (ID = ?)
```

which the database rejects, e.g. (H2):

```
org.h2.jdbc.JdbcSQLSyntaxErrorException: Syntax error in SQL statement
"UPDATE TODO SET STATUS =  [*]WHERE (ID = ?)"
```

The enum literal is correctly resolved and emitted only when the target field is qualified by an identification variable (case 2 above).

### Minimal standalone reproduction

`persistence.xml` (RESOURCE_LOCAL, H2 in-memory):

```xml
<persistence xmlns="https://jakarta.ee/xml/ns/persistence"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
             xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence https://jakarta.ee/xml/ns/persistence/persistence_3_2.xsd"
             version="3.2">
  <persistence-unit name="pu" transaction-type="RESOURCE_LOCAL">
    <provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
    <class>com.example.domain.Todo</class>
    <exclude-unlisted-classes>true</exclude-unlisted-classes>
    <properties>
      <property name="jakarta.persistence.jdbc.driver" value="org.h2.Driver"/>
      <property name="jakarta.persistence.jdbc.url" value="jdbc:h2:mem:test;DB_CLOSE_DELAY=-1"/>
      <property name="jakarta.persistence.jdbc.user" value="sa"/>
      <property name="jakarta.persistence.schema-generation.database.action" value="drop-and-create"/>
    </properties>
  </persistence-unit>
</persistence>
```

Driver/runner:

```java
import jakarta.persistence.*;

public class Repro {
    public static void main(String[] a) {
        EntityManagerFactory emf = Persistence.createEntityManagerFactory("pu");
        // seed one row with id=1 first (persist a Todo), then:
        try {
            em.getTransaction().begin();
            em.createQuery("update Todo set status=com.example.domain.Status.PENDING where id=:id")
              .setParameter("id", 1L)
              .executeUpdate();   // throws: SET STATUS =  WHERE ...
            em.getTransaction().commit();
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

### Notes / impact

- Reproduced layer is purely EclipseLink's Hermes JPQL parser (`UpdateQueryVisitor` / `ExpressionBuilderVisitor$PathResolver` enum-literal resolution); no other framework involved.
- This blocks Jakarta Data `@Query` repository methods that set an enum via a literal, because JDQL omits the alias by convention.
- Workarounds: qualify the field with an alias, or bind the value as a parameter instead of an inline literal.

renatsaf:
## Problem

A bulk `UPDATE`/`DELETE` that does not declare an identification variable loses a fully qualified enum constant:

```sql
UPDATE Todo SET status = com.example.domain.Status.PENDING WHERE id = :id
```

generates `UPDATE TODO SET STATUS =  WHERE (ID = ?)`. The constant is lost wherever it appears in such a query (H2):

| JPQL, no identification variable | Before |
|---|---|
| `SET status = <enum>` | `SET STATUS = `, invalid SQL |
| `WHERE status = <enum>` | `QueryException: Object comparisons can only be used with OneToOneMappings` |
| `WHERE status IN (<enum>)` | `STATUS IN ()`, no row matched, no error |
| `SET status = NULLIF(<enum>, status)` | `NULLIF(, STATUS)`, invalid SQL |

This blocks Jakarta Data `@Query` repository methods, whose JDQL omits aliases.

## Root cause

`FullyQualifyPathExpressionVisitor` qualifies every path of such a query with a virtual `this` identification variable, the enum constant included. `ExpressionBuilderVisitor` then navigates `com.example.domain.Status.PENDING` from the entity: the first segment is not a mapping, path resolution stops, and the entity expression itself is used as the value. The existing enum-constant fallback (`resolveEnumConstant`) only runs when the identification variable does not resolve, which never happens for the virtual `this`.

## Fix

In `ExpressionBuilderVisitor.PathResolver`, a path qualified with the implicit `this` whose first segment is neither a mapping nor a query key of the entity is resolved as an enum constant. The parse tree and validation are unchanged, and navigations such as `address.city` or `room.doors` are resolved against the entity as before.

The first commit of this PR changed `FullyQualifyPathExpressionVisitor` to leave multi-segment paths unqualified. As reported in review, that broke nested navigations (`manager.phoneNumbers IS EMPTY` at runtime, `address.city` in the tools validator): without mapping metadata the parser cannot tell a navigation from an enum constant. That commit is reverted.

Type resolution needs no change: `TypeResolver` and the tools `ResolverBuilder` already fall back to the enum type; each query was checked against its aliased form.

## Tests

- `JUnitJPQLJakartaDataNoAliasTest` (runtime): enum constant in `SET`, `SET` + `WHERE`, `DELETE ... WHERE =`, `DELETE ... IN`; `room.status = <enum>`; `room.doors IS NOT EMPTY` in `UPDATE` and `DELETE`.
- `AbstractSemanticValidatorTest`: unqualified `UPDATE`/`DELETE` with `managerEmployee.phoneNumbers IS EMPTY`, `address.city`, `embeddedAddress.city`, `shelfLife.soldDate`, and enum constants in `=` and `IN`.
- `org.eclipse.persistence.jpa.jpql`: 114263 tests, 0 failures. `JUnitJPQLJakartaDataNoAliasTest` on H2: 40/40; without the fix the 5 new enum tests fail. Full `jpa.test.jpql` on H2 with and without the fix: no other test changed status.

Fixes #2758

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/FullyQualifyPathExpressionVisitor.java
+++ b/jpa/org.eclipse.persistence.jpa.jpql/src/main/java/org/eclipse/persistence/jpa/jpql/parser/FullyQualifyPathExpressionVisitor.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2006, 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -147,7 +148,38 @@ public final class FullyQualifyPathExpressionVisitor extends AbstractTraverseChi
         }
     }
 
+    @Override
+    public void visit(UpdateItem expression) {
+
+        // The update target (left side) is always qualified with the virtual identification
+        // variable, even when it navigates through a relationship or an embeddable (for example
+        // "address.city"), so that it can be resolved and validated against the entity.
+        Expression stateFieldExpression = expression.getStateFieldPathExpression();
+        if (stateFieldExpression instanceof AbstractPathExpression) {
+            qualify((AbstractPathExpression) stateFieldExpression);
+        }
+
+        // The new value (right side) is only qualified when it is a single, unqualified attribute.
+        // A multi-segment path is a fully qualified enum constant (e.g.
+        // "com.example.Status.PENDING") and must be left as-is so it gets resolved as an enum
+        // literal rather than as a state field of the entity (see issue #2758).
+        expression.getNewValue().accept(this);
+    }
+
     private void visitAbstractPathExpression(AbstractPathExpression expression) {
+
+        // Only a single, unqualified attribute (e.g. "status") is qualified with the virtual
+        // identification variable. A multi-segment path is left untouched: it is either a fully
+        // qualified enum constant ("com.example.Status.PENDING", see issue #2758) or a navigation
+        // that is resolved later, and prepending the virtual identification variable would wrongly
+        // replace its leading segment. This is consistent with the way an unqualified SELECT query
+        // leaves such paths as-is. The update target is handled separately by visit(UpdateItem).
+        if (expression.pathSize() == 1) {
+            qualify(expression);
+        }
+    }
+
+    private void qualify(AbstractPathExpression expression) {
 
         if (!expression.startsWithDot()) {
 
--- a/jpa/org.eclipse.persistence.jpa.jpql/src/test/java/org/eclipse/persistence/jpa/tests/jpql/parser/JPQLExpressionTestJakartaData.java
+++ b/jpa/org.eclipse.persistence.jpa.jpql/src/test/java/org/eclipse/persistence/jpa/tests/jpql/parser/JPQLExpressionTestJakartaData.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2024 Contributors to the Eclipse Foundation. All rights reserved.
+ * Copyright (c) 2024, 2026 Contributors to the Eclipse Foundation. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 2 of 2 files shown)


robertpatrick:
@renatsaf I compiled the parser source in an isolated directory and compared the PR with the parent’s original visitor. These are executed grammar/semantic-validation results, without a database, and found the following issues:

1. Unaliased DELETE with contact.phoneNumbers IS EMPTY: COLLECTION_VALUED_PATH_EXPRESSION_NOT_RESOLVABLE
2. Unaliased UPDATE with the same predicate: Same failure
3. Nested scalar predicate contact.city = 'x', through the generic tools validator: Field-resolution and undeclared-variable errors

Can you please have a look at the following recommendations?

- Preserve qualification for collection-valued paths. They cannot represent enum literals.
- Preserve existing nested scalar navigation while distinguishing enum literals. 
- Add focused regression tests for each of the failing cases

renatsaf:
@robertpatrick Confirmed. All three reproduce with the tools validators, and 1 and 2 also fail at runtime (`JPQLException: The collection-valued path 'owner.phoneNumbers' cannot be resolved to a valid association field`). Case 3 passes at runtime only because the core semantic validator retries an unresolvable state field path with the implicit `this` and rewrites it; the tools validator does not.

Keeping the qualification for collection-valued paths alone would still leave case 3, and the underlying problem remains: without mapping metadata the parser cannot tell `address.city` from `com.example.Status.DONE`. The parser change is therefore reverted (543112e), and the enum constant is resolved in `ExpressionBuilderVisitor`, where the descriptor is available: a path qualified with the implicit `this` whose first segment is neither a mapping nor a query key of the entity is resolved as an enum constant (a2c9db5). Navigations are qualified and resolved exactly as before.

Regression tests for each case:

- `AbstractSemanticValidatorTest.test_ValidQuery_06` / `_07`: `managerEmployee.phoneNumbers IS EMPTY` in `DELETE` / `UPDATE`
- `AbstractSemanticValidatorTest.test_ValidQuery_08` - `_10`: `address.city`, `embeddedAddress.city`, `shelfLife.soldDate`
- `JUnitJPQLJakartaDataNoAliasTest.testUpdateQueryImplicitThisVariableAndCollectionValuedNavigation`, `testDeleteQueryImplicitThisVariableAndCollectionValuedNavigation`, `testUpdateQueryImplicitThisVariableAndEnumInNavigation` (runtime)

Running the queries against H2 also showed the original defect is not limited to `SET`: `WHERE status IN (<enum>)` matched no rows without an error, and `WHERE status = <enum>` failed with an object-comparison `QueryException`. Both are covered by the new runtime tests.

PR #2269, opened 2024-09-24
Left Join Fetch on Embeddable ElementColection fails & NamedQuery wit…

kalinchan:
# Left Join Fetch
The error occurs when the Entities are compared to see if there is any change done and thus an update needs to be done to the database.

 

[2020-09-01T12:05:10.724+0200] [Payara 5.201] [WARNING] [] [javax.enterprise.web] [tid: _ThreadID=107 _ThreadName=http-thread-pool::http-listener-1(4)] [timeMillis: 1598954710724] [levelValue: 900] [[
StandardWrapperValve[be.rubus.payara.support.ticket1455.ApplicationConfig]: Servlet.service() for servlet be.rubus.payara.support.ticket1455.ApplicationConfig threw exception
java.lang.NullPointerException
at org.eclipse.persistence.internal.descriptors.InstanceVariableAttributeAccessor.getAttributeValueFromObject(InstanceVariableAttributeAccessor.java:83)
at org.eclipse.persistence.mappings.DatabaseMapping.getAttributeValueFromObject(DatabaseMapping.java:691)
at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.compareObjects(AbstractDirectMapping.java:412)
at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.compareForChange(AbstractDirectMapping.java:386)
at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.createObjectChangeSetThroughComparison(DeferredChangeDetectionPolicy.java:192)
at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.createObjectChangeSet(DeferredChangeDetectionPolicy.java:152)
at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.calculateChanges(DeferredChangeDetectionPolicy.java:95)
at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.calculateChangesForExistingObject(DeferredChangeDetectionPolicy.java:61)
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.calculateChanges(UnitOfWorkImpl.java:713)
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.commitToDatabaseWithChangeSet(UnitOfWorkImpl.java:1568)
[...]

Issue #921, opened 2020-10-15
JPA Left Join Fetch on Embeddable ElementColection fails


Cousjava:
Calling left_fetch on an embeddable elementCollection of entities fails. 

Reproducer attached, go to rest/test/failnamed to get
`org.eclipse.persistence.exceptions.ValidationException
Exception Description: The aggregate object [class be.rubus.payara.support.ticket1455.model.Tag] cannot be directly registered in the UnitOfWork.  It must be associated with the source (owner) object.
	at org.eclipse.persistence.exceptions.ValidationException.cannotRegisterAggregateObjectInUnitOfWork(ValidationException.java:622)`

The equivalent SQL query of `SELECT t1.ID, t1.BODY, t1.TITLE, t0.TAGTEXT FROM BLOGPOST t1 LEFT OUTER JOIN BlogPost_TAGS t0 ON (t0.BlogPost_ID = t1.ID)` called directly on the database works.

This bugs occurs with both 2.7 Eclipselink on Payara with H2DB and 3.0 Eclipselink on GlassFish with derby.

[fish-406.zip](https://github.com/eclipse-ee4j/eclipselink/files/5383830/fish-406.zip)

Cousjava:
From debugging and a heap dump I found that there are two instances of RelationalDescriptor the are for the `Tag` class (the `@Embeddable` entity) in the reproducer, one of which has not had all its fields initialised, while the second one has. 

This gives rise to exceptions like `java.lang.NullPointerException at org.eclipse.persistence.internal.descriptors.InstanceVariableAttributeAccessor.getAttributeValueFromObject(InstanceVariableAttributeAccessor.java:83)
`

PR #2786, opened 2026-07-23
[master] Do not calculate changes for aggregates registered in the UnitOfWork

### LINKED ISSUE #1455: bug when querying an embedded class inside an element collection and commit (without changes)
There seems to be a fundamental bug when handling with element collections of embeddable:

Short description of the bug:

I query a set of embeddables that are contained in an element collection, then I begin a transaction and commit without changes, I get a NPE.

Detailed description of the bug:

I have two classes (one entity class called "Entity" with id called "id" and one embeddable class called "Embed")

```java
public class Entity {
	private int id;
	private Set<Embed> embeds;
}

public class Embed {
	String property;
}
```

The annotation is the following:

```xml
	<entity class="main.Entity" >
		<attributes>
			<id name="id">
			</id>
			<element-collection name="embeds">
			</element-collection>
		</attributes>
	</entity>

	<embeddable class="main.Embed">
		<attributes>
			<basic name="property">
			</basic>
		</attributes>
	</embeddable>
```

Then I insert some sample data in the DB, with for example the following code:

```java
Entity inputEntity = new Entity();
inputEntity.setId(14);

Embed inputEmbed = new Embed();
inputEmbed.setProperty("prop");

Set<Embed> embeds = new HashSet<Embed>();
embeds.add(inputEmbed);
inputEntity.setEmbeds(embeds);

em.getTransaction().begin();
em.merge(inputEntity);
em.getTransaction().commit();
```
		
Everything works fine so far. Now, I start a different program  (which results in NPE):

```java
TypedQuery<Embed> query = em.createQuery(
				"SELECT y FROM main.Entity x JOIN x.embeds y ",
						Embed.class);
		
List<Embed> embeds = query.getResultList();
		
em.getTransaction().begin();
		
em.getTransaction().commit();
```
		
I get the following error:

```
Exception in thread "main" jakarta.persistence.RollbackException: Exception [EclipseLink-69] (Eclipse Persistence Services - ${release.version}.${build.qualifier}): org.eclipse.persistence.exceptions.DescriptorException

Exception Description: A NullPointerException was thrown while extracting a value from the instance variable [property] in the object [main.Embed].

Internal Exception: java.lang.NullPointerException

Mapping: org.eclipse.persistence.mappings.DirectToFieldMapping[property-->PROPERTY]
Descriptor: RelationalDescriptor(main.Embed --> [])
	at org.eclipse.persistence.internal.jpa.transaction.EntityTransactionImpl.commit(EntityTransactionImpl.java:163)
	at main.Main.main(Main.java:53)

Caused by: Exception [EclipseLink-69] (Eclipse Persistence Services - ${release.version}.${build.qualifier}): org.eclipse.persistence.exceptions.DescriptorException
Exception Description: A NullPointerException was thrown while extracting a value from the instance variable [property] in the object [main.Embed].
Internal Exception: java.lang.NullPointerException
Mapping: org.eclipse.persistence.mappings.DirectToFieldMapping[property-->PROPERTY]
Descriptor: RelationalDescriptor(main.Embed --> [])
	at org.eclipse.persistence.exceptions.DescriptorException.nullPointerWhileGettingValueThruInstanceVariableAccessor(DescriptorException.java:1280)
	at org.eclipse.persistence.internal.descriptors.InstanceVariableAttributeAccessor.getAttributeValueFromObject(InstanceVariableAttributeAccessor.java:95)
	at org.eclipse.persistence.mappings.DatabaseMapping.getAttributeValueFromObject(DatabaseMapping.java:691)
	at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.compareObjects(AbstractDirectMapping.java:409)
	at org.eclipse.persistence.mappings.foundation.AbstractDirectMapping.compareForChange(AbstractDirectMapping.java:382)
	at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.createObjectChangeSetThroughComparison(DeferredChangeDetectionPolicy.java:197)
	at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.createObjectChangeSet(DeferredChangeDetectionPolicy.java:156)
	at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.calculateChanges(DeferredChangeDetectionPolicy.java:98)
	at org.eclipse.persistence.descriptors.changetracking.DeferredChangeDetectionPolicy.calculateChangesForExistingObject(DeferredChangeDetectionPolicy.java:63)
	at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.calculateChanges(UnitOfWorkImpl.java:727)
	at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.commitToDatabaseWithChangeSet(UnitOfWorkImpl.java:1582)
	at org.eclipse.persistence.internal.sessions.RepeatableWriteUnitOfWork.commitRootUnitOfWork(RepeatableWriteUnitOfWork.java:288)
	at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.commitAndResume(UnitOfWorkImpl.java:1234)
	at org.eclipse.persistence.internal.jpa.transaction.EntityTransactionImpl.commit(EntityTransactionImpl.java:138)
```	
	
	
I use the following versions:

```xml
<dependency>
	<groupId>org.eclipse.persistence</groupId>
	<artifactId>eclipselink</artifactId>
	<version>3.0.2</version>
</dependency>
		
<dependency>
	<groupId>jakarta.persistence</groupId>
	<artifactId>jakarta.persistence-api</artifactId>
	<version>3.0.0</version>
</dependency>
```
	
	
I did some debugging and found out what and why causes the error:

First thing that seems to be odd is that in the commit there are different Embedd Objects which are compared to each other for changes (in the class AbstractDirectMapping in the function compareObjects). In order to compare these objects during the commit the program uses the ClassDescriptor for the Embedded Class "Embed". Especially, it takes the DirectToFieldMapping (for its field "property") in order to access the fields and compare for change. However, if I look at the attributeAccessor of this mapping I observe that "attributeField" is set to NULL.

This ultimatively causes the error in the class InstanceVariableAttributeAccessor Line 83 "return this.attributeField.get(anObject);".
I dont know if an attributeField of an attributeAccessor for an embeddable class should generally be NULL or not.

My questions:
Can this be fixed easily? Is this expected behavior? What are possible workarounds? I tried to query the Entity class (instead of the Embed class)
and iterated of all embeds, this work without errors, even if i modify the embedds and commit the changes.
And the final and most important question for me is are the other bugs when working with element collections of embeddeded classes,
and should element collections of embeddeded be avoided if possible due to buggy behavior?

I would be very happy to hear from you.
Best regards

Test project attached:
[sources.zip](https://github.com/eclipse-ee4j/eclipselink/files/8274747/sources.zip)

andilem:
Fixes #1455. Related: #921, #1465 (and the approach previously attempted in #2269).

### Symptom

Committing a UnitOfWork fails with:

```
Exception [EclipseLink-70] (Eclipse Persistence Services)
org.eclipse.persistence.exceptions.DescriptorException
Exception Description: A NullPointerException was thrown while extracting a value
through the method [getX] in the object [SomeEmbeddable].
Internal Exception: java.lang.NullPointerException
Mapping: org.eclipse.persistence.mappings.DirectToFieldMapping[x-->X]
Descriptor: RelationalDescriptor(SomeEmbeddable --> [])
```

from `UnitOfWorkImpl.calculateChanges`, whenever an `@Embeddable` instance ends up in the
UnitOfWork's clone mapping. Note the empty table list in the descriptor — that is the tell.

### Analysis

Aggregates have no identity of their own, and EclipseLink deliberately refuses to register
them: `internalRegisterObject`, `registerNewObject` and `registerExistingObject` all throw
`cannotRegisterAggregateObjectInUnitOfWork` when `descriptor.isDescriptorTypeAggregate()`.

The read path bypasses those guards. `ObjectBuilder.buildWorkingCopyCloneFromRow` writes the
clone straight into `unitOfWork.getCloneMapping()`, and it is reached for an aggregate whenever
the aggregate rows are built by an `ObjectLevelReadQuery` of their own. Two routes:

* **`@BatchFetch` on an `@ElementCollection` of an `@Embeddable`.** When the owning entity is
  built from a `ReadAllQuery` row, `ForeignReferenceMapping.valueFromRow` takes the batch branch,
  `CollectionMapping.executeBatchQuery` runs a `ReadAllQuery` whose descriptor is the aggregate
  descriptor, and `ReadAllQuery.registerResultInUnitOfWork` registers the embeddables. Reproduces
  reliably when the owner is read into the UnitOfWork rather than the shared cache — e.g. under
  `LockModeType.PESSIMISTIC_WRITE`, which begins the transaction early and isolates the read.
* **Selecting the embeddable directly**, as in #1455 (`SELECT y FROM Entity x JOIN x.embeds y`),
  or a `LEFT JOIN FETCH` on an embeddable element collection (#921, #1465).

Once such an object is in the clone mapping, `calculateChanges` resolves its descriptor via
`getDescriptor(object)`. That returns the **prototype** aggregate descriptor held by the session —
not the per-mapping clone that `AggregateCollectionMapping.initializeReferenceDescriptor` builds
and initializes. The prototype is never initialized, so its mappings' attribute accessors have no
reflective `Method`/`Field`, and comparing the object throws.

### Fix

Skip aggregates in that loop. Their changes are computed by the owning `AggregateObjectMapping` /
`AggregateCollectionMapping` against the initialized descriptor clone, so calculating them again
from the root loop is both wrong and redundant.

### Note on scope

This guard stops the aggregate from being change-tracked, but leaves it in the clone mapping.
The other iterations over that map are `basicPrintRegisteredObjects()` (debug output),
`revertAndResume()` and `validateObjectSpace()` — the latter two would still trip over an
aggregate, though neither is called internally by EclipseLink or the JPA layer. A deeper fix
would keep aggregates out of `cloneMapping` in `ObjectBuilder.buildWorkingCopyCloneFromRow`
altogether; happy to take that route instead if maintainers prefer it.

### Testing

Verified against a local application: the reproducer above fails without the patch and passes with it, and a full run of 92 test tasks shows no regressions. No test is included here — glad to add one to the JPA suite if you'd
like, with a pointer to where it best belongs.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/sessions/UnitOfWorkImpl.java
@@ -719,6 +719,18 @@ public class UnitOfWorkImpl extends AbstractSession implements org.eclipse.persi
 
             ClassDescriptor descriptor = getDescriptor(object);
 
+            // Aggregates (embeddables) have no identity of their own and must never be change-tracked as
+            // roots: their changes are calculated by the owning AggregateObjectMapping /
+            // AggregateCollectionMapping, which uses its own initialized *clone* of the aggregate
+            // descriptor. The descriptor returned here is the prototype registered on the session; its
+            // mappings are never initialized, so the attribute accessors have no reflective Method/Field
+            // and comparing the object throws a NullPointerException wrapped in DescriptorException.
+            // Aggregates can reach the clone mapping through ObjectBuilder.buildWorkingCopyCloneFromRow,
+            // which bypasses the cannotRegisterAggregateObjectInUnitOfWork guards of the register methods.
+            if (descriptor.isDescriptorTypeAggregate()) {
+                continue;
+            }
+
             // Update any derived id's.
             updateDerivedIds(object, descriptor);
 
--- a/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/mapping/TestAggregateChangeCalculation.java
+++ b/jpa/eclipselink.jpa.test.jse/src/it/java/org/eclipse/persistence/jpa/test/mapping/TestAggregateChangeCalculation.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     07/23/2026 - Andreas Lemmer
+//       - 1455: Do not calculate changes for aggregates registered in the UnitOfWork
+package org.eclipse.persistence.jpa.test.mapping;
+
+import java.util.List;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.EntityManagerFactory;
+import jakarta.persistence.LockModeType;
+
+import org.eclipse.persistence.jpa.test.framework.DDLGen;
+import org.eclipse.persistence.jpa.test.framework.Emf;
+import org.eclipse.persistence.jpa.test.framework.EmfRunner;
+import org.eclipse.persistence.jpa.test.mapping.model.AggregateChangeEmbeddable;
+import org.eclipse.persistence.jpa.test.mapping.model.AggregateChangeEntity;
+import org.eclipse.persistence.jpa.test.mapping.model.BatchFetchAggregateChangeEntity;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Aggregates (embeddables) have no identity of their own, so they must never be change-tracked as
+ * roots of the UnitOfWork. They can nevertheless end up in the UnitOfWork's clone mapping, because
+ * ObjectBuilder.buildWorkingCopyCloneFromRow bypasses the guards of the register methods whenever
+ * the aggregate rows are built by an ObjectLevelReadQuery of their own.
+ * <p>
+ * Committing such a UnitOfWork used to fail with a DescriptorException wrapping a
+ * NullPointerException, because the descriptor UnitOfWorkImpl.calculateChanges resolves for the
+ * aggregate is the uninitialized prototype registered on the session rather than the initialized
+ * clone held by the owning mapping.
+ */
+@RunWith(EmfRunner.class)
+public class TestAggregateChangeCalculation {
+
+    @Emf(name = "aggregateChangeEMF", createTables = DDLGen.DROP_CREATE,
+            classes = { AggregateChangeEntity.class, AggregateChangeEmbeddable.class })
+    private EntityManagerFactory emf;
+
+    @Emf(name = "batchFetchAggregateChangeEMF", createTables = DDLGen.DROP_CREATE,
+            classes = { BatchFetchAggregateChangeEntity.class, AggregateChangeEmbeddable.class })
+    private EntityManagerFactory batchFetchEmf;
+
+    /**
+     * Selecting the embeddables of an element collection directly registers them in the UnitOfWork.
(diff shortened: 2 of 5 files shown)


arjantijms:
>  No test is included here — glad to add one to the JPA suite if you'd
like, with a pointer to where it best belongs.

Yes please on the test. Probably in jpa/eclipselink.jpa.test.jse/src/it/java/ with @Emf/EmfRunner. The #1455 query plus commit, and a second for the @BatchFetch + PESSIMISTIC_WRITE route; both fail on master today.

dmatej (approved):
Seems reasonable to me.

robertpatrick:
@arjantijms LGTM

arjantijms:
Der Lukas is nicht langer da, der @lukasj ist verschwunden. Let's merge it now, thanks! :)

PR #2806, opened 2026-08-11
#2780: stop caching concrete subclass calls to avoid pinning UnitOfWork sessions

### LINKED ISSUE #2780: Memory leak: ObjectLevelReadQuery.concreteSubclassCalls pins the executing UnitOfWork (and its entire working-copy graph) through the cached DatabaseCall
**Describe the bug**

When a JOINED multi-table inheritance hierarchy is read polymorphically (default shouldOuterJoinSubclasses = false), EclipseLink prepares one SQL call per concrete subclass and caches it in ObjectLevelReadQuery.concreteSubclassCalls (ExpressionQueryMechanism.selectAllRowsFromConcreteTable / selectOneRowFromConcreteTable).

The cached DatabaseCall keeps strong references it does not need, and since the owning query is permanent (a mapping's selectionQuery, a named query, a DescriptorQueryManager.cachedExpressionQueries entry or a JPQL parse-cache entry — all reachable from the ServerSession/Project), each cache entry permanently pins one RepeatableWriteUnitOfWork together with its entire registered working-copy object graph.

There are two independent anchors:

cachedCall.query → execution-time query clone → DatabaseQuery.session = the UnitOfWork. Worse, the anchor is refreshed on every cache hit: the hit path calls setCall(cachedCall), and DatasourceCallQueryMechanism.setCall() does call.setQuery(getQuery()) on the cached instance (DatasourceCallQueryMechanism.java:843-848 in 4.0.4). So the cached call always points at the most recent UnitOfWork that traversed that (query, concrete subclass) pair — there is always exactly one UoW pinned per entry, rotating but never released.

cachedCall.parameters may contain ParameterExpressions whose ExpressionBuilder.session captured the session that was active during prepare — a second, independent UnitOfWork anchor (ExpressionBuilder.session is only needed during normalization, which has already happened).

Notably, the prepare code already avoids the analogous problem for the translation row — selectAllRowsFromConcreteTable nulls query.translationRow around the put with the comment "Must clear the translation row to avoid in-lining parameters" (ExpressionQueryMechanism.java:2818-2830) — but the call's query back-reference and the parameter expressions' builder sessions receive no such treatment.

Production evidence (Eclipse MAT over a 4.1 GB heap dump):

~350 live RepeatableWriteUnitOfWork / ClientSession instances that should have been collected long ago.
A single lazy @OneToMany selection query targeting a JOINED root with 14 concrete subclasses retained 557 MB, all under its concreteSubclassCalls map: one cached SQLCall alone retained 225 MB = 112 MB via call.query.session (a full UoW working-copy graph) + 112 MB via call.parameters (a second, disjoint UoW graph captured at prepare time).
Path to GC roots of the pinned UoW: Project.orderedDescriptors → RelationalDescriptor.mappings → OneToManyMapping.selectionQuery (ReadAllQuery) → concreteSubclassCalls (ConcurrentHashMap) → SQLCall.query (execution clone) → session = RepeatableWriteUnitOfWork.
Adding the other reservoirs (named queries, cachedExpressionQueries of two other JOINED hierarchies, JPQL parse cache), the mechanism accounted for ~1 GB (23 %) of the heap and eventually produced OutOfMemoryError in production.
Entities loaded into those pinned UoWs are duplicated once per pinned UoW: with only 700 rows in one of our tables, the heap contained 31,234 instances (~45 working copies each).

**To Reproduce**

EclipseLink version: 4.0.4. The relevant code is byte-identical in 4.0.9 (ExpressionQueryMechanism.java has no diff at all) and functionally identical in 5.0.1 (only import/generics cleanup), so all current release lines are affected.
Java/JDK version: not JDK-specific (observed under JDK 21 / ZGC).
Database: MySQL 8 (not database-specific).
Context: plain JPA (also reproduces under Spring ORM; weaving static or dynamic is irrelevant).
Entity model:

```
@Entity
public class Parent {
    @Id @GeneratedValue Long id;
    @OneToMany(fetch = FetchType.LAZY, mappedBy = "parent")
    Set<Child> children = new HashSet<>();
}

@Entity
@Inheritance(strategy = InheritanceType.JOINED)
public abstract class Child {
    @Id @GeneratedValue Long id;
    @ManyToOne Parent parent;
}

@Entity public class ChildA extends Child { String a; }
@Entity public class ChildB extends Child { String b; }
```
Steps:

```
// 1. Persist one Parent with one ChildA and one ChildB, commit, clear.

// 2. In a new transaction, trigger the polymorphic lazy load:
em.getTransaction().begin();
Parent p = em.find(Parent.class, parentId);
p.getChildren().size();               // prepares + caches one call per concrete subclass
em.getTransaction().commit();
em.close();                           // the UnitOfWork should now be collectable... but is not

// 3. Inspect the permanent selection query:
ServerSession server = emf.unwrap(ServerSession.class);
OneToManyMapping mapping = (OneToManyMapping) server.getDescriptor(Parent.class)
        .getMappingForAttributeName("children");
ObjectLevelReadQuery selectionQuery = (ObjectLevelReadQuery) mapping.getSelectionQuery();
for (DatabaseCall cached : selectionQuery.getConcreteSubclassCalls().values()) {
    System.out.println(cached.getQuery().getSession());
    // -> RepeatableWriteUnitOfWork of the CLOSED EntityManager, strongly reachable forever
}
```
A jcmd <pid> GC.run + GC.class_histogram confirms the RepeatableWriteUnitOfWork count never drops; repeating step 2 with new EntityManagers keeps exactly one UoW pinned per (query, concrete subclass) entry, each retaining every working copy it ever registered.

**Expected behavior**

The per-subclass call cache should retain only what is needed to re-execute the SQL (query string, fields, parameter descriptors) — never the execution-time query clone, its session/UnitOfWork, or expression builders still pointing at a session. UnitOfWorks must become collectable after commit/close regardless of which cached calls they prepared or hit.

**Additional context**

Why the retained references are dead weight (verified against 4.0.4 sources):

Execution never uses the cached call's query: DatasourceCallQueryMechanism.executeCall() clones the call and does clonedCall.setQuery(this.query) before translate() (DatasourceCallQueryMechanism.java:273-281). The query field on the cached instance is only ever written (by setCall on hits), never read.
Sharing the parameter list template is safe: translate() replaces the list on the execution clone (setParameters(new ArrayList<>(...)), DatasourceCall.java:956), it never mutates the template.
ParameterExpression.getValue(...) receives the session as an argument (ParameterExpression.java:217/225); ExpressionBuilder.session is only read during prepare-time normalization.
Suggested fix (both cache sites in ExpressionQueryMechanism, lines 2829 and 2893 in 4.0.4):

At put time, cache a stripped shallow clone: DatabaseCall cached = (DatabaseCall) call.clone(); cached.setQuery(null); (optionally also clear ExpressionBuilder.session of any Expression in its parameters).
On cache hit, pass a clone to setCall((DatabaseCall) cached.clone()) so setCall's call.setQuery(getQuery()) re-pollutes a throwaway instance instead of the cached one.
We have been running exactly this behavior in production via a reflection-installed ConcurrentHashMap subclass over the concreteSubclassCalls field (put = sanitized stripped clone, get = defensive clone) plus setExpressionQueryCacheMaxSize(0) on JOINED descriptors: pinned UnitOfWorks went from ~350 down to 31 (only the genuinely live ones) and the count stays flat, with no functional regressions — which also empirically confirms that the stripped references are never needed.

I can provide the MAT screenshots (dominator tree, cached SQLCall expansion) and a runnable reproducer if useful.

adityaanikam:
Fixes #2780.

`selectAllRowsFromConcreteTable` and `selectOneRowFromConcreteTable` in `ExpressionQueryMechanism` cache the prepared `DatabaseCall` per concrete subclass on the query, in `concreteSubclassCalls`. The cached call keeps a path back to the `UnitOfWork` that was active when it was first built, through its parameter expressions, the `ExpressionBuilder`, the `SQLSelectStatement` and that statement's query. The cache is never cleared, so a long-lived shared query can keep that unit of work reachable for as long as the process runs.

As discussed above, removing the cache is too broad, and fixing the retention itself is a larger change. This PR adds a way to opt out instead:

- `ObjectLevelReadQuery.setShouldCacheConcreteSubclassCalls(boolean)`, which defaults to `true`, so existing behaviour is unchanged
- the query hint `eclipselink.query.cache-concrete-subclass-calls` (`QueryHints.CACHE_CONCRETE_SUBCLASS_CALLS`) to set it from JPA

When it is `false`, both methods neither read nor write `concreteSubclassCalls` and build the call on every execution.

The setting belongs to the query it is set on. A query that prepares itself from an equivalent cached query (`prepareFromCachedQuery`, then `prepareFromQuery`) keeps its own value, because `prepareFromQuery` only copies properties that affect the SQL. `copyFromQuery`, used when a query is morphed into another query type, carries it over like the other per-query settings.

`ObjectLevelReadQueryConcreteSubclassCallsCacheTest` covers the default, the setter, `clone()`, `copyFromQuery()`, and `prepareFromQuery()` in both directions (cached query enabled with the new query disabled, and the reverse) for `ReadAllQuery` and `ReadObjectQuery`.

### DIFF
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/QueryHints.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/config/QueryHints.java
@@ -491,6 +491,22 @@ public final class QueryHints {
     public static final String PREPARE = "eclipselink.prepare";
 
     /**
+     * "eclipselink.query.cache-concrete-subclass-calls"
+     * <p>Configures the query to not cache the prepared call used to read a concrete
+     * subclass for an inheritance query.
+     * By default, this call is cached and reused on every execution of a shared
+     * (e.g. named) query, to avoid the cost of rebuilding it every time.
+     * A cached call can retain a reference to whichever unit of work first triggered
+     * its build, which for a long-lived shared query can pin that unit of work (and
+     * everything it references) in memory indefinitely. Disabling this rebuilds the
+     * call on every execution instead, trading that memory retention for extra CPU cost.
+     * Valid values are:  HintValues.FALSE, HintValues.TRUE,
+     * "" could be used instead of default value HintValues.FALSE
+     * @see org.eclipse.persistence.queries.ObjectLevelReadQuery#setShouldCacheConcreteSubclassCalls(boolean)
+     */
+    public static final String CACHE_CONCRETE_SUBCLASS_CALLS = "eclipselink.query.cache-concrete-subclass-calls";
+
+    /**
      * "eclipselink.jdbc.cache-statement"
      * <p>Configures if the query will cache its JDBC statement.
      * This allows queries to use parameterized SQL with statement caching.
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/internal/queries/ExpressionQueryMechanism.java
@@ -2885,8 +2885,9 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
         // PERF: First check the subclass calls cache for the prepared call.
         // Must clear the translation row to avoid in-lining parameters unless not a prepared query.
         boolean shouldPrepare = query.shouldPrepare();
+        boolean shouldCacheCall = shouldPrepare && query.shouldCacheConcreteSubclassCalls();
         DatabaseCall call = null;
-        if (shouldPrepare) {
+        if (shouldCacheCall) {
             call = query.getConcreteSubclassCalls().get(query.getReferenceClass());
         }
         if (call == null) {
@@ -2897,11 +2898,13 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
             setSQLStatement(buildConcreteSelectStatement());
             // Must also build the call.
             super.prepareSelectAllRows();
-            if (shouldPrepare) {
+            if (shouldCacheCall) {
                 if (query.hasJoining()) {
                     query.getConcreteSubclassJoinedMappingIndexes().put(query.getReferenceClass(), query.getJoinedAttributeManager().getJoinedMappingIndexes_());
                 }
                 query.getConcreteSubclassCalls().put(query.getReferenceClass(), (DatabaseCall)this.call);
+            }
+            if (shouldPrepare) {
                 query.setTranslationRow(translationRow);
             }
         } else {
@@ -2949,8 +2952,9 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
         // PERF: First check the subclass calls cache for the prepared call.
         // Must clear the translation row to avoid in-lining parameters unless not a prepared query.
         boolean shouldPrepare = query.shouldPrepare();
+        boolean shouldCacheCall = shouldPrepare && query.shouldCacheConcreteSubclassCalls();
         DatabaseCall call = null;
-        if (shouldPrepare) {
+        if (shouldCacheCall) {
             call = query.getConcreteSubclassCalls().get(query.getReferenceClass());
         }
         if (call == null) {
@@ -2961,11 +2965,13 @@ public class ExpressionQueryMechanism extends StatementQueryMechanism {
             setSQLStatement(buildConcreteSelectStatement());
             // Must also build the call.
             super.prepareSelectOneRow();
-            if (shouldPrepare) {
+            if (shouldCacheCall) {
                 if (query.hasJoining()) {
                     query.getConcreteSubclassJoinedMappingIndexes().put(query.getReferenceClass(), query.getJoinedAttributeManager().getJoinedMappingIndexes_());
                 }
                 query.getConcreteSubclassCalls().put(query.getReferenceClass(), (DatabaseCall)this.call);
+            }
+            if (shouldPrepare) {
                 query.setTranslationRow(translationRow);
             }
         } else {
--- a/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/ObjectLevelReadQuery.java
+++ b/foundation/org.eclipse.persistence.core/src/main/java/org/eclipse/persistence/queries/ObjectLevelReadQuery.java
@@ -202,6 +202,16 @@ public abstract class ObjectLevelReadQuery extends ObjectBuildingQuery {
     /** Allow concrete subclasses calls to be prepared and cached for inheritance queries. */
     protected Map<Class<?>, DatabaseCall> concreteSubclassCalls;
 
+    /**
+     * Allow the concrete subclasses call cache to be disabled for inheritance queries.
+     * Enabled (true) by default, matching the existing behavior. A cached call for a
+     * long-lived shared query (such as a named query) retains a reference to whichever
+     * unit of work first triggered its build, which can pin that unit of work in memory
+     * for as long as the query is cached; disabling the cache avoids this at the cost of
+     * rebuilding the call on every execution.
+     */
+    protected boolean shouldCacheConcreteSubclassCalls = true;
+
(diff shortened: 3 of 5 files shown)


arjantijms:
> This removes the cache entirely so both methods always rebuild the call,

Maybe that is a bit too much. The cache is there for a reason. Maybe better to look at the retention problem itself?

> This removes the cache entirely so both methods always rebuild the call, which is what @ vrivon validated in production on https://github.com/eclipse-ee4j/eclipselink/issues/2780:

Did not @vrivon ran a sanitized-clone workaround, not cache removal?

vrivon:
Both, in that order — and the order is the whole point.

I first ran the sanitized-clone workaround: cache a shallow clone with setQuery(null), additionally null ExpressionBuilder.session on the parameters' expressions, and hand out defensive clones on cache hits. It looked like it worked — pinned UoW counts dropped immediately. It did not hold. Over the following days the live set grew back at ~480 MB/day to 3.6 GB, and a second heap dump showed the anchor was still there through a third path I had missed:

`SQLCall.parameters → ParameterExpression.localBase → ExpressionBuilder.statement → SQLSelectStatement.query → session = RepeatableWriteUnitOfWork`

Only after that did I disable the cache entirely, and that is what held: 3.6 GB → ~1.1 GB, flat since, allocation stalls from dozens a day to zero. I've corrected the issue body to retract the earlier claim.

To be clear about what that does and doesn't argue for: I'm not proposing cache removal as the upstream fix, and I agree with you that it's too much. I disabled it in my own application because I was patching EclipseLink from the outside by reflection, and I couldn't safely mutate internal structures whose invariants I'm not in a position to verify — the failed sanitized clone is precisely what that risk looks like in practice. Upstream doesn't have that constraint and should fix the retention itself.

For a retention fix, the three anchors identified so far are: cachedCall.query (re-set on every cache hit by setCall()), ExpressionBuilder.session (captured by SQLSelectStatement.normalize() from the query's current session rather than the root session), and the statement/query back-reference reachable from the cached call's parameters. The third one is what survives the two obvious fixes.

One data point on cost, offered neutrally rather than as an argument: with the cache disabled I measured no CPU or latency difference in my workload (heavy polymorphic lazy loading over JOINED hierarchies). That may simply mean my workload isn't where the cache earns its keep.

arjantijms:
Thanks for the extra data point. 

The question remains what to do with this PR then? The problem seems real, but completely removing the cache is probably not the best option. Without fully addressing the true core issue (fixing the retention itself), simply introducing an option to disable the cache might be a path worth considering.

adityaanikam:
Pushed a new approach: instead of removing the concrete subclass call cache entirely, this adds a query hint (`eclipselink.query.cache-concrete-subclass-calls`, default `true`) to opt out of it per query. Default behavior is unchanged. Disabling it avoids the UnitOfWork retention from #2780, at the cost of rebuilding the call on every execution instead of caching it.

robertpatrick:
@adityaanikam In looking at this PR, I have a question about prepared-query reuse and the desired behavior.  What I see is the following, which was observed for both `ReadAllQuery` and `ReadObjectQuery`.  What I saw was the following:

| Cached query’s setting | New query requests | Setting observed at execution |
| ---- | ---- | ---- |
| Enabled | Disabled | Enabled |
| Disabled | Enabled | Disabled |

This appears to be happening due to the cached setting being copied into the new query before the new query's execution clone is created, which causes the new query to execute with the cached query's setting instead of the new query's requested setting.

If this is intentional, it would be very confusing (at least to me).  It would be good to:

- clarify the intended behavior
- make sure that there are tests verifying the behavior you want
- edit the description of the PR to reflect the actual change instead of the old disabling the cache behavior

adityaanikam:
@robertpatrick Thanks for checking this, it was not intentional. `prepareFromQuery` copied the flag from the cached query, but that method only copies properties that affect the SQL, and this is a per-query setting. I removed it from `prepareFromQuery` and copy it in `copyFromQuery` with the other per-query settings instead. The tests now cover both rows of your table for `ReadAllQuery` and `ReadObjectQuery`, and I rewrote the description to describe the current change.

PR #2819, opened 2026-08-23
fix: support Java records as JPA @IdClass

hantsy:
@
Support a `java.lang.Record` used as a JPA `@IdClass`.

## Root cause

`CMPPolicy.createPrimaryKeyInstance()` assumed the id class could be built via a no-arg constructor followed by `accessor.setValue(...)`. Records have neither a no-arg constructor nor writable fields.

## Fix

Detect a record `pkClass` and build the primary key instance via `RecordInstantiationPolicy` using the record canonical constructor.

## Test

`TestRecordIdClass` uses a `RecordIdClass(UUID id, String code)` record as the `@IdClass` of `RecordIdClassEntity`, covering persist/find and `PersistenceUnitUtil.getIdentifier`.

Depends on #2817 (record `@Embeddable` policy plumbing).
@


hantsy:
Merged into #2817 — the record `@IdClass` fix is now part of the combined record-support PR. Closing in favor of #2817.

PR #2827, opened 2026-08-27
Apply skip/limit/batchSize to Mongo FindIterable before opening the cursor

### LINKED ISSUE #2814: Mongo NoSQL FIND ignores Query.setMaxResults() / setFirstResult() because skip/limit are applied after iterator()
**Describe the bug**

On the MongoDB NoSQL adapter, JPA `Query.setMaxResults()` and `Query.setFirstResult()` are accepted and mapped onto the interaction spec (`mongo.limit` / `mongo.skip`), but they are not applied to the MongoDB find.

`MongoDatabaseInteraction` opens the cursor with `FindIterable.iterator()` **before** calling `skip()`, `limit()`, and `batchSize()`. The Mongo Java driver executes the find when `iterator()` is called and copies the current limit into the `FindOperation` at that moment. A later `limit()` only mutates the iterable; it does not change the already-opened cursor.

The adapter then walks that unlimited cursor and materializes every matching document into a `MongoListRecord` (`ArrayList`). On a large collection this can fail with:

```
java.lang.OutOfMemoryError: Required array length 2147483639 + 1204 is too large
```

instead of returning a page of results.

**To Reproduce**

- EclipseLink version: 4.0.6 (`org.eclipse.persistence` + `org.eclipse.persistence.nosql`)
- Java/JDK version: 17
- JEE Server: WildFly / JBoss EAP 8 (also reproducible in a standalone `RESOURCE_LOCAL` SE test)
- Database provider/version: MongoDB 3.6+
- Driver: `org.mongodb:mongo-java-driver:3.12.14` (legacy `com.mongodb.MongoClient` API used by the 4.0.6 adapter)

Entity:

```java
import jakarta.persistence.Entity;
import jakarta.persistence.GeneratedValue;
import jakarta.persistence.Id;
import org.bson.types.ObjectId;
import org.eclipse.persistence.nosql.annotations.DataFormatType;
import org.eclipse.persistence.nosql.annotations.Field;
import org.eclipse.persistence.nosql.annotations.NoSql;

@Entity
@NoSql(dataType = "Item", dataFormat = DataFormatType.MAPPED)
public class Item {

    @Id
    @GeneratedValue
    @Field(name = "_id")
    private ObjectId id;

    private String name;

    public Item() {
    }

    public Item(String name) {
        this.name = name;
    }

    public ObjectId getId() {
        return id;
    }

    public String getName() {
        return name;
    }
}
```

`persistence.xml`:

```xml
<persistence xmlns="https://jakarta.ee/xml/ns/persistence" version="3.0">
    <persistence-unit name="test-mongo-pu" transaction-type="RESOURCE_LOCAL">
        <provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
        <class>Item</class>
        <properties>
            <property name="eclipselink.target-database"
                      value="org.eclipse.persistence.nosql.adapters.mongo.MongoPlatform"/>
            <property name="eclipselink.nosql.connection-spec"
                      value="org.eclipse.persistence.nosql.adapters.mongo.MongoConnectionSpec"/>
            <property name="eclipselink.nosql.property.mongo.host" value="localhost"/>
            <property name="eclipselink.nosql.property.mongo.port" value="27017"/>
            <property name="eclipselink.nosql.property.mongo.db" value="testdb"/>
        </properties>
    </persistence-unit>
</persistence>
```

Code:

```java
EntityManagerFactory entityManagerFactory = Persistence.createEntityManagerFactory("test-mongo-pu");
EntityManager em = entityManagerFactory.createEntityManager();

em.getTransaction().begin();
for (int i = 0; i < 25; i++) {
    em.persist(new Item("item-" + i));
}
em.getTransaction().commit();

List<Item> page = em.createQuery("SELECT i FROM Item i", Item.class)
        .setFirstResult(0)
        .setMaxResults(10)
        .getResultList();

// page.size() is 25 (all documents), not 10
System.out.println(page.size());

em.close();
entityManagerFactory.close();
```

**Expected behavior**

`page.size()` is at most 10. MongoDB receives a `find` with `limit: 10` (and `skip` when `setFirstResult` is used).

**Additional context**

`MongoPlatform.buildCallFromStatement()` already maps JPA pagination onto the interaction:

https://github.com/eclipse-ee4j/eclipselink/blob/4.0.6/foundation/org.eclipse.persistence.nosql/src/main/java/org/eclipse/persistence/nosql/adapters/mongo/MongoPlatform.java#L301-L308

```java
if (readQuery.getFirstResult() > 0) {
    interaction.setProperty(SKIP, readQuery.getFirstResult());
}
if (readQuery.getMaxRows() > 0) {
    interaction.setProperty(LIMIT, readQuery.getMaxRows());
}
```

Those properties are then copied onto `MongoInteractionSpec` in `buildInteractionSpec()`. The values never reach the server because FIND applies them after the cursor is opened:

https://github.com/eclipse-ee4j/eclipselink/blob/4.0.6/foundation/org.eclipse.persistence.nosql/src/main/java/org/eclipse/persistence/internal/nosql/adapters/mongo/MongoDatabaseInteraction.java#L154-L179

```java
FindIterable<Document> iterable = collection.find(object);
if (sort != null) {
    iterable.sort(sort);
}
MongoCursor<Document> cursor = iterable.iterator(); // find already sent with limit=0
try {
    if (mongoSpec.getSkip() > 0) {
        iterable.skip(mongoSpec.getSkip());          // too late
    }
    if (mongoSpec.getLimit() != 0) {
        iterable.limit(mongoSpec.getLimit());        // too late
    }
    if (mongoSpec.getBatchSize() != 0) {
        iterable.batchSize(mongoSpec.getBatchSize());
    }
```

Driver confirmation for mongo-java-driver 3.12.14:

- `FindIterable.limit()` only mutates `findOptions` on the iterable
  (https://github.com/mongodb/mongo-java-driver/blob/r3.12.14/driver-sync/src/main/com/mongodb/client/internal/FindIterableImpl.java#L69-L73)
- `iterator()` immediately calls `execute()` → `asReadOperation()`
  (https://github.com/mongodb/mongo-java-driver/blob/r3.12.14/driver-sync/src/main/com/mongodb/client/internal/MongoIterableImpl.java#L92-L150)
- `Operations.createFindOperation()` copies `options.getLimit()` into a new `FindOperation` at construction time
  (https://github.com/mongodb/mongo-java-driver/blob/r3.12.14/driver-core/src/main/com/mongodb/internal/operation/Operations.java#L184-L190)

Calling `limit()` / `skip()` / `batchSize()` after `iterator()` cannot change that cursor.

Suggested fix: apply `sort`, `skip`, `limit`, and `batchSize` on the `FindIterable` **before** `iterator()`:

```java
FindIterable<Document> iterable = collection.find(object);
if (sort != null) {
    iterable.sort(sort);
}
if (mongoSpec.getSkip() > 0) {
    iterable.skip(mongoSpec.getSkip());
}
if (mongoSpec.getLimit() != 0) {
    iterable.limit(mongoSpec.getLimit());
}
if (mongoSpec.getBatchSize() != 0) {
    iterable.batchSize(mongoSpec.getBatchSize());
}
try (MongoCursor<Document> cursor = iterable.iterator()) {
    // materialize results
}
```

A Maven JPA bug-test-case project can be generated from EclipseLink's archetype if a runnable attachment is needed:

```
mvn archetype:generate -DarchetypeGroupId=org.eclipse.persistence -DarchetypeArtifactId=org.eclipse.persistence.bug.jpa-archetype -DarchetypeVersion=5.0.0-SNAPSHOT -DgroupId=eclipselink.bug.testcase -DartifactId=jpa-testcase
```

adityaanikam:
﻿Fixes #2814

FindIterable.iterator() executes the query and snapshots the cursor's current options. Calling skip(), limit(), or batchSize() after that point mutates the (now-irrelevant) options object but has no effect on the already open cursor, so Query.setMaxResults() / setFirstResult() are silently ignored on Mongo FIND operations.

Move those three calls to before iterable.iterator() so they're actually applied.

Added a unit test using JDK dynamic proxies to fake the MongoDB driver interfaces (no live MongoDB instance needed), asserting skip/limit/batchSize are all called before iterator().

### DIFF
--- a/foundation/org.eclipse.persistence.nosql/src/main/java/org/eclipse/persistence/internal/nosql/adapters/mongo/MongoInteraction.java
+++ b/foundation/org.eclipse.persistence.nosql/src/main/java/org/eclipse/persistence/internal/nosql/adapters/mongo/MongoInteraction.java
@@ -167,17 +167,20 @@ public class MongoInteraction implements Interaction {
                 if (sort != null) {
                     iterable.sort(sort);
                 }
+                // skip/limit/batchSize must be set before the cursor is opened: FindIterable.iterator()
+                // executes the query and snapshots the current options, so calls made after it have no
+                // effect on the already-open cursor.
+                if (mongoSpec.getSkip() > 0) {
+                    iterable.skip(mongoSpec.getSkip());
+                }
+                if (mongoSpec.getLimit() != 0) {
+                    iterable.limit(mongoSpec.getLimit());
+                }
+                if (mongoSpec.getBatchSize() != 0) {
+                    iterable.batchSize(mongoSpec.getBatchSize());
+                }
                 MongoCursor<Document> cursor = iterable.iterator();
                 try {
-                    if (mongoSpec.getSkip() > 0) {
-                        iterable.skip(mongoSpec.getSkip());
-                    }
-                    if (mongoSpec.getLimit() != 0) {
-                        iterable.limit(mongoSpec.getLimit());
-                    }
-                    if (mongoSpec.getBatchSize() != 0) {
-                        iterable.batchSize(mongoSpec.getBatchSize());
-                    }
                     if (!cursor.hasNext()) {
                         return null;
                     }
--- a/foundation/org.eclipse.persistence.nosql/src/test/java/org/eclipse/persistence/internal/nosql/adapters/mongo/MongoInteractionFindSkipLimitOrderTest.java
+++ b/foundation/org.eclipse.persistence.nosql/src/test/java/org/eclipse/persistence/internal/nosql/adapters/mongo/MongoInteractionFindSkipLimitOrderTest.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+package org.eclipse.persistence.internal.nosql.adapters.mongo;
+
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.bson.Document;
+import org.junit.Test;
+
+import com.mongodb.client.FindIterable;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoCursor;
+import com.mongodb.client.MongoDatabase;
+
+import jakarta.resource.cci.Record;
+
+/**
+ * https://github.com/eclipse-ee4j/eclipselink/issues/2814
+ *
+ * FindIterable.iterator() executes the query and snapshots the current cursor
+ * options: skip/limit/batchSize calls made after it have no effect on the
+ * already-open cursor. This verifies MongoInteraction applies them to the
+ * FindIterable before opening the cursor, using fake (proxy-based) MongoDB
+ * driver objects so the test needs no live MongoDB instance.
+ */
+public class MongoInteractionFindSkipLimitOrderTest {
+
+    @Test
+    public void skipLimitAndBatchSizeAreAppliedBeforeTheCursorIsOpened() throws Exception {
+        List<String> callOrder = new ArrayList<>();
+
+        MongoCursor<Document> cursor = proxy(MongoCursor.class, callOrder, method -> {
+            switch (method.getName()) {
+                case "hasNext":
+                    return false;
+                case "close":
+                    return null;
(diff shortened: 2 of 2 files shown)


robertpatrick:
@arjantijms This PR looks good to me but there is a project-level issue that impacts its new test class. 

The `nosql` module explicitly disables Surefire so the supplied test (which executes correctly and covers the change) will not execute in CI.   I tried just enabling Surefire and that fails miserably due to picking up the integration tests.  In theory, we could explicitly add the test to the Failsafe execution but that makes for ongoing maintenance issues (i.e., having to explicitly add new tests to the Failsafe list).  Not sure what we want to do.

adityaanikam:
@robertpatrick @arjantijms Thanks for checking. The integration tests Surefire picks up come from `src/it/java`, which the parent POM adds as a test source directory (`add-integration-test-sources`), so plain Surefire also runs the 68 `*Test` classes under `org/eclipse/persistence/testing` there.

One option that avoids listing tests in Failsafe is to replace `<skip>true</skip>` with an include pattern, the same way `eclipselink.core.test` limits its default Surefire run to `org/eclipse/persistence/testing/tests/junit/**/*Test`. I pushed that as a separate commit using `org/eclipse/persistence/internal/nosql/**/*Test`. Locally, `mvn -pl foundation/org.eclipse.persistence.nosql test` now runs only `MongoInteractionFindSkipLimitOrderTest`, which passes, and it fails with the `MongoInteraction` change reverted. Unit tests added under that package later would run without further configuration.

Happy to drop the commit if you would rather handle it another way.

Issue #2833, opened 2026-09-02
"The select item is missing from the result variable declaration" issue due to concurrent access of mutable SelectClauseVisitor


adamsaghy:
**Describe the bug**
Useful resources are
- Stack trace
```
Test testLoanCOBPartitioningQuery() FAILED (4.7s)  

[...]

PR #2835, opened 2026-09-02
ISSUE-2833: Avoid using mutable singleton visitor in ResultVariableFactory

adamsaghy:
**Issue:**

 https://github.com/eclipse-ee4j/eclipselink/issues/2833

**Fix:**

**Avoid using mutable, singleton visitor in the factory class**

Each `buildExpression()` invocation creates its own `SelectClauseVisitor`, so concurrent JPQL parses cannot overwrite an another parse’s supported state.


robertpatrick:
@arjantijms LGTM

